PR #16, opened 2025-05-23
Fix Attribute rolling. Updated some depreciated values based on warni…

IrkTheImp:
Fixed attributes to now properly bring up the dialog and roll dice. Update nameAttr to valueAttr based on the values being depreciated.


WBHarry (approved):
Awesome. Yeah there's definately a bunch of changes like these needed in different places. :+1:

PR #33, opened 2025-05-23
feat(damageThresholds): PCs use armor base + current level

### LINKED ISSUE #23: add armor items thresholds to PC hp thresholds + clean up thresholds in PC
Base thresholds are present in the armors, we need to update PCs with base thresholds + level.

let's leave class, heritage and domain features for later.

emiliojc:
closes #23

### DIFF
--- a/module/applications/sheets/pc.mjs
+++ b/module/applications/sheets/pc.mjs
@@ -5,22 +5,6 @@ import DhpLevelup from '../levelup.mjs';
 import AncestrySelectionDialog from '../ancestrySelectionDialog.mjs';
 import DaggerheartSheet from './daggerheart-sheet.mjs';
 
-function computeDamageThresholds(armor, currentLevel) {
-    if (!armor) {
-        return {
-            major: currentLevel,
-            severe: currentLevel * 2
-        };
-    }
-    const {
-        baseThresholds: { major = 0, severe = 0 }
-    } = armor.system;
-    return {
-        major: major + currentLevel,
-        severe: severe + currentLevel
-    };
-}
-
 const { ActorSheetV2 } = foundry.applications.sheets;
 export default class PCSheet extends DaggerheartSheet(ActorSheetV2) {
     constructor(options = {}) {
@@ -301,11 +285,6 @@ export default class PCSheet extends DaggerheartSheet(ActorSheetV2) {
         if (context.inventory.length === 0) {
             context.inventory = Array(1).fill(Array(5).fill([]));
         }
-
-        this.document.damageThresholds = computeDamageThresholds(
-            context.source.items.find(x => x.type === 'armor'),
-            context.source.system.levelData.currentLevel
-        );
 
         context.classFeatures = (
             this.multiclassFeatureSetSelected
--- a/module/data/pc.mjs
+++ b/module/data/pc.mjs
@@ -380,9 +380,27 @@ export default class DhpPC extends foundry.abstract.TypeDataModel {
 
         this.evasion = this.class?.system?.evasion ?? 0;
         // this.armor.value = this.activeArmor?.baseScore ?? 0;
+        this.damageThresholds = this.computeDamageThresholds();
 
         this.applyLevels();
         this.applyEffects();
+    }
+
+    computeDamageThresholds() {
+        // TODO: missing weapon features and domain cards calculation
+        if (!this.armor) {
+            return {
+                major: this.levelData.currentLevel,
+                severe: this.levelData.currentLevel * 2
+            };
+        }
+        const {
+            baseThresholds: { major = 0, severe = 0 }
+        } = this.armor.system;
+        return {
+            major: major + this.levelData.currentLevel,
+            severe: severe + this.levelData.currentLevel
+        };
     }
 
     applyLevels() {
--- a/templates/sheets/parts/health.hbs
+++ b/templates/sheets/parts/health.hbs
@@ -7,7 +7,7 @@
         </div>
 
         <div class="threshold-box">
-            {{document.damageThresholds.major}}
+            {{document.system.damageThresholds.major}}
         </div>
         <div class="threshold-spacer">
             <i class="fa-solid fa-caret-left"></i>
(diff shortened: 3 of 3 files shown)


WBHarry:
Hm. If it's just based on the actor's level and equipped armor, maybe it should be a derived property or a getter method of the PC datamodel instead of placing it in the sheet. :thinking:

module/data/pc.mjs
```diff
@@ -92,11 +88,6 @@ export default class DhpPC extends foundry.abstract.TypeDataModel {
                 min: new fields.NumberField({ initial: 1, integer: true }),
                 max: new fields.NumberField({ initial: 6, integer: true })
             }),
-            damageThresholds: new fields.SchemaField({
-                minor: new fields.NumberField({ initial: 0, integer: true }),
-                major: new fields.NumberField({ initial: 0, integer: true }),
-                severe: new fields.NumberField({ initial: 0, integer: true })
-            }),
```
moliloo, line 99: i prefer to maintain the dataModel of the actors we use, avoid create schemas dynamically, if we need a data from something, it have to be in dataModel

templates/sheets/parts/health.hbs
```diff
@@ -2,24 +2,20 @@
...
-            {{document.system.damageThresholds.minor}}
-        </div>
         <div class="threshold-spacer">
-            <i class="fa-solid fa-caret-left"></i>
             <div class="health-category">{{localize "DAGGERHEART.Sheets.PC.Health.Minor"}}</div>
         </div>
 
         <div class="threshold-box">
-            {{document.system.damageThresholds.major}}
+            {{document.damageThresholds.major}}
```
moliloo: keep the data in system, avoid change the hierarchy from documents

templates/sheets/parts/health.hbs
```diff
@@ -2,24 +2,20 @@
...
         </div>
         <div class="threshold-spacer">
             <i class="fa-solid fa-caret-left"></i>
             <div class="health-category">{{localize "DAGGERHEART.Sheets.PC.Health.Major"}}</div>
         </div>
 
         <div class="threshold-box">
-            {{document.system.damageThresholds.severe}}
+            {{document.damageThresholds.severe}}
```
moliloo: keep the data in system, avoid change the hierarchy from documents

module/applications/sheets/pc.mjs
```diff
@@ -286,6 +302,11 @@ export default class PCSheet extends DaggerheartSheet(ActorSheetV2) {
             context.inventory = Array(1).fill(Array(5).fill([]));
         }
 
+        this.document.damageThresholds = computeDamageThresholds(
+            context.source.items.find(x => x.type === 'armor'),
+            context.source.system.levelData.currentLevel
+        );
```
moliloo: make sure to when you create the data, keep this inside of  `actor.system` propriety, i think is a bit odd create this outside this, since all the dataModel is create inside the `actor.system` propriety

WBHarry:
@moliloo 
If we do keep DamageThreshold as a saved DB field, it does meant we would have to update it every time a PC changes their level or equips/unequips armor.

If we add it as a derived property, then on world startup, and every time an actor experiences a DB update, it will be recalculated automatically.
Thinking like this:
```
prepareDerivedData() {
   const { baseThresholds } = this.getArmor();
   this.damageThresholds =  {
       major: armor ? this.levelData.currentLevel + baseThresholds.major : this.levelData.currentLevel * 2,
       severe: armor ? this.levelData.currentLevel + baseThresholds.severe : this.levelData.currentLevel * 2,
   };
}
```
This way we don't risk forgetting to recalculate a static DB field, and we can still always access `PC.system.damageThresholds` anywhere without dynamically generating it at the time.

emiliojc:
yeah, I was thinking like @WBHarry when implementing the armor read. I can move it no problem to the data model but I was thinking of keeping it fully dynamic due to weapon features that increase thresholds or armors (instead of saving it in the data model)

image

emiliojc:
And checking there seem even temporary bonuses to HP thresholds :/

image

WBHarry:
If other factors such as weapon features further modify the thresholds, then I think it's an even better idea to have it as a derived property. Means we could check for such features and add that logic upon `prepareDerivedData`. Same for effects the PC will have on them.
Even if we had a field directly in the DB, it could never be used in isolation even if we kept it updated, still requiring us to derive based on it 😄 

**Edit**
I gave it a try just to remind myself, and yes, even if you directly edit an item belonging to an actor's `items` collection it will trigger `prepareDerivedData` again, so it's safe.

module/applications/sheets/pc.mjs
emiliojc: moved data to the dataModel, 100% right

module/data/pc.mjs
emiliojc: I preferred to make it dynamic, or derived, due to the number of things that can modify the damageThresholds: weapon features, domain cards, armor that you can change, etc.

module/data/pc.mjs
```diff
@@ -389,21 +380,35 @@ export default class DhpPC extends foundry.abstract.TypeDataModel {
 
         this.evasion = this.class?.system?.evasion ?? 0;
         // this.armor.value = this.activeArmor?.baseScore ?? 0;
-        this.damageThresholds = this.class?.system?.damageThresholds ?? { minor: 0, major: 0, severe: 0 };
+        this.damageThresholds = this.computeDamageThresholds();
```
WBHarry, line 383: This works for me, but leaving it to @moliloo to sign of on it, incase there are other considerations :+1:

moliloo:
If `_prepareData` its triggered for every change, so its fine to me to create dinamically, i thought `_prepareData` actives onde when open the sheet.

But even dinamically, i still prefeer to keep then in system propriety, thats ok to you? @emiliojc

module/data/pc.mjs
moliloo: Thats fair, you can keep it derived then

module/data/pc.mjs
moliloo: I dont have others considerations, resolvind this conversation

PR #39, opened 2025-05-23
added T4 equipment and class features

cptn-cosmo:


WBHarry:
Awesome.

One thing though. I notice the `daggerheart.css` file got changed around a lot. If you didn't, you should run `npm install` again to make sure you have `prettify` downloaded. Then you need to either make a new commit or run `npx prettify . --write` (the second is proably a more sure bet in this case).
Should make the css file go back to the same state, and have you good from now on 🙃

WBHarry:
Actually, nvm. I think prettifier didn't touch the style class earlier. Should be good :+1:

PR #50, opened 2025-05-25
fix sheet instance issue

IrkTheImp:
remove id from sheets to allow multiple sheets of same type to be open. Added id to roll selection so only a single roll modal will be open. No more spawning infinite dice rollers by click the roll icon on a PC sheet.  It's possible this was causing other weirdness as well.


IrkTheImp:
close #50

WBHarry:
> LGTM
> 
> Makes perfect sense, but I think @WBHarry added this originally. Maybe there was a reason for it? 🤔

No particular reason. Just ended up with IDs when mass making stuff 🙃

PR #51, opened 2025-05-25
fix height issue, missing text, and red color

### LINKED ISSUE #31: description field disappears from features
"Feature" Items have a description field - however, it disappears when changing the Feature Type to something else from the "Normal" Value. It should always be available.

IrkTheImp:
close #31 

Made removed window height, so now it resizes. Added missing text field. Removed super red background color I'm thinking was added by mistake.

### DIFF
--- a/module/applications/sheets/feature.mjs
+++ b/module/applications/sheets/feature.mjs
@@ -14,7 +14,8 @@ export default class FeatureSheet extends DaggerheartSheet(ItemSheetV2) {
         tag: 'form',
         id: 'daggerheart-feature',
         classes: ['daggerheart', 'sheet', 'feature'],
-        position: { width: 600 },
+        position: { width: 600, height: 600 },
+        window: { resizable: true },
         actions: {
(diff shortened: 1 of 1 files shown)


emiliojc:
Description field doesn't disappear 👏 but I cannot resize. Shouldn't there be a parameter `resizeble: true` for the resize to work. We can leave the position as 600 for now to use as default.

image

module/applications/sheets/feature.mjs
```diff
@@ -14,7 +14,7 @@ export default class FeatureSheet extends DaggerheartSheet(ItemSheetV2) {
         tag: 'form',
         id: 'daggerheart-feature',
         classes: ['daggerheart', 'sheet', 'feature'],
-        position: { width: 600, height: 600 },
+        position: { width: 600 },
```
emiliojc:
```suggestion
        position: { width: 600, height: 600 },
        window: { resizable: true },
```

PR #52, opened 2025-05-25
Embedding Duality Rolls

### LINKED ISSUE #1: Embedding Rolls needs to be added
https://foundryvtt.com/article/dice/

`The boulders drop onto the player character, dealing [[2d12]] bludgeoning damage.`

WBHarry:
Closes #1 
**_The -1,041 deletions are mainly that I finally went ahead and removed the outcommented v.1.3 version of the PC sheet. It was like half the file. No actual changes made in the file. 😆_** 

Added duality roll functionality to chat and text.

- Roll buttons can be generated using the: `[[/dr]]` syntax.
```
[[/dr]]
[[/dr attribute="agility"]]
```
image
The buttons are rendered this way anywhere enrichment is used on text. 
Notably, for the button to actually do anything we need to attach the function on render. Currently this is done for **ChatMessages**, **Journals** and **Applications**. Alternatively could be moved into system replacement classes for these, such as our DhpChatMesssage class, but I didn't do that in this Issue.  
image

- Direct rolls can be made using the: `/dr` syntax.
```
/dr
/dr attribute="instinct" advantage=true
/dr hope="d20" fear="d20" attribute="instinct" disadvantage=true
```
image

### DIFF
--- a/styles/daggerheart.css
+++ b/styles/daggerheart.css
@@ -1561,13 +1561,13 @@
   font-size: 17px;
 }
 .chat-message.duality.hope {
-  background: linear-gradient(0, brown 40px, black);
+  background: linear-gradient(0, rgba(165, 42, 42, 0.6) 40px, rgba(0, 0, 0, 0.6));
 }
 .chat-message.duality.fear {
-  background: linear-gradient(0, blue, #0f0f61);
+  background: linear-gradient(0, rgba(0, 0, 255, 0.6), rgba(15, 15, 97, 0.6));
 }
 .chat-message.duality.critical {
-  background: linear-gradient(0, purple, #250825);
+  background: linear-gradient(0, rgba(128, 0, 128, 0.6), rgba(37, 8, 37, 0.6));
 }
 .chat-message.duality .dice-roll {
   color: var(--color-dark-1);
--- a/styles/variables/colors.less
+++ b/styles/variables/colors.less
@@ -22,13 +22,13 @@
 
 /* Duality */
 @hope: #ffe760;
-@hopeBackgroundStart: black;
-@hopeBackgroundEnd: brown;
+@hopeBackgroundStart: rgba(0, 0, 0, 0.6);
+@hopeBackgroundEnd: rgba(165, 42, 42, 0.6);
 @fear: #0032b1;
 @fearAccent: #2555cd;
-@fearBackgroundStart: #0f0f61;
-@fearBackgroundEnd: blue;
+@fearBackgroundStart: rgba(15, 15, 97, 0.6);
+@fearBackgroundEnd: rgba(0, 0, 255, 0.6);
 @critical: #430070;
 @criticalAccent: #66159c;
-@criticalBackgroundStart: #250825;
-@criticalBackgroundEnd: purple;
+@criticalBackgroundStart: rgba(37, 8, 37, 0.6);
+@criticalBackgroundEnd: rgba(128, 0, 128, 0.6);


module/helpers/utils.mjs
```diff
@@ -82,3 +71,32 @@ export const generateId = (title, length) => {
         .join('');
     return Number.isNumeric(length) ? id.slice(0, length).padEnd(length, '0') : id;
 };
+
+export const rollCommandToJSON = text => {
+    try {
+        return JSON.parse(`{${text.replaceAll(' ', ',').replace(/(\w+(?==))(=)/g, '"$1":')}}`);
```
WBHarry: This can probably be done better or more neatly.
It modifies a string with this sort of formating: 
`hope="d20" attribute="agility" advantage=true`
-->
`{ "hope": "d20", "attribute": "agility", "advantage": true }`

Doing this so we can parse the parameter contents of a rollCommand straight to JSON. The benefit being that it easily handles the user inputing the possible parameters in any order.

cptn-cosmo:
> Currently this is done for ChatMessages, Journals and Applications.

Does this mean it is not available yet for compendium items/features?

WBHarry:
> > Currently this is done for ChatMessages, Journals and Applications.
> 
> Does this mean it is not available yet for compendium items/features?

Items and Feature sheets are Applications, so it should handle them there. I haven't tried yet though.
I can check :grin:

module/helpers/utils.mjs
emiliojc: a bit of chatgpt magic:

```js
/**
 * Parse a space-delimited list of key=value pairs (with optional
 * double-quoted values) into a plain object, coercing booleans and numbers.
 *
 * @param {string} text
 * @returns {{ [key: string]: string|number|boolean }|null}
 *
 * @example
 * rollCommandToJSON('hope="d20" attribute="agility" advantage=true')
 * // → { hope: "d20", attribute: "agility", advantage: true }
 *
 * @see {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll String.prototype.matchAll}
 * @see {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number Number}
 */
function rollCommandToJSON(text) {
  // Match key="quoted string"  OR  key=unquotedValue
  const PAIR_RE = /(\w+)=("(?:[^"\\]|\\.)*"|\S+)/g;
  const result = {};
  for (const [, key, raw] of text.matchAll(PAIR_RE)) {
    let value;
    if (raw.startsWith('"') && raw.endsWith('"')) {
      // Strip the surrounding quotes, un-escape any \" sequences
      value = raw.slice(1, -1).replace(/\\"/g, '"');
    } else if (/^(true|false)$/i.test(raw)) {
      // Boolean
      value = raw.toLowerCase() === 'true';
    } else if (!Number.isNaN(Number(raw))) {
      // Numeric
      value = Number(raw);
    } else {
      // Fallback to string
      value = raw;
    }
    result[key] = value;
  }
  return Object.keys(result).length > 0 ? result : null;
}
```

emiliojc: What’s improved
- Regex clarity
    We explicitly capture two cases:
        "quoted string" (allowing escaped quotes)
        unquoted tokens (\S+)

- Type coercion
        Booleans via /^(true|false)$/i
        Numbers via Number(raw) + isNaN check

- No thrown errors
    We never call JSON.parse, so a malformed pair simply isn’t matched, and we return null only if no pairs are found.

- Maintainability
    Each transformation step is isolated and documented, making future tweaks (e.g. supporting single-quotes) straightforward.

emiliojc:
The total is not correctly displayed, seems like it's not adding up hope + fear dice.
image

WBHarry:
> The total is not correctly displayed, seems like it's not adding up hope + fear dice.

Oh, it's not supposed to. It's a Duality Roll after all. Either the Fear dice is used, or the Hope dice.
In a way that makes the info on the right meaningless in this case, but I just went with displaying it like a standard roll at the moment.

cptn-cosmo:
As discussed on discord, this should be changed to add up the hope and fear dice :grimacing:

WBHarry:
Added some styling to make Hope/Fear results more obvious.
There's color variables for `hope`, `fear` and `critical` to mess around with. I moved it over to gold, blue and purple. Something to iterate on.
image

WBHarry:
Hope and Fear is added together for Total and DiceTotal now. :Grin:
image

module/helpers/utils.mjs
WBHarry: Went ahead and added this in with the addition that an empty string returns `{}` so it works with the existing logic. :+1:

emiliojc:
don't merge so @moliloo gives the final thumbsup with design stuff (though I see everyone agrees in discord 😂 )

WBHarry:
@moliloo Final look if anything should be changed right away 😁 
Added a `DualityRollColorScheme` system setting.

**Colorful**
image

**Normal**
image

moliloo:
> @moliloo Final look if anything should be changed right away 😁 Added a `DualityRollColorScheme` system setting.
@WBHarry 

I would recommend decrease the saturation from colors you use on Colorful ChatMessage, for some reason, my eyes hurt a bit when a see for a short time bright colors like this, but if most of the team agree if this in discord, you can go ahead with pr

PR #54, opened 2025-05-25
feat(subclassFeatures): fix adding abilities

emiliojc:


WBHarry:
Actually, wait. We lost something here.
Now it only works for FoundationFeatures because the rest of the matching was removed.

WBHarry:
Think you misstakenly copy pasted the if cases :grin:
```
if (event.currentTarget.classList.contains('foundation-tab')) featureField = 'foundation';
        else if (event.currentTarget.classList.contains('foundation-tab')) featureField = 'foundation';
        else if (event.currentTarget.classList.contains('foundation-tab')) featureField = 'foundation';
```

emiliojc:
fixed! great catch, that happens for no tests + only checking on foundation 😂

PR #55, opened 2025-05-25
Review Inventory

### LINKED ISSUE #29: Review inventory/equipment in PC
PCs currently can only hold 2 sets of weapons, 1 active, 1 "inventory weapon". They can only carry 1 armor, which is equipped.

The inventory tab only holds `consumable` and `miscelaneous`.

## Goal
Let's have a more robust inventory where you can put whatever you want, and "equipped weapon/armor".

WBHarry:
Closes #29 
- Weapons and armor are now stored like any other item on the PC, in the foundry standard `Items` embeddedCollection.
- Weapons and armors are considered equipped to the PC if their boolean `equipped` property is set.

When equipping a new weapon, currently equipped weapons will be automatically unequipped if neccessary. EX:
1) Broadsword and Shortsword equipped
2) Click to equip Battleaxe (2 handed)
3) Broadsword and Shortsword are unequipped since both hands are needed and Battleaxe is equipped.
4) Profit


module/applications/sheets/pc.mjs
```diff
@@ -1160,23 +1202,29 @@ export default class PCSheet extends DaggerheartSheet(ActorSheetV2) {
...
-                            ? await fromUuid(this.document.system.activeWeapons.primary.uuid)
-                            : this.document.system.activeWeapons.secondary && item.system.secondary
-                              ? await fromUuid(this.document.system.activeWeapons.secondary.uuid)
+                        this.document.system.equippedWeapons.primary && !item.system.secondary
+                            ? await fromUuid(this.document.system.equippedWeapons.primary.uuid)
+                            : this.document.system.equippedWeapons.secondary && item.system.secondary
+                              ? await fromUuid(this.document.system.equippedWeapons.secondary.uuid)
                               : null;
                     await existing?.delete();
                     itemData.system.active = true;
```
emiliojc: Drag and drop doesn't equip a weapon/armor. It adds to the inventory but looks like it's missing the equipped (like here has `active`).

module/applications/sheets/pc.mjs
```diff
@@ -1160,23 +1202,29 @@ export default class PCSheet extends DaggerheartSheet(ActorSheetV2) {
...
 
                     const existing =
-                        this.document.system.activeWeapons.primary && !item.system.secondary
-                            ? await fromUuid(this.document.system.activeWeapons.primary.uuid)
-                            : this.document.system.activeWeapons.secondary && item.system.secondary
-                              ? await fromUuid(this.document.system.activeWeapons.secondary.uuid)
+                        this.document.system.equippedWeapons.primary && !item.system.secondary
+                            ? await fromUuid(this.document.system.equippedWeapons.primary.uuid)
+                            : this.document.system.equippedWeapons.secondary && item.system.secondary
+                              ? await fromUuid(this.document.system.equippedWeapons.secondary.uuid)
```
emiliojc: We should keep in mind that we can equip the same weapon twice. e.g.: Rogue with two daggers.

Can we extract the logic to check the weapon status?

Secondary or Primary weapons should not matter for equipment. We can have equipped 2 one-handed weapons, even if they are 2 daggers or 2 maces.

WBHarry: If I'm not misstaken, then Secondary/Primary does matter.
image

The way I'm reading this, it says explicitly that a character can only have 1 Primary Weapon equipped, regardless of if it's one-handed or not. IE, you can't equip 2 daggers, because it's a primary weapon. You'd have to equip something like 1 Dagger and 1 Small Dagger (secondary).

WBHarry: I went ahead and extracted the logic for Unequipping what's neccessary to go through with an Equip action. It made the code a lot nicer :grin:

module/applications/sheets/pc.mjs
WBHarry: Fixed :+1:

module/applications/sheets/pc.mjs
emiliojc: completely missed that part! thanks man!

WBHarry:
Ah, nope, something's off in the calculation for hands being free there.
I'll fix it up.

Might not be used in the upcoming sheet change, but all the same :tongue:

PR #58, opened 2025-05-26
fix(compendium): remove duplicated entries

emiliojc:


WBHarry:
Hm, the underlying issue isn't fixed though.
If you run `npm run pushLDBtoYML` to store the Pack DB values to JSON, then the duplicates are returned again.

I can have a look at what I missed up with the recursive mapping :eyes:

WBHarry:
I imagine it probably has something to do with both `Armor` and `Weapons` being subfolders to `Item`, and when extracting from `Tier1` it goes wrong because both `Armor` and `Weapons` have `Tier1`. Some faulty path.

PR #59, opened 2025-05-26
Enhace feature sheet and apply new standard style to applications

moliloo:
## What's new
- New application style for sheet when apply the `.dh-style` class
- Light mode in new application mode style
- Add new parchments files as assets
- Refactor feature template folder structure
- Prettify and refactor feature.js sheet to handle more partials
- Add new en langs and refactor to remove capitalize propriety 

### Dark Mode Preview
https://github.com/user-attachments/assets/8195d718-bd40-475d-a5ce-888e06c30d64

### Light Mode Preview
https://github.com/user-attachments/assets/82704e5b-0bfe-43fc-9840-e01e2cc4ced5


WBHarry:
Hm, I think the `elements.less` button style is applying the `padding: 0` property to broadly.

As is:
image

With `padding: 0` removed from the `elements.less` button style
image

WBHarry:
I think the `legend` styling should set `fontWeight: bold´ to make the title stick out more from the text within.
image

WBHarry:
A very optional thing, but something to consider.
Especially in the darkmode it's kind of hard to notice that there is a field there at all until you hover it, especially for an input that doesn't have an icon in it.

If you think it's a good idea, can just add a 1px outline to the standard state of Input and Select. Otherwise feel free to ignore this :smile:

https://github.com/user-attachments/assets/cded135b-1e54-486e-8cc4-1a4b2e4dfc9e

PR #60, opened 2025-05-26
Daggerheart.css changing line-endings on start/checkout

WBHarry:
It's line endings that keep being changed because GIT changes them to Linux format.
I added git settings to always keep it in windows format. Someone who runs on Linux will have to tell us if they get the issue now or not :laugh:

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text eol=crlf


WBHarry:
Actually. Appearantly it still happens. Moving this to `Draft` instead. It's wierd.

WBHarry:
Don't know why this happens. I'm not currently looking into it though, and I'm tired of seeing this issue in Pull Requests, so I'm closing it 👅

PR #65, opened 2025-05-26
Duality Roll Followup

WBHarry:
- The styling ended up broken when DualityRollColor=Normal. Fixed.
- Fixed a few situations where the labels ended up wrong on modifier chips
- Updated Dice So Nice hope/fear/advantage/disadvantage presets
- Changed DualityRollColor setting to be on Client so every player can use their prefered setting.
- Moved the roll total down into the bottom right together with the result label (EX:`14 With Hope`)


emiliojc:
In normal mode rolling for damage opens the rolls.

https://github.com/user-attachments/assets/932cd3a1-d7f2-4440-bf3f-91b5ebcf5c9a

emiliojc:
the preset changes for duality rolls are quite good, but I'd not make them mandatory. If someone wants to change their presets I'd add a config to overwrite those

Not a blocker, though

WBHarry:
> the preset changes for duality rolls are quite good, but I'd not make them mandatory. If someone wants to change their presets I'd add a config to overwrite those
> 
> Not a blocker, though

I went ahead and added a system Appearance Menu 😎 

https://github.com/user-attachments/assets/ef3133f8-02d3-4b6b-b4ba-222cd7907ee1

I didn't apply the colors to the chat messages. I think that's a good follow up task so we can finish this and not have to leave the duality message styling broken :laugh:

WBHarry:
> In normal mode rolling for damage opens the rolls.
> 
>  ChatDuality_edit.mp4

Good catch! Fixed. I forgot to change the class matching the listener.

Issue #89, opened 2025-05-31
FEATURE: GM Fear Display


Dapoulp:
I create this issue to add it to the board, but it's mostly done.
Current looks:
Option 1 - Token (non hover & hover):
![Image](https://github.com/user-attachments/assets/dca5dd0e-8087-40f5-bfd0-1855dad05e4f) ![Image](https://github.com/user-attachments/assets/f3811f97-10cb-469d-97b8-9d6f1f0370d5)

Option 2 - Bar (non hover & hover):
![Image](https://github.com/user-attachments/assets/f0d730d1-c390-4a3d-874b-4ef9f6eb6a1c) ![Image](https://github.com/user-attachments/assets/71b53c76-1bd2-4b51-9646-4a036f600e13)

Settings added:
- Maximum amount of Fear (default 12)
- Fear display style: Token / Bar / Hide

Also, it can be moved where we want.

Tell me if you think about things to improve it.

Dapoulp:
Updated to make it resizable.
If you tried it before, run "game.user.unsetFlag('daggerheart', 'app.resources.position')" in your browser console before

PR #90, opened 2025-05-31
Feature/89 gm fear display

Dapoulp:
Related Issue: https://github.com/Foundryborne/daggerheart/issues/89


Dapoulp:
Updated to be resizable and for light mode

PR #100, opened 2025-06-01
Levelup Remake

### LINKED ISSUE #56: Level Up Rework
- The level up options have changed since the playtest. Going to go over and update.
- Looking for any bugs that have appeared.
- Looking over the way levelups are handled for possible improvements.

WBHarry:
Closes #56 

Remade the levelup process.
It still gets complicated, but it's at least understandable to me now 🙃 

There are a few things I haven't solved yet, but I think it'd be good to try and bring this in to free up the Data Remodel for the PC sheet while solving the below things in a followup issue.

**Current Known Caveats**
- Subclass is not finished. I'd like to have the new datamodel before having a go at it.
- Multiclass is finished in levelup, but I'm not currently saving it to the PC. I'd like to have the new datamodel first.
- Doesn't consider already picked Traits yet --> You can pick the same one again the next level without needing to reach a new Tier. Also a bit bound by the data model.

The remaining Caveats will be handled in #109


WBHarry:
After some discussion, I'm going to retool things a bit.

Instead of allowing you to level from level X --> Y and selecting everything at the same time,
We will allow leveling from X --> Y, and then handling each level separately.
image

There are some things that are very messy to handle UI by allowing someone to cross multiple Tiers in one go (selecting attributes many times because the mark gets cleared upon entering a new tier etc).

------
The idea will be to have an advance arrow that takes you to the next level once you've done all selections. Ultimately you get a `To Summary` arrow where things are summarised and you can accept the level up.

emiliojc:
Highlighting the pending choices, or not allowing to go forward to `Summary` would be good, right now it doesn't show how many Trat advancements I need to select and it can be a bit confusing. (`1/2` next to Increased Traits maybe?)

If you move to summary before selecting everything, you cannot finish leveling up, but also doesn't show any "error".

image

image

emiliojc:
Leveling up multiple times certainly doesn't work well, I was trying to be very mean with the system and the sheet just breaks. Will avoid testing it for now.

emiliojc:
The Summary works weird, it shows previous level up too? I leveled up the character to level 2, then now to level 3.

image

emiliojc:
When selecting Experiences to level up, it doesn't show the experience name, just the obj id. Evasion appears as `undefined` when increasing it. Also I think due to this `undefined`, future level-ups show an increase in evasion, but it's not applied.
image
image
image

emiliojc:
Selecting choices from a previous Tier seems a bit strange in the breakdown. Do we need to show that "tier breakdown" ? Is it mostly to show the user what they have available to select?

image

emiliojc:
These are the only things I could find testing the first few levels. Leveling mostly one by one. Later I'll try to test the rest of the way to 10.

I love the work on it, and everything seems pretty faithful to the game itself. I'd say that the `Summary` needs more work, I'd like to know in the summary from which to which level I'm going for example. And if we want to show a history of all previous breakdowns that sounds very cool.

Something like (just an idea):

Level up from 3 -> 4
You got: 
Threshold x -> y
domain card - S
hit point 5 -> 6

Previous level ups
You got: 
Threshold x -> y
domain card - R, T, Y
Stress 5 -> 6

Dapoulp:
Don't know what you think but doing one level at once isn't simplier ? In term of code and for the user, getting everything all the time feels a bit confusing to me.

WBHarry:
Yeah, I'm midway through redoing that. It's in the PR 😜

Den mån 2 juni 2025 14:29Dapoulp ***@***.***> skrev:

module/applications/levelup.mjs
```diff
@@ -1,371 +1,675 @@
...
+                context.subclassCards = [];
+                if (advancementChoices.subclass?.length > 0) {
+                    for (var subclass of possibleSubclasses) {
+                        const data = await foundry.utils.fromUuid(subclass.uuid);
+                        const selected = selectedSubclasses.some(x => x.uuid === data.uuid);
+                        context.subclassCards.push({
+                            ...data.toObject(),
+                            uuid: data.uuid,
+                            // disabled:
+                            //     !selected && subclassSelections.length === context.advancementChoices.subclass.length,
```
moliloo: Did you still need this lines? If no, pls, remove later

module/applications/sheets/pc.mjs
```diff
@@ -167,13 +167,26 @@ export default class PCSheet extends DaggerheartSheet(ActorSheetV2) {
...
+            .forEach(element => element.addEventListener('change', this.onFeatureInputBlur.bind(this)));
+        htmlElement
+            .querySelectorAll('.experience-description')
+            .forEach(element => element.addEventListener('change', this.experienceDescriptionChange.bind(this)));
+        htmlElement
+            .querySelectorAll('.experience-value')
+            .forEach(element => element.addEventListener('change', this.experienceValueChange.bind(this)));
+        // htmlElement
+        //     .querySelectorAll('[data-item]')
+        //     .forEach(element => element.addEventListener.on('change', this.itemUpdate.bind(this)));
```
moliloo: remove unnecessary comments

module/helpers/utils.mjs
```diff
@@ -131,3 +132,94 @@ export const setDiceSoNiceForDualityRoll = (rollResult, advantage, disadvantage)
...
+        whitelist: Object.keys(options).map(key => {
+            const option = options[key];
+            return {
+                value: key,
+                name: game.i18n.localize(option.label),
+                src: option.src
+            };
+        }),
+        maxTags: maxTags,
+        // callbacks: { invalid: this.onAddTag },
```
moliloo: This callback will be used later?

styles/less/global/elements.less
```diff
@@ -288,3 +288,115 @@
         }
     }
 }
+
+.theme-light {
+    .application {
+        .component.dh-style {
+            &.card-preview-container {
```
moliloo: i think here theres no need to make such nesting. you can make something like `.theme-light.application.component.dh-style.card-preview-container {}`

styles/less/global/elements.less
```diff
@@ -288,3 +288,115 @@
...
+                    color: var(--color-light-5);
+                }
+            }
+        }
+    }
+}
+
+.application {
+    .component.dh-style {
+        &.card-preview-container {
```
moliloo: the same applies here, i could do something like `.application.component.dh-style.card-preview-container {}`

styles/less/global/feature-section.less
moliloo, line 1: I dont get whats happen here, prettier change something here?

styles/less/global/item-header.less
moliloo, line 1: I dont get whats happen here, prettier change something here?

styles/less/global/sheet.less
moliloo, line 1: I dont get whats happen here, prettier change something here?

styles/less/global/tab-actions.less
moliloo, line 1: I dont get whats happen here, prettier change something here?

styles/levelup.less
```diff
@@ -0,0 +1,261 @@
+.theme-light {
+    .daggerheart.levelup {
```
moliloo, line 2: you could use `.theme-light.daggerheart.levelup {}` to avoid nest to much

styles/levelup.less
```diff
@@ -0,0 +1,261 @@
...
+                        position: absolute;
+                        text-align: center;
+                        top: 4px;
+                        background: grey;
+                        padding: 0 12px;
+                        border-radius: 6px;
+                    }
+
+                    img {
+                        height: 124px; // Can it be dynamically sized? Won't follow any window resizing like this.
```
moliloo: We can think in resizing later, but height will change dynamically only with values like 100%

styles/resources.less
```diff
@@ -1,46 +1,49 @@
 :root {
-    --primary-color-fear: rgba(9, 71, 179, .75);
-    --secondary-color-fear: rgba(9, 71, 179, .75);
+    --primary-color-fear: rgba(9, 71, 179, 0.75);
+    --secondary-color-fear: rgba(9, 71, 179, 0.75);
     --shadow-text-stroke: -1px -1px 0 #000, 1px -1px 0 #000, -1px 1px 0 #000, 1px 1px 0 #000;
-    --fear-animation : background .3s ease, box-shadow .3s ease, border-color .3s ease, opacity .3s ease;
+    --fear-animation: background 0.3s ease, box-shadow 0.3s ease, border-color 0.3s ease, opacity 0.3s ease;
```
moliloo: Please, create var colors in `colors.less`, them you import the file to this file

styles/resources.less
```diff
@@ -1,46 +1,49 @@
...
+        padding: 0.5rem;
         #resource-fear {
             display: flex;
             flex-direction: row;
-            gap: .5rem 0.25rem;
+            gap: 0.5rem 0.25rem;
             flex-wrap: wrap;
             i {
                 font-size: var(--font-size-18);
                 // flex: 1 1 calc(25% - 0.25rem);
```
moliloo: I think we can remove this commented line

templates/views/levelup/parts/multiclass-preview-card.hbs
moliloo, line 1: There is a reason to not reuse `templates/components/card-preview.hbs` and create another similar template?

templates/views/levelup/parts/selectable-card-preview.hbs
moliloo, line 1: There is a reason to not reuse templates/components/card-preview.hbs and create another similar template?

styles/levelup.less
WBHarry: We'll revisit this at a later time then :+1:

styles/resources.less
WBHarry: This was some prettier thing going on, because this is the Fear Tracker from another issue.
But, I went ahead and moved it over 😉

templates/views/levelup/parts/multiclass-preview-card.hbs
WBHarry: I removed `multiclass-preview-card.hbs` by moving the extra data-properties out to a parent component.

templates/views/levelup/parts/selectable-card-preview.hbs
WBHarry: I don't think reusing will work here. They do have some notable differences in design. 
But, it's used for subclasses - and subclasses is something that we will revisit after the DataModel redesign, so it will definately be touched then.

module/helpers/utils.mjs
WBHarry: Just some legacy left from an implementation in one of my modules I took the tagify base from :grin:

styles/resources.less
Dapoulp: Just a little thing, there is no need to add a "0" in css when you've a value like "0.75"

WBHarry: Ah, I just pulled it in.

But yes. It's not a choice I made. It was prettier changing it around.

Dapoulp: ah ok

PR #108, opened 2025-06-06
Combat and CombatTracker

### LINKED ISSUE #105: Combat/CombatTracker Overview
There are existing classes for Combat and the CombatTracker, but they were made for Application V1 so they currently don't work.
They were also probably making use of Playtest rules that got changed.

Going to update them to V2 and give a look at what to keep and what not to. Opening up a discord thread once things can be properly looked at.

WBHarry:
Closes #105 

I've added in a Daggerheart Combat and CombatTracker class with functionality pertinent to the system.

### DIFF
--- a/styles/daggerheart.css
+++ b/styles/daggerheart.css
@@ -1318,6 +1318,9 @@
 .combat-sidebar .encounter-controls.combat .control-buttons {
   width: min-content;
 }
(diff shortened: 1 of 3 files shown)


WBHarry:
Final layout after a lot of discord finetuning :grin:
image

PR #111, opened 2025-06-07
110 - Class Data Model

### LINKED ISSUE #110: Class Data Model
> - Add logic in the preCreate to set the Class or Multiclass field in the Actor.
> - Add logic in the onDelete to unset the Class or Multiclass field in the Actor.

I'm thinking I'll add a Class and Multiclass `ForeignDocumentUUIDField` to the `PC` datamodel.
These fields will then be filled/emptied in `preCreate`/`onDelete` of the Class Item.

WBHarry:
Closes #110 

Added PreCreate/Create/Delete logic for Class/Subclass and set it as ForeingUUIDDocument fields in PC.
Should I do some logic for adding/removing Class/Subclass features from the PC aswell when the class/subclass is removed? 🤔

### DIFF
--- a/module/data/item/subclass.mjs
+++ b/module/data/item/subclass.mjs
@@ -1,3 +1,4 @@
+import ForeignDocumentUUIDField from '../fields/foreignDocumentUUIDField.mjs';
 import BaseDataItem from './base.mjs';
 
 export default class DHSubclass extends BaseDataItem {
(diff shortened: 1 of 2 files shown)


module/documents/item.mjs
joaquinpereyra98, line 1: it would be beneficial to move the `preCreate` logic from Documents to the `preCreate` method of TypeDataModel, and similarly, transfer the logic of `_onCreateOperation` into TypeDataModel’s `_onCreate` method.
and, transfer the logic of `_onDeleteOperation` into TypeDataModel’s `_onDelete` method.

WBHarry: > it would be beneficial to move the `preCreate` logic from Documents to the `preCreate` method of TypeDataModel, and similarly, transfer the logic of `_onCreateOperation` into TypeDataModel’s `_onCreate` method. and, transfer the logic of `_onDeleteOperation` into TypeDataModel’s `_onDelete` method.

Awesome, wasn't aware you could define that per DataModel.
I moved the logic over to `Class.mjs` and `Subclass.mjs` TypedDataModels :+1:

WBHarry:
The basic handling is done then.

Haven't done any logic for the Class/Subclass Features since it's a bit up in the air atm.

PR #122, opened 2025-06-09
Added a basic BaseDataActor data model

### LINKED ISSUE #121: BaseActorModel
> - metadata: I consider that the definition of the static metadata getter is useful in the project.
> - Define common field for the actors types.
> - Definition of getRollData, especially useful if we want the data used for rolls to vary according to the type of actor.
> - It might be advisable to create a description schema containing all descriptions. ( description, background, appearance, connections, toneAndFeel, potentialAdversaries, motivesAndTactics)
> Note: Do not use HTMLfields in non-static fields(ArrayField, SetFields, TypedObjectField), it is advisable to keep a single field and make the distinction in the content.

Adding a common BaseActorModel for Character/Adversary and Environment to extend.

WBHarry:
Closes #121 

- Gathered all actor data models in the `actor` folder, exporting them under `_module.mjs`.
- I set up `BaseDataActor` and made Character/Adversary/Environment extend it.
- Metadata for label and type.
- The only common field between them is `description`, which is now defined on `BaseDataActor`.

Not really sure if there is much else to define. I don't see the purpose in extending `getRollData` for the different actors.
I don't think all HTML fields need to be grouped under one schema field.

### DIFF
--- a/module/data/actor/base.mjs
+++ b/module/data/actor/base.mjs
@@ -1,11 +1,11 @@
 /**
  * Describes metadata about the actor data model type
- * @typedef {Object} ItemDataModelMetadata
+ * @typedef {Object} ActorDataModelMetadata
  * @property {string} label - A localizable label used on application.
  * @property {string} type - The system type that this data model represents.
  */
 export default class BaseDataActor extends foundry.abstract.TypeDataModel {
-    /** @returns {ItemDataModelMetadata}*/
+    /** @returns {ActorDataModelMetadata}*/
     static get metadata() {
         return {
             label: 'Base Actor',
@@ -27,8 +27,8 @@ export default class BaseDataActor extends foundry.abstract.TypeDataModel {
      * @param {object} [options] - Options which modify the getRollData method.
      * @returns {object}
      */
-    getRollData(options = {}) {
-        const data = this.getRollData();
+    getRollData() {
+        const data = { ...this };
         return data;
     }
 }


module/data/actor/base.mjs
```diff
@@ -0,0 +1,34 @@
...
+            description: new fields.HTMLField({ required: true, nullable: true })
+        };
+    }
+
+    /**
+     * Obtain a data object used to evaluate any dice rolls associated with this Item Type
+     * @param {object} [options] - Options which modify the getRollData method.
+     * @returns {object}
+     */
+    getRollData(options = {}) {
```
joaquinpereyra98: Critical Error: infinite recursion.
remplaze by:
```js
const data = { ...this  };
return data;
```

module/data/actor/base.mjs
```diff
@@ -0,0 +1,34 @@
+/**
+ * Describes metadata about the actor data model type
+ * @typedef {Object} ItemDataModelMetadata
+ * @property {string} label - A localizable label used on application.
+ * @property {string} type - The system type that this data model represents.
+ */
+export default class BaseDataActor extends foundry.abstract.TypeDataModel {
+    /** @returns {ItemDataModelMetadata}*/
```
joaquinpereyra98: Minor: change `ItemDataModelMetadata` to `ActorDataModelMetadata`

module/data/actor/base.mjs
WBHarry: Hah, yeah I should have checked that. Nice catch.
Fixed :grin:

module/data/actor/base.mjs
WBHarry: Fixed :+1:

PR #128, opened 2025-06-12
Chat Message Fixing

WBHarry:
Fixed so the different chat themes work again


module/documents/actor.mjs
Dapoulp, line 1: L.147: Should be `advantage = config.event.altKey ? true :  config.event.ctrlKey ? false : null,`

PR #134, opened 2025-06-13
Enhancement/merge duality chat tpl

Dapoulp:


WBHarry:
Apart from that, it's all looking good. Great to get rid of the reload 👍

Dapoulp:
I was wondering about the discard if it was forgottent.
I'll remove it. 

I forgot, when does the discard class you be applied ?

EDIT: I'm looking at the prvious templates, the CSS rules was to remove the filter from the discarded css class, but why just not adding the class ?

WBHarry:
> I was wondering about the discard if it was forgottent. I'll remove it.
> 
> I forgot, when does the discard class you be applied ?
> 
> EDIT: I'm looking at the prvious templates, the CSS rules was to remove the filter from the discarded css class, but why just not adding the class ?

Don't think there was a reason. For Duality rolls specifically, I think the class should simply not be added.
ATM it's applied on every Hope/Fear roll to the lower dice.
It -should- be applied to Adversaries that roll with Advantage/Disadvantage, since they are actually doing `kh`/`kl` and dropping one dice.

PR #172, opened 2025-06-24
[FEATURE] #149 - New style for character sheets

### LINKED ISSUE #149: [Feature] Rework in Character Sheet
Rework the Character Sheet based on [Figma Mockup](https://www.figma.com/design/9RQB0pnFkM4995fzAmAW2k/Daggerheart?node-id=0-1&t=mdVHJTQLQUS7b3z3-1)

moliloo:
closes #149

### DIFF
--- a/styles/daggerheart.css
+++ b/styles/daggerheart.css
@@ -3469,6 +3469,9 @@ div.daggerheart.views.multiclass {
 .application.sheet.daggerheart.actor.dh-style.character .character-header-sheet .name-row input[type='text']:hover {
   outline: 2px solid light-dark(#222, #f3c267);
 }
+.application.sheet.daggerheart.actor.dh-style.character .character-header-sheet .name-row .level-div {
+  white-space: nowrap;
+}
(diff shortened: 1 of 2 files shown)


Dapoulp:
Just a detail: In Character Sheet, we should add "white-space: nowrap" on the Level label, so it doesn't switch to 2 lines when user reduces the sheet width.

EDIT: Buttons to switch items list view doesn't work right now ?

moliloo:
> Buttons to switch items list view doesn't work right now ?
Yes, atm switch buttons doesn't work, but the views are ready

WBHarry:
We could go ahead and just add in the mentioned `nowrap` styling. Otherwise I think we're all good to pull it in, so we can start and work on the separate parts to wire things up 😄

moliloo:
If `nowrap` style is the only thing remaining to merge this branch, i will suggest someone add this for me, at this moment i can't do that because of my work and i don't want hold a merge for hours because of one style

PR #190, opened 2025-06-28
Fix Class HitPoints

WBHarry:
Added HitPoints on class for use on character. We were still using the old system where all classes had the same Hit Points.
- There is now a field on class to enter the value.
- It's used on character in lieu of the previous static 6 base.
- Went ahead and also randomly updated the DomainConfig to use the official domain icons, since they're now available.


moliloo (approved):
Ready to merge, no needed squash this commits

PR #193, opened 2025-06-28
Irk/83 homebrew currency

### LINKED ISSUE #83: (Homebrew) Custom Currency Name and Denominations
_"If it's okay if I make a purely personal request- I would really love the ability to rename the currency. Both what the denominations are called and what the currency is called. Not every setting uses Gold being able to call it "stacks of bills" instead of "handfuls of gold" or "bags of rupees" or "crates of bottlecaps" would really help the flexibility"_

Posted by a user on Discord.
Low priority, but could be neat.

IrkTheImp:
Should now be able to change the default name categories for currency. 

For some reason on the reset function of the homebrew settings it didn't want to localize all the strings, so I have it call a recursive function to localize any strings that need it. 

Not sure if we actually need or plan to use the gold.hbs anymore, but just in case I updated it as well. 

close #83

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -101,6 +101,7 @@
                     "traitArray": { "label": "Initial Trait Modifiers" }
                 },
                 "Currency": {
+                    "enabled": "Enable Overrides",
                     "title": "Currency Overrides",
                     "currencyName": "Currency Name",
                     "coinName": "Coin Name",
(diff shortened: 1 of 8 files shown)


moliloo:
we will not use gold.hbs anymore, i just forget to delete the legacy templates

PR #228, opened 2025-07-01
[Feature] 155 - New Adversary Sheet

### LINKED ISSUE #155: [Feature] Adversary Sheet Rework

moliloo:
## Description

- Refactor Adversary Class Sheet
- New Templates Sheets for Adversary Class Sheet
- Add Adversary- Seetings Application
- Add new fields to `en.json`

- close #155 

## Type of Change

- [x] New feature

## How Has This Been Tested?

- [x] Manual testing

chrome_NAOOMqDnht

## Checklist

---


WBHarry:
I can't make a review since I added to the PR, but:

I think we should move the `Add Buttons` for experience and actions into the Fieldset legend as `+` buttons as we do in many other places.
image

WBHarry:
Did a little push for some small things I felt would be more complicated to explain than fix 😁 
- The Notes page had enriched set to the wrong property, so it didn't show any text when not in edit mode.
- The Adversary Settings application had the same ID set as the Actions Config, so when you opened an Action, the Settings application was force closed. I removed the ID since you could have multiple Adversary Settings applications open at the same time.

WBHarry:
Added so that UseItem (includes attack use) and ToChat works.

Since we have a dedicated edit view, should we remove the context menu from attack/actions in the Adversary Sheet? 🤔

WBHarry:
The damage and effects sections don't do anything atm.
Should we simplify this down to just a damage string field since this is the `Simple Attack`?
image

moliloo:
> I can't make a review since I added to the PR, but:
> 
> I think we should move the `Add Buttons` for experience and actions into the Fieldset legend as `+` buttons as we do in many other places.

Hmm, i dont know, since this is Modal ~~aka Application~~ i think we can have more clear action buttons, and we have more space to do this

moliloo:
> The damage and effects sections don't do anything atm. Should we simplify this down to just a damage string field since this is the `Simple Attack`?

Yes, definitely, i forget to remove these fields

WBHarry:
> > I can't make a review since I added to the PR, but:
> > I think we should move the `Add Buttons` for experience and actions into the Fieldset legend as `+` buttons as we do in many other places.
> 
> Hmm, i dont know, since this is Modal ~aka Application~ i think we can have more clear action buttons, and we have more space to do this

Alright, then I'm good with keeping it this way :+1:

Aside from the Damage/Effects and the removal of the context-menu button as mentioned above, I think everything seems to work.

WBHarry:
The Reaction roll ends up adding the `Difficulty` instead of being a straight 1d20, but that's a separate issue we should make another Issue for.

moliloo:
> Aside from the Damage/Effects and the removal of the context-menu button as mentioned above, I think everything seems to work.

Okay, i will remove the Damage and Effects fieldsets later

PR #237, opened 2025-07-01
229 - Narrative Countdown Window Update

### LINKED ISSUE #229: [BUG] - Narrative Countdown Window improvements
**Describe the bug**
Currently, when you create a Narrative Countdown, they only appear in the narrative countdown window. 
while that Window can be minimized, it does retain it's window frame at all times (Preferably it would behave like the fear tracker). In addition, when minimized the window can't be dragged to the bottom of the canvas, as even in the minimized view the window handling assumes the full window size.

**To Reproduce**
1. Add a narrative countdown
2. minimize the narrative countdown window
3. try moving the narrative countdown window to the bottom section of  Foundry

**Expected behavior**
The minimized window should be draggable to the bottom of the Foundry window.
ideally the window frame should disappear like with the fear token window.

WBHarry:
## Description
- Changed so that the countdown application doesn't make use of Minimized mode anymore.
  Instead there's a header button where a user with permission can switch the view from `Simple` to `Edit`. 
- The application can be moved around and resized as expected (fixes the issue of not being able to put it at the bottom)
- The Narrative Countdown and Encounter Countdown applications store their own user flags to save their position and mode from last useage. 

- Closes #229 

## Type of Change
- [x] Bug fix
- [ ] New feature
- [x] Code cleanup/refactor
- [ ] Documentation update
- [ ] Test coverage
- [ ] Dependency update
- [ ] Configuration change

## How Has This Been Tested?

- [x] Manual testing
- [ ] Other:


module/applications/roll.mjs
```diff
@@ -54,7 +54,7 @@ export class DHRoll extends Roll {
     }
 
     static async buildPost(roll, config, message) {
-        console.log(config)
+        console.log(config);
```
moliloo: Remove dev `console.log` debugers

moliloo (approved):
Its good to merge now

PR #244, opened 2025-07-02
Bug - Unused Handlebar Preloads

WBHarry:
Removed non-existing template preload. Annoying error that was always showing up :grin:


Ikraik (approved):
approving

PR #245, opened 2025-07-02
Item Tooltip Expansion

WBHarry:
Added DhTooltipManager aswell as placeholder rendering for Armor/Weapon/DomainCard.
This is mostly a baseline so that designs can be worked on, and potential extra functionality we want identified so we can expand on the Manager as needed..


moliloo (approved):
For me its good to merge

PR #248, opened 2025-07-02
[Bug] Potential Adversary When Removed

WBHarry:
Fixed so edit/view/delete of potential adversary is handled if the adversary Actor has been removed from the world

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -20,6 +20,11 @@
         }
     },
     "DAGGERHEART": {
+        "UI": {
+            "notifications": {
+                "adversaryMissing": "The linked adversary doesn't exist in the world."
+            }
+        },
         "Settings": {
             "Menu": {
                 "Automation": {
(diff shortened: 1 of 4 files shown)


lang/en.json
```diff
@@ -403,6 +403,9 @@
                     "label": "Event",
                     "description": ""
                 }
+            },
+            "Notifications": {
+                "AdversaryMissing": "The linked adversary doesn't exist in the world."
             }
```
moliloo: - Would be better set all notifications messages outsite proprieties, keeping the path `DAGGERHEART.Notifications.adversaryMissing`, or even `DAGGERHEART.UI.notifications.adversaryMissing` and we can set the tooltip messages inside UI as well

- change `AdversaryMissing` to `adversaryMissing`, like i said to you before, keeping the proprieties in lowercase

moliloo (commented):
Beside my lang review, the rest of commit is good to merge to me

lang/en.json
WBHarry: Changed to `DAGGERHEART.UI.notifications.adversaryMissing`.
I've been a bit lazy with the lang file, since we do have a task to restructure it later. Bad me :grin:

moliloo (approved):
Its good to merge

PR #250, opened 2025-07-03
219-fix weapon action damage and display of calculation.

### LINKED ISSUE #219: Weapon Damage Finetuning
Weapons can have a bonus to their damage, so it's not just a dice.
Image

Atm we do have a string field where you can write all of it, but it doesn't seem to bite. In the picture below, the first message is from using the weapon with a damage of `d6 + 2` on the weapon, the second is just `d6`. The first one breaks as no damage button gets presented.
Image

We should also make sure the damage formula shown in the sidebar of the Character sheet reflects the actual damage. In this case it should be <Proficiency>d6 + 1 + 2 (or just <Proficiency>d6 + 3)
Image

IrkTheImp:
## Description

Fixed the damage portion of the attack action created from the weapon damage formula doesn't try to put the whole formula into the die field. Updated the damage chat message to display the total modifiers included in the roll for easier to see the calculation to get the total result. 

- Fixes #219 
- Closes #219 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Code cleanup/refactor
- [ ] Documentation update
- [ ] Test coverage
- [ ] Dependency update
- [ ] Configuration change

## How Has This Been Tested?

- [x] Manual testing
- [ ] Other:

Damage Roll

## Checklist

## Additional Comments

---

### DIFF
--- a/module/data/item/weapon.mjs
+++ b/module/data/item/weapon.mjs
@@ -34,7 +34,6 @@ export default class DHWeapon extends BaseDataItem {
             damage: new fields.SchemaField({
                 dice: new fields.StringField({ choices: SYSTEM.GENERAL.diceTypes, initial: 'd6' }),
                 bonus: new fields.NumberField({ nullable: true, initial: null }),
-                value: new FormulaField({ initial: 'd6' }),
                 type: new fields.StringField({
                     required: true,
                     choices: SYSTEM.GENERAL.damageTypes,


module/data/action/action.mjs
Dapoulp, line 1: cleanBaseDamage is not a good solution.
Instead, modify how Weapon damage is configured by making it 2 inputs:
- 1 select with dice types options
- 1 input number for bonus

module/applications/roll.mjs
Dapoulp, line 1: A regex + creating a new async Roll seems too much to me.
Why not simply removing all die result to the total to get the total modifier ?
`config.roll.modifierTotal = config.roll.total - config.roll.dice.reduce((a,c) => a + c.total, 0)`

Dapoulp (changes requested):
Added comments to files

module/applications/roll.mjs
Dapoulp: Also, there is a modifierTotal for the Hit Roll too. Make the getModifierTotal method inside DHRoll Class and call the method with the roll as parameter where we need it.

Dapoulp: Hmmm... The problem with my solution is negative dice. But it also doesn't work with that regex. If the formula is "1d6 - 1d8 + 5", you'll get " -5" as modifier.
We need a more solid modifierTotal method.

Dapoulp: That's why I did:
```js
config.roll.modifierTotal = 0;
        for (let i = 0; i < roll.terms.length; i++) {
            if (
                roll.terms[i] instanceof foundry.dice.terms.NumericTerm &&
                !!roll.terms[i - 1] &&
                roll.terms[i - 1] instanceof foundry.dice.terms.OperatorTerm
            )
                config.roll.modifierTotal += Number(`${roll.terms[i - 1].operator}${roll.terms[i].total}`);
        }
```
in D20Roll postEvaluate method, it works but it's ugly

Dapoulp (approved):
I'll need to use the modifier method outside of config, but I'll change it in my critical damagebranch, so we can accept this PR

Issue #260, opened 2025-07-04
[BUG] - Summon Action should have a Range value


cptn-cosmo:
**Describe the bug**
Several spells that summon something (Like Magic Hand) have a range in which you can summon the actor.

**Expected behavior**
When adding the summon action a range should be specified

Dapoulp:
Fixed when Action Settings Dialog templates got updated

PR #266, opened 2025-07-05
fix: Fix action range display logic to include target and range fields

gotenxds:
Fixes https://github.com/Foundryborne/daggerheart/issues/260

image


Dapoulp:
I'm working on Action Setting Dialog, I moved/splitted some of those files.
I removed the range/target condition because Range is base field found on all actions, so no need to check for it.

Dapoulp:
I'll create the PR with Action Setting style soon. I close this one.

PR #277, opened 2025-07-05
Fix some css & resources select

Dapoulp:
## Description

- Just minor fixes


WBHarry (approved):
:robot: :+1:

PR #288, opened 2025-07-07
Iss195 create buttons in items sent to chat

### LINKED ISSUE #195: [Feature] Actions should have action buttons
action that leads to a action, like actives and reactions, should contain the buttons to do that action

Image

Psitacus:
## Description

Added the functionality to send actions to the chat card as long as they are defined as actions in an item. Did this for all the items that seemed to have the ability to be sent to chat, so anything that can be sent to chat now has a "source" field which is necessary to duplicate the item temporarily, so anyone that has permission to roll the item can roll it, not only the token who has the item in their inventory

- Fixes #(issue)
- Closes #195 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Code cleanup/refactor
- [ ] Documentation update
- [ ] Test coverage
- [ ] Dependency update
- [ ] Configuration change

## How Has This Been Tested?

- [x] Manual testing
- [ ] Other:

roll from chat

## Checklist

## Additional Comments

Didn't get to test summons, beastform, macros as I didn't explore how those work, but they should work similarly if they are based on an item in the sheet

---


module/data/action/baseAction.mjs
```diff
@@ -568,7 +568,11 @@ export default class DHBaseAction extends foundry.abstract.DataModel {
             img: this.img,
             name: this.name,
             description: this.description,
-            actions: []
+            actions: [],
+            source: {
+                actor: this.item?.actor?.uuid ?? this.item?.actor?.id ?? null,
```
joaquinpereyra98: If the actor exists, it should always have a uuid, so there's no need to fall back to id. You can simplify this line to just:
```js
 actor: this.item?.actor?.uuid ?? null
```

module/data/action/baseAction.mjs
```diff
@@ -568,7 +568,11 @@ export default class DHBaseAction extends foundry.abstract.DataModel {
...
+                item: this.item?.uuid ?? this.item?.id ?? null
```
joaquinpereyra98: If the item exists, it should always have a uuid, so there's no need to fall back to id.

module/data/chat-message/abilityUse.mjs
```diff
@@ -8,6 +8,10 @@ export default class DHAbilityUse extends foundry.abstract.TypeDataModel {
             img: new fields.StringField({}),
             name: new fields.StringField({}),
             description: new fields.StringField({}),
+            source: new fields.SchemaField({
+                actor: new fields.StringField({ nullable: true }),
```
joaquinpereyra98: I would recommend the use of `DocumentUUIDField` for better data validation.
```js
 actor: new fields.DocumentUUIDField({ nullable: true, type: "Actor" }),
 item: new fields.DocumentUUIDField({ nullable: true, Type: "Item" }),
```

module/documents/item.mjs
```diff
@@ -124,8 +125,13 @@ export default class DHItem extends foundry.documents.Item {
             img: this.img,
             name: this.name,
             description: this.system.description,
-            actions: []
+            actions: this.system.actions ?? [],
+            source: {
+                actor: this.actor?.uuid ?? this.actor?.id ?? null,
```
joaquinpereyra98: the same comment as above

module/applications/ui/chatLog.mjs
```diff
@@ -279,9 +281,66 @@ export default class DhpChatLog extends foundry.applications.sidebar.tabs.ChatLo
...
+            }
+
+            const sourceItem = await foundry.utils.fromUuid(sourceItemUuid);
+            if (!sourceItem) {
+                ui.notifications.error(game.i18n.localize('DAGGERHEART.UI.Notifications.sourceItemNotFound'));
+                return;
+            }
+
+            // Create a temporary copy of the item with a unique name
+            const tempItemData = sourceItem.toObject();
```
joaquinpereyra98: the whole logic of creating a second item for subsequent deletion is odd and unnecessary.

module/applications/ui/chatLog.mjs
```diff
@@ -279,9 +281,66 @@ export default class DhpChatLog extends foundry.applications.sidebar.tabs.ChatLo
...
+            tempItemData._id = foundry.utils.randomID(); // Give it a new ID
+            const originalName = tempItemData.name;
+            tempItemData.name = `${tempItemData.name} (${foundry.utils.randomID()})`;
+            
+            // Create the temporary item on the selected actor
+            const [tempItem] = await actor.createEmbeddedDocuments('Item', [tempItemData]);
+            
+            // Immediately rename it back to the original name for display purposes
+            await tempItem.update({ name: originalName });
```
joaquinpereyra98: There is no point in creating an item and then updating it

joaquinpereyra98 (changes requested):
The `abilityUseButton `callback function seems unusually structured, I recommend reviewing it.

Also, please merge the main branch to resolve the current conflicts.

module/applications/ui/chatLog.mjs
Psitacus: how do you suggest we do this instead? 

- the chat card is created by a token
- items have a parent field, which seems like it needs to be filed in
- the button on the chat will have to reference back to the actor unless we create a copy. If the person has permission to use the chat card but the current selected token doesn't have the item, the char card action would fail. This is unexpected. Would the preferred behaviour to only allow the original token to have permission to do the action? 

For instance, a common situation that might happen is that the GM is trying to click an action in the chat while having a different actor selected than the original one. If the item isn't present on the token that the GM has currently selected, they wouldn't be able to use the chat card.

The only alternative I can see is to always have the item be related to the original token and all the actions done by the item use the original token as the parent, but this is not the default behaviour in other systems I've seen.

Psitacus: nevermind, I misinterpreted how DnD does this. I confused how they apply damage and stuff with how they roll items. I'll just use the original item and actor for everything.

module/data/action/baseAction.mjs
Psitacus: done

module/data/action/baseAction.mjs
Psitacus: done

module/data/chat-message/abilityUse.mjs
Psitacus: done

module/applications/ui/chatLog.mjs
Psitacus: fixed

module/applications/ui/chatLog.mjs
Psitacus: fixed

Psitacus:
@joaquinpereyra98 , all changes done

PR #292, opened 2025-07-07
284 - Armor/Weapon Feature Improvements

### LINKED ISSUE #284: Armor and WeaponFeatures - Actions/Effects
Both Armor and WeaponFeatures need to have their functionality gone over.
- Add missing Effects and Actions on features.
- Ensure the effect values are properly resolved via our ActiveEffect class. 
My thinking is that a normal `changes` path should look on it's current parent's data model for the value.   
`system.resources.stress.value`
A `changes` path starting with `@` should look at the data model of the item it was transferred from for the value, if such exists.
`@system.tier`

WBHarry:
## Description
- Fixed so that active effects can interpolate their values from the origin item by prefacing with `ITEM.`,
   EX: `ITEM.@system.tier`
- Added missing effects and actions to Armor and Weapon Features.
- Added bonuses and rules needed to fascilitate the effects

- Closes #284 

## Type of Change
- [x] Bug fix
- [x] New feature

## How Has This Been Tested?
- [x] Manual testing
- [ ] Other:


WBHarry:
Seems I managed ot get a lot of the less files back here. Going to have to fix it 🙃

WBHarry:
Or huh, no. Just a lot of prettier fixes I guess?

module/data/action/baseAction.mjs
Dapoulp, line 1: Changes back updates made to l.340 & l.393

module/data/action/baseAction.mjs
```diff
@@ -337,7 +337,7 @@ export default class DHBaseAction extends foundry.abstract.DataModel {
         const resources = config.costs
             .filter(c => c.enabled !== false)
             .map(c => {
-                return { type: c.type, value: (c.total ?? c.value) * -1 };
+                return { type: c.type, value: c.total ?? c.value };
```
Dapoulp: Put the `* -1` back or it add the cost on use instead of removing it

module/data/action/baseAction.mjs
```diff
@@ -389,8 +389,15 @@ export default class DHBaseAction extends foundry.abstract.DataModel {
...
+
+        const resources = this.actor.system.resources;
         return realCosts.reduce(
-            (a, c) => a && this.actor.system.resources[c.type]?.value >= (c.total ?? c.value),
+            (a, c) =>
+                a &&
+                !(
+                    resources[c.type]?.value + (c.total ?? c.value) >
+                    (resources[c.type]?.maxTotal ?? resources[c.type]?.total)
+                ),
```
Dapoulp: We want to compare the cost to the current value, not the current value + cost to the total

module/data/action/baseAction.mjs
WBHarry: Fixes with a check for `maxTotal` to invert logic for stress and potentially other such resources.

Dapoulp (approved):

module/data/action/baseAction.mjs
WBHarry: Fixes with a check for `maxTotal` to invert logic for stress and potentially other such resources.

PR #294, opened 2025-07-08
Beastform Improvements

WBHarry:
Previously applying a Beastform created a Beastform Effect on the character that controlled the transformation.
It wasn't present and editable on the Beastform Item --> You had to make additional effects to actually modify the traits or other things when transforming.

Now that Beastform is exposed on the Beastform Item so you can edit it. The remove button is disabled on the Beastform Effect since it is mandatory to the function.

I also added a Subject Texture filePath field for use with Dynamic Token Rings.

### DIFF
--- a/templates/sheets/global/tabs/tab-effects.hbs
+++ b/templates/sheets/global/tabs/tab-effects.hbs
@@ -17,7 +17,7 @@
                     <span>{{effect.name}}</span>
                     <div class="controls">
                         <a data-action="editDoc" data-type="ActiveEffect" data-doc-id="{{effect.id}}"><i class="fa-solid fa-pen-to-square"></i></a>
-                        <a data-action="deleteDoc" data-type="ActiveEffect" data-doc-id="{{effect.id}}" {{#if effect.mandatory}}disabled{{/if}}><i class="fa-solid fa-trash icon-button {{#if effect.mandatory}}disabled{{/if}}"></i></a>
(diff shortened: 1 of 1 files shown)


templates/sheets/global/tabs/tab-effects.hbs
```diff
@@ -17,7 +17,7 @@
                     <span>{{effect.name}}</span>
                     <div class="controls">
                         <a data-action="editDoc" data-type="ActiveEffect" data-doc-id="{{effect.id}}"><i class="fa-solid fa-pen-to-square"></i></a>
-                        <a data-action="deleteDoc" data-type="ActiveEffect" data-doc-id="{{effect.id}}"><i class="fa-solid fa-trash"></i></a>
+                        <a data-action="deleteDoc" data-type="ActiveEffect" data-doc-id="{{effect.id}}" {{#if effect.mandatory}}disabled{{/if}}><i class="fa-solid fa-trash icon-button {{#if effect.mandatory}}disabled{{/if}}"></i></a>
```
joaquinpereyra98: To condition a diabled attribute you can use the {{disabled}} helper.
```hbs
<button type="submit" {{disabled myValue}}>Submit</button>
```
https://foundryvtt.com/api/v13/functions/foundry.applications.handlebars.disabled.html

WBHarry: Good thing to use, yup.
Swapped it in :+1:

PR #299, opened 2025-07-08
In Front Template

WBHarry:
Daggerheart doesn't make use of the standard Foundry thin cones like D&D for burning hands or the like.
Instead it uses `In Front`. 
image

- Changed the default measureTemplate angle to 180.
- Changed the name and image of "cone" in the UI Left Template Layer to something for `In Front`. For the moment I'm essentially overriding the cone which saves a lot of special handling throughout different applications.
If we later find a need to place out other types of cones from this menu we can change things. 
- I didn't change the `webm` tooltip preview from the standard cone one. This is easily done, but I left it as a later task.
image


moliloo (approved):
Its good to merge to me

PR #300, opened 2025-07-08
Feature/armor stack uses on damage

Dapoulp:


module/applications/dialogs/damageReductionDialog.mjs
```diff
@@ -210,9 +210,17 @@ export default class DamageReductionDialog extends HandlebarsApplicationMixin(Ap
...
+            this.resolve();
         }
 
         await super.close({});
     }
+
+    static async armorStackQuery({actorId, damage}) {
+        return new Promise(async (resolve, reject) => {
+            const actor = await fromUuid(actorId);
+            if(!actor || !actor?.testUserPermission(game.user, CONST.DOCUMENT_OWNERSHIP_LEVELS.OWNER)) reject();
```
joaquinpereyra98: instead of `actor.testUserPermission(game.user, CONST.DOCUMENT_OWNERSHIP_LEVELS.OWNER` you could use `actor.isOwner`.

module/applications/dialogs/damageReductionDialog.mjs
```diff
@@ -210,9 +210,17 @@ export default class DamageReductionDialog extends HandlebarsApplicationMixin(Ap
...
+            new DamageReductionDialog(resolve, reject, actor, damage).render(true);
```
joaquinpereyra98: `.render(true); -> .render({ force: true });`

Dapoulp: render(true) is a shortcut you can use when this is the only option

WBHarry: It is, but it is the V1 way. It will be removed in an upcoming Foundry version.
I'm trying to unlearn this myself :grin:

module/documents/actor.mjs
```diff
@@ -1,9 +1,17 @@
...
+import { emitAsGM, emitAsOwner, GMUpdateEvent, socketEvent } from '../systemRegistration/socket.mjs';
 import DamageReductionDialog from '../applications/dialogs/damageReductionDialog.mjs';
 import { LevelOptionType } from '../data/levelTier.mjs';
 import DHFeature from '../data/item/feature.mjs';
 
+<<<<<<< HEAD
+export default class DhpActor extends Actor {
+    // Return the first active player with the ownership
+    get owner() {
+        const user = this.hasPlayerOwner && game.users?.players.find(p => p.active && this.ownership[p.id ?? ""] === CONST.DOCUMENT_OWNERSHIP_LEVELS.OWNER);
```
joaquinpereyra98: could jsut do
```js
const user = game.users.players.find(u => this.testUserPermission(u, "OWNER") && u.active);
```

module/applications/dialogs/damageReductionDialog.mjs
Dapoulp: You sure about that ? When they plan to remove something it's added as commentary. Here, they only talk about the second _options object

WBHarry: Hm...I could be wrong. Was sure I'd seen that somewhere in the code. Ah well.

Dapoulp: `// TODO change to {force: true} once everything is ApplicationV2`
They haven't done it themself yet xD

joaquinpereyra98: While it's not incorrect to use it this way, it relies on a legacy approach.
For better compatibility with future releases, it's more correct and safer to use the `{ force: true }`

module/documents/actor.mjs
```diff
@@ -1,9 +1,17 @@
 import DamageSelectionDialog from '../applications/dialogs/damageSelectionDialog.mjs';
-import { GMUpdateEvent, socketEvent } from '../systemRegistration/socket.mjs';
+import { emitAsGM, emitAsOwner, GMUpdateEvent, socketEvent } from '../systemRegistration/socket.mjs';
 import DamageReductionDialog from '../applications/dialogs/damageReductionDialog.mjs';
 import { LevelOptionType } from '../data/levelTier.mjs';
 import DHFeature from '../data/item/feature.mjs';
 
+<<<<<<< HEAD
```
WBHarry: Seems to be a merge conflict marker left here :eyes:

module/applications/dialogs/damageReductionDialog.mjs
Dapoulp: Done.
I kept the `this.hasPlayerOwner` so we don't even need to go to the find for non player character

module/applications/dialogs/damageReductionDialog.mjs
Dapoulp: Done

module/documents/actor.mjs
Dapoulp: Done

module/documents/actor.mjs
Dapoulp: Fixed

PR #303, opened 2025-07-08
Bugfix - Companion Levelup Features

WBHarry:
- Fixed so that features gained from companion levelup show up on the character sheet again.
- Changed so that these Feature items are actually embedded onto the character itself, rather than trying to use it from the companions item list. This is much more robust.
- The features list in the levelup data tracks if the levelup feature is in the partner's item list so it can be removed on delevel.


moliloo (approved):
To me its good to merge

PR #312, opened 2025-07-10
[BUG] DiceSoNice fix  

WBHarry:
Fixed DiceSoNice integration again.
- The colour setting method had misstakenly gone away during some rework.
- Some fields had fallen away from DualityDice data model, causing an error during chatMessage rerender.

### DIFF
--- a/module/data/chat-message/dualityRoll.mjs
+++ b/module/data/chat-message/dualityRoll.mjs
@@ -1,19 +1,6 @@
 import DHAdversaryRoll from './adversaryRoll.mjs';
 
 export default class DHDualityRoll extends DHAdversaryRoll {
-    static defineSchema() {
-        const fields = foundry.data.fields;
-
-        return {
-            ...super.defineSchema(),
-            result: new fields.SchemaField({
-                duality: new fields.NumberField({ integer: true }),
-                label: new fields.StringField(),
-                total: new fields.NumberField({ integer: true })
-            })
-        };
-    }
-
     get messageTemplate() {
         return 'systems/daggerheart/templates/ui/chat/duality-roll.hbs';
     }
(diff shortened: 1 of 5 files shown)


module/data/chat-message/dualityRoll.mjs
Dapoulp, line 1: What's the point of those added fields ?
Result is inside the roll property not directly into DHDualityRoll, so it won't be setted at anytime.

module/dice/dualityRoll.mjs
```diff
@@ -142,5 +142,13 @@ export default class DualityRoll extends D20Roll {
...
             label: roll.totalLabel
         };
+
+        const advantageState =
+            config.roll.advantage == this.ADV_MODE.ADVANTAGE
+                ? true
+                : config.roll.advantage == this.ADV_MODE.DISADVANTAGE
+                  ? false
+                  : null;
+        setDiceSoNiceForDualityRoll(roll, advantageState);
```
Dapoulp: I don't think this will works.
D20Roll postEvaluate modify config.roll.advantage, it's not a single value anymore at this time but an object.
Why not just use the current `config.roll.advantage.type` (-1,0 or 1) and update setDiceSoNiceForDualityRoll instead ?

Dapoulp (changes requested):

module/dice/dualityRoll.mjs
WBHarry: I just copy pasted the old thing we used 😁 
Updated to make use of `config.roll.advantage` :+1:

module/data/chat-message/dualityRoll.mjs
WBHarry: Some weirdness that happened last night. It does work without them, so removed.

PR #328, opened 2025-07-12
[Feature] Item Resource Support

WBHarry:
## Description
- I've added the ability for items to have a resource tied to them.
- This can be a type of token, a dice or something else.
- The resource is displayed in the Item's `inventory-item.hbs` display in sheets so it can be incremented and decremented.
- Actions on the Item can use this resource as a cost.

Also added so that consumable items use the same template for increasing and decreasing their quantity.


styles/less/global/elements.less
```diff
@@ -112,17 +112,17 @@
             margin: 5px;
             height: inherit;
             .tag {
-                box-shadow: 0 0 0 1.1em #E5E5E5 inset;
+                box-shadow: 0 0 0 1.1em #e5e5e5 inset;
```
moliloo: I think we have a color for this in `ultils/colors.less` could you change this to the corresponding variações color?

templates/dialogs/dice-roll/resourceDice.hbs
```diff
@@ -0,0 +1,16 @@
+<div>
```
moliloo: Could you change this nester from `<div>` to `<section>`, i recomend avoid use `<div>` since this is a generic element, and section is a better semantic to handle this

templates/sheets/global/partials/inventory-item.hbs
```diff
@@ -1,128 +1,138 @@
...
+                <a class="item-name" data-action="attackRoll">{{item.name}}</a>
+            {{else}}
+                <div class="item-name">{{item.name}}</div>
+            {{/if}}
+            {{#if  (eq type 'weapon')}} 
+                <div class="item-tags">
+                    {{#if isSidebar}}
+                        <div class="item-labels">
+                            <div class="label">
+                                {{!-- {{localize (concat 'DAGGERHEART.CONFIG.Traits.' item.system.attack.roll.trait '.short')}} --}}
```
moliloo: Will we not use trait tag in sidebar?

templates/sheets/global/partials/inventory-item.hbs
```diff
@@ -1,128 +1,138 @@
...
+                                {{localize (concat 'DAGGERHEART.CONFIG.Range.' item.system.attack.range '.short')}}
+                                <span> - </span>
+                                {{item.system.attack.damage.parts.0.value.dice}}{{#if item.system.attack.damage.parts.0.value.bonus}} + {{item.system.attack.damage.parts.0.value.bonus}}{{/if}}
+                                {{!-- ({{localize (concat 'DAGGERHEART.CONFIG.DamageType.' item.system.attack.damage.parts.0.type '.abbreviation')}}) --}}
```
moliloo: I remember we change from label to a icon, so i think we could delete it

templates/ui/chat/resource-roll.hbs
```diff
@@ -0,0 +1,3 @@
+<div class="daggerheart chat resource-roll">
```
moliloo, line 1: I dont get it, is this the header from chat messages? Só i think we can change it to header, right?

moliloo (changes requested):

templates/ui/chat/resource-roll.hbs
WBHarry: This is just content for a chat message, so it's not a whole section or a header. At least at the moment, our other chat message content templates just start with a div.
```js
const cls = getDocumentClass('ChatMessage');
const msg = new cls({
    user: game.user.id,
    content: await foundry.applications.handlebars.renderTemplate(
        'systems/daggerheart/templates/ui/chat/resource-roll.hbs', <-------------
        {
            user: this.actor.name,
            name: this.item.name
        }
    )
});
```

styles/less/global/elements.less
WBHarry: Swapped for @beige :+1

templates/dialogs/dice-roll/resourceDice.hbs
WBHarry: Fixed :+1:

templates/sheets/global/partials/inventory-item.hbs
WBHarry: Was commented out sometime in main. Added it back in :+1:

templates/sheets/global/partials/inventory-item.hbs
WBHarry: Deleted :+1:

templates/ui/chat/resource-roll.hbs
moliloo: Ohh, i see, so you can keep the `<div>`

PR #332, opened 2025-07-13
actors datas max/total rework

Dapoulp:
## Description

- Remove bonus from actors data properties
- Handle properties upgrade from Effects directly on the value/max
- Update itemConfigs to match changes


WBHarry (approved):
Shouldn't be any issues with levelup. 👍

PR #342, opened 2025-07-15
[Fix] Automation Fixes

WBHarry:
## Description
- Split the automation setting for Hope/Fear into two, one for GM and one for Players.
- Removed Countdown automation setting. It was unncessary.
- Removed countdown hooks. Auto update is now handled directly using a static method defined in DhCountdowns.


moliloo (approved):
Its good to merge

PR #355, opened 2025-07-15
354 - ActionToken Visibility

### LINKED ISSUE #354: [BUG] - Players currently only see their own spotlight tokens.
**Describe the bug**
Players currently only see their own spotlight tokens, if using the optional rule to track `Spotlight Token`. In order to self-adjust the players need to see each others' `Spotlight Tokens`. Therefore at least optionally that should be possible (hardly see why not seeing them should be an option even).

**To Reproduce**
1. Enable the optional rule in system configurations.
2. Run a multiplayer test with at least two PCs and at least one of them not owned by the logged in non-GM player. They won't see the other player's `Spotlight Tokens`.

**Expected behavior**
See above, they need to see them.

WBHarry:
Closes #354 
Changed so that players can see each others actionTokens but can't interact with others tokens.

### DIFF
--- a/styles/less/ui/combat-sidebar/token-actions.less
+++ b/styles/less/ui/combat-sidebar/token-actions.less
@@ -2,7 +2,6 @@
     .token-actions {
         align-self: stretch;
         display: flex;
-        align-items: top;
         justify-content: center;
         gap: 16px;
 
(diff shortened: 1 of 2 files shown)


itteerde:
Great. Works just perfect for me/us. Like the love for detail changing the cursor, too.

PR #359, opened 2025-07-16
[Feature/Fixes] Adversary Touchups

WBHarry:
Little grab-bag of things I noticed while checking up on Adversary.
- Non-transferred, active ActiveEffects are now displayed on the canvas token.
image
- Added a setting for Horde Damage (lowered damage on half hit points or lower) to the Adversary attack.
- There's an automation setting for automatically applying this via active effect.
- Fixed light-mode color for D20 dice in roll dialog for adversaries.
- Bunch of localization fixes


moliloo (approved):
To me its good to merge

PR #361, opened 2025-07-16
[Fix] Resource Improvements

WBHarry:
Based on this:
image

Fixed everything, except I'm not sure about the simple resource focus loss. Couldn't reproduce that.

### DIFF
--- a/module/applications/dialogs/resourceDiceDialog.mjs
+++ b/module/applications/dialogs/resourceDiceDialog.mjs
@@ -67,7 +67,7 @@ export default class ResourceDiceDialog extends HandlebarsApplicationMixin(Appli
 
     static async rerollDice() {
         const max = itemAbleRollParse(this.item.system.resource.max, this.actor, this.item);
-        const diceFormula = `${max}d${this.item.system.resource.dieFaces}`;
+        const diceFormula = `${max}${this.item.system.resource.dieFaces}`;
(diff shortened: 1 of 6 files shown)


module/applications/dialogs/resourceDiceDialog.mjs
```diff
@@ -67,7 +67,7 @@ export default class ResourceDiceDialog extends HandlebarsApplicationMixin(Appli
 
     static async rerollDice() {
         const max = itemAbleRollParse(this.item.system.resource.max, this.actor, this.item);
-        const diceFormula = `${max}d${this.item.system.resource.dieFaces}`;
+        const diceFormula = `${max}${this.item.system.resource.dieFaces}`;
```
moliloo, line 70: removing the `d` will not break the dice formula?

WBHarry: Nah, the dieFaces is a string property, and I changed so that it's now selected via a <Select> with these options which we use in other places for dice:

```
export const diceTypes = {
    d4: 'd4',
    d6: 'd6',
    d8: 'd8',
    d10: 'd10',
    d12: 'd12',
    d20: 'd20'
};
```

moliloo: ohh, okay, thats was the last things im unsure about this pr, merging into main rn

PR #367, opened 2025-07-17
[Fix] Downtime Rework

WBHarry:
- Changed Downtime layout
- Downtime Activities are now controlled by properties under `character.system.rules.rest` so that characters can have shortrest and longrest activities available regardless of what sort of downtime they do, depending on Features.

### DIFF
--- a/styles/less/dialog/downtime/downtime-container.less
+++ b/styles/less/dialog/downtime/downtime-container.less
@@ -11,6 +11,10 @@
     font-family: @font-body;
 
     .downtime-container {
+        .activities-grouping {
+            width: 280px;
+        }
+
         .activities-container {
             width: 100%;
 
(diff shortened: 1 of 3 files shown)


styles/less/dialog/downtime/downtime-container.less
```diff
@@ -1,81 +1,71 @@
...
-        .downtime-header {
-            margin: 0;
-            color: light-dark(@dark-blue, @golden);
-            text-align: center;
-        }
+.theme-light .daggerheart.dh-style.views.downtime {
+    .downtime-container .activity-container .activity-selected-marker {
+        background-image: url(../assets/parchments/dh-parchment-dark.png);
+    }
+}
```
moliloo: Can you explain why `dh-parchment-dark` is used in the `.theme-light`?

WBHarry: My idea is to have some additional contrast so you really notice the number.
If we end up disliking it, we can change it later 😄

WBHarry: I am mercurial - one vote swayed me, and I swapped it now 🙃

PR #395, opened 2025-07-22
[Feature] Allow Dice Reroll in ChatMessage

### LINKED ISSUE #171: [QoL][Feature] Action - Allow Dice reroll in Chat Message
Some abilities allow a player to reroll the Hope or Fear Die, same for damage die.
We need to find a way to display that in the Chat Message, with an option to keep the highest or override the previous result

WBHarry:
- Closes #171

Added reroll functionality for Duality Rolls and D20 rolls in chat.

### DIFF
--- a/daggerheart.mjs
+++ b/daggerheart.mjs
@@ -54,7 +54,7 @@ Hooks.once('init', () => {
         DualityDie
     };
 
-    CONFIG.Dice.rolls = [...CONFIG.Dice.rolls, ...[DHRoll, DualityRoll, D20Roll, DamageRoll]];
+    CONFIG.Dice.rolls = [...CONFIG.Dice.rolls, DHRoll, DualityRoll, D20Roll, DamageRoll];
     CONFIG.MeasuredTemplate.objectClass = placeables.DhMeasuredTemplate;
 
     const { DocumentSheetConfig } = foundry.applications.apps;
--- a/module/applications/ui/chatLog.mjs
+++ b/module/applications/ui/chatLog.mjs
@@ -46,20 +46,14 @@ export default class DhpChatLog extends foundry.applications.sidebar.tabs.ChatLo
         html.querySelectorAll('.target-indicator').forEach(element =>
             element.addEventListener('click', this.onToggleTargets)
         );
-        html.querySelectorAll('.advantage').forEach(element =>
-            element.addEventListener('mouseenter', this.hoverAdvantage)
-        );
-        html.querySelectorAll('.advantage').forEach(element =>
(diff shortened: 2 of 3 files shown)


daggerheart.mjs
```diff
@@ -49,6 +49,11 @@ Hooks.once('init', () => {
         DamageRoll: DamageRoll
     };
 
+    CONFIG.Dice.terms = {
+        ...CONFIG.Dice.terms,
+        DualityDie
+    };
+
     CONFIG.Dice.rolls = [...CONFIG.Dice.rolls, ...[DHRoll, DualityRoll, D20Roll, DamageRoll]];
```
joaquinpereyra98: I would appreciate it if you edited this
```js
CONFIG.Dice.rolls = [...CONFIG.Dice.rolls, DHRoll, DualityRoll, D20Roll, DamageRoll];
```

module/applications/ui/chatLog.mjs
```diff
@@ -309,4 +312,29 @@ export default class DhpChatLog extends foundry.applications.sidebar.tabs.ChatLo
 
         action.use(event);
     };
+
+    rerollEvent = async (event, message) => {
```
joaquinpereyra98: why use an arrow notation and when called use the `.call` method?
Why do they use `call` and arrow notation if it is already an instance method?
I would prefer that they not use arrow notation on methods and check when methods like call and bind need to be used.

module/dice/dualityRoll.mjs
```diff
@@ -186,4 +194,53 @@ export default class DualityRoll extends D20Roll {
...
+                    new foundry.dice.terms.Die({
+                        ...term,
+                        faces: term._faces,
+                        results: term.results.filter(x => !x.rerolled)
+                    })
+                ],
+                options: { appearance: {} }
+            };
+            const diceSoNicePresets = getDiceSoNicePresets();
+            switch (target.dataset.type) {
```
joaquinpereyra98:
```js
const type = target.dataset.type;
if (diceSoNicePresets[type]) {
    diceSoNiceRoll.dice[0].options = { appearance: diceSoNicePresets[type] };
}
```

daggerheart.mjs
WBHarry: 👍

module/applications/ui/chatLog.mjs
WBHarry: I removed all the arrow notations and needless `.call` useage to make ya happy

module/dice/dualityRoll.mjs
WBHarry: Nice one 👍

PR #399, opened 2025-07-24
[Fix] Itemlink Redux Revengeance

### LINKED ISSUE #396: [BUG] Feature Linking to Items
The way items like Ancestry are currently keeping track of what items are linked to them is not done very well.
Going to remake it in more robust way.

WBHarry:
Moved the previous `subType` property that previously used to be on Feature to be stored in a data structure on `features` type fields. This should make things much more stable.

- Closes #396


module/applications/sheets/api/base-item.mjs
```diff
@@ -178,26 +178,30 @@ export default class DHBaseItemSheet extends DHApplicationMixin(ItemSheetV2) {
...
         });
     }
 
     /**
      * Remove a feature from the item.
      * @type {ApplicationClickAction}
      */
-    static async #deleteFeature(_, target) {
+    static async #deleteFeature(_, element) {
+        const target = element.closest('[data-item-uuid]');
```
joaquinpereyra98, line 170: This is already done in the getDocFromElement function, I don't know why this line was added.

module/applications/sheets/api/application-mixin.mjs
```diff
@@ -341,7 +341,20 @@ export default function DHApplicationMixin(Base) {
                 {
                     name: 'CONTROLS.CommonEdit',
                     icon: 'fa-solid fa-pen-to-square',
-                    callback: target => getDocFromElement(target).sheet.render({ force: true })
+                    callback: async target => {
+                        const doc = getDocFromElement(target);
+                        const appId = this.element.id;
+                        doc.apps[appId] = this;
+                        const app = await doc.sheet.render({ force: true });
+                        app.addEventListener(
```
WBHarry: Hm..should we really add the current document to the `apps` of every single editable item linked to it just because you open it?

IE: You open an Ancestry with 3 Features and 2 subclasses. With this solution, when you open Feature#1, we set the `apps` to automatically rerender the Ancestry Sheet when updates happen on Feature#1. If we worked with `onFirstRender` and `onClose`, then we'd have to register the Ancestry sheet application in the apps of all the 5 documents right away, and then remove them onClose.

joaquinpereyra98: > If we worked with onFirstRender and onClose, then we'd have to register the Ancestry sheet application in the apps of all the 5 documents right away, and then remove them onClose.

No, You can do the same in these methods, just add a new option with the current document and check if that option exists when the application is rendered it is added.

templates/sheets/items/class/settings.hbs
```diff
@@ -57,7 +57,7 @@
                             <img class="image" src="{{document.system.characterGuide.suggestedSecondaryWeapon.img}}" />
                             <span>{{document.system.characterGuide.suggestedSecondaryWeapon.name}}</span>
                             <div class="controls">
-                                <i data-action="removeSuggestedItem" data-target="suggestedSecondaryWeapon" class="fa-solid fa-trash icon-button"></i>
+                                <a><i data-action="removeSuggestedItem" data-target="suggestedSecondaryWeapon" class="fa-solid fa-trash icon-button"></i></a>
```
joaquinpereyra98: Add action on anchor elements, not on the icon.

module/applications/sheets/api/application-mixin.mjs
WBHarry: I fiddled around a little with the idea of using `_onFirstRender` and `_onClose`. I felt it got difficult.

For my part, I'm good with leaving this as is and having it as something to redo at a later time if someone can make a good implementation.

module/applications/sheets/api/application-mixin.mjs
```diff
@@ -496,7 +524,7 @@ export default function DHApplicationMixin(Base) {
          * Renders an embedded document.
          * @type {ApplicationClickAction}
          */
-        static #editDoc(_event, target) {
+        static async #editDoc(_event, target) {
```
joaquinpereyra98: Made the asynchronous method but didn't await anything. Either remove the async, or await the render methods.

module/applications/sheets/api/application-mixin.mjs
```diff
@@ -341,7 +356,20 @@ export default function DHApplicationMixin(Base) {
                 {
                     name: 'CONTROLS.CommonEdit',
                     icon: 'fa-solid fa-pen-to-square',
-                    callback: target => getDocFromElement(target).sheet.render({ force: true })
+                    callback: async target => {
```
joaquinpereyra98: modify this callback

module/documents/tooltipManager.mjs
```diff
@@ -13,7 +13,7 @@ export default class DhTooltipManager extends foundry.helpers.interaction.Toolti
             if (item) {
                 const type = actionId ? 'action' : item.type;
                 const description = await TextEditor.enrichHTML(item.system.description);
-                for (let feature of item.system.features) {
+                for (let feature of item.system?.features ?? []) {
                     feature.system.enrichedDescription = await TextEditor.enrichHTML(feature.system.description);
```
joaquinpereyra98: This should not take into account whether features is an array of Documents or an array of object type, items?
I've also seen that they localize this on the tmeplate? That doesn't make sense.

WBHarry (commented):
In DamageRoll, `roll.roll` is a function that hasn't bene called if you use a damage action with `ShiftClick`. Errors out.
image

templates/sheets/items/class/settings.hbs
```diff
@@ -42,7 +42,7 @@
                             <img class="image" src="{{document.system.characterGuide.suggestedPrimaryWeapon.img}}" />
                             <span>{{document.system.characterGuide.suggestedPrimaryWeapon.name}}</span>
                             <div class="controls">
-                                <i data-action="removeSuggestedItem" data-target="suggestedPrimaryWeapon" class="fa-solid fa-trash icon-button"></i>
+                                <a><i data-action="removeSuggestedItem" data-target="suggestedPrimaryWeapon" class="fa-solid fa-trash icon-button"></i></a>
```
joaquinpereyra98: Add action on anchor elements, not on the icon.

templates/sheets/items/class/settings.hbs
```diff
@@ -72,7 +72,7 @@
                             <img class="image" src="{{document.system.characterGuide.suggestedArmor.img}}" />
                             <span>{{document.system.characterGuide.suggestedArmor.name}}</span>
                             <div class="controls">
-                                <i data-action="removeSuggestedItem" data-target="suggestedArmor" class="fa-solid fa-trash icon-button"></i>
+                                <a><i data-action="removeSuggestedItem" data-target="suggestedArmor" class="fa-solid fa-trash icon-button"></i></a>
```
joaquinpereyra98: Add action on anchor elements, not on the icon.

templates/sheets/items/class/settings.hbs
```diff
@@ -86,13 +86,15 @@
...
-                            <img class="image" src="{{this.img}}" />
-                            <span>{{this.name}}</span>
-                            <div class="controls">
-                                <i data-action="removeItemFromCollection" data-target="invetory.take" data-uuid="{{this.uuid}}" class="fa-solid fa-trash icon-button"></i>
+                        {{#if this}}
+                            <div class="suggested-item item-line" data-action="editDoc" data-item-uuid="{{this.uuid}}">
+                                <img class="image" src="{{this.img}}" />
+                                <span>{{this.name}}</span>
+                                <div class="controls">
+                                    <a><i data-action="removeItemFromCollection" data-target="inventory.take" data-uuid="{{this.uuid}}" class="fa-solid fa-trash icon-button"></i></a>
```
joaquinpereyra98: Add action on anchor elements, not on the icon.

joaquinpereyra98 (changes requested):

module/applications/sheets/api/base-item.mjs
WBHarry: It's added because we need to know the type in the filter function down below
```js
.filter(x => target.dataset.type !== x.type || x.item.uuid !== feature.uuid)
```

templates/sheets/items/class/settings.hbs
WBHarry: Fixed :+1:

module/applications/sheets/api/application-mixin.mjs
WBHarry: Removed async. Originally I did other stuff in editDoc that needed it, but it wa removed again. :+1:

module/applications/sheets/api/application-mixin.mjs
WBHarry: Should be good now

templates/sheets/items/class/settings.hbs
WBHarry: Fixed :+1:

templates/sheets/items/class/settings.hbs
WBHarry: Fixed :+1:

templates/sheets/items/class/settings.hbs
WBHarry: Fixed :+1:

module/documents/tooltipManager.mjs
WBHarry: This I don't get.
Why does it matter if the data coming in is a document or an object type?

The features are not localized on the template, the description is rendered in the `tooltipTags.hbs` partial.
It's a very flexible render because it's used for Features and Actions from different sorts of data sources.
`<div class="tooltip-tag-description">{{{localize (ifThen feature.description feature.description (ifThen feature.system.enrichedDescription feature.system.enrichedDescription feature.system.description))}}}</div>`

module/applications/sheets/api/base-item.mjs
WBHarry: Could expand `getDocFromElement` to return both the doc and the matched HTMLElement, but eh..it's used in a ton of places, and it makes those implementations messier.

PR #402, opened 2025-07-24
Refactor/actions v2

Dapoulp:
There is still thing I want to do but we can already merge that part and create compendium stuff


module/data/action/baseAction.mjs
```diff
@@ -180,38 +117,14 @@ export default class DHBaseAction extends foundry.abstract.DataModel {
...
-            costs: costsConfig,
-            uses: usesConfig,
-            data: actorData
-        };
+        let config = this.prepareConfig(event);
+        for(let i = 0; i < this.constructor.extraSchemas.length; i++) {
+            let clsField = this.constructor.getActionField(this.constructor.extraSchemas[i]);
+            if(clsField?.prepareConfig) {
+                const keep = clsField.prepareConfig.call(this, config);
+                if(config.isFastForward && !keep) return;
```
WBHarry, line 125: I set up a Feature with a Damage Action and gave it damage - custom formula 1d6.
Using it via the Damage button has the action exited out of via FastForward (after clsField = RangeField).
image

module/documents/item.mjs
```diff
@@ -107,9 +124,9 @@ export default class DHItem extends foundry.documents.Item {
 
     async use(event) {
         const actions = this.system.actionsList;
-        if (actions?.length) {
-            let action = actions[0];
-            if (actions.length > 1 && !event?.shiftKey) {
+        if (actions?.size) {
+            let action = actions.first(0);
```
WBHarry: Using a Feature that has an action on it on either an Adversary or a Character error out here.  actions (`ActionCollection`) does not have method first.

WBHarry (changes requested):

module/data/action/baseAction.mjs
Dapoulp: Fixed

module/documents/item.mjs
Dapoulp: Fixed

module/applications/sheets-configs/action-config.mjs
```diff
@@ -210,8 +193,10 @@ export default class DHActionConfig extends DaggerheartSheet(ApplicationV2) {
 
     static addDamage(event) {
         if (!this.action.damage.parts) return;
-        const data = this.action.toObject();
-        data.damage.parts.push({});
+        const data = this.action.toObject(),
+            part = {};
+        if(this.action.actor.isNPC) part.value = { multiplier: 'flat' };
```
WBHarry: If you make a Feature and add an AttackAction to it and go to edit in the Effect tab - you can't click the `+` to add damage, because `this.action.actor` is undefined.

Dapoulp: Fixed

WBHarry (approved):
:robot: :+1:

PR #408, opened 2025-07-25
Reduce Action Selection click number & Style Action Type Dialog

Dapoulp:
## Description

1. Item Action Selection Dialog: Make Action button, clicking on it directly trigger the action
2. Action Type Selection Dialog: Styling

## How Has This Been Tested?

- [x] Manual testing
- [ ] Other:

image image


Dapoulp:
I'ven't check if icons were free, doing that

WBHarry (changes requested):
For the Action Select, I think we should swap the coloration to be like the Experience chips. We need it more washed out so that action images that are orange/yellow/red:ish in colour won't look weird on the strongly colored button
image

WBHarry:
Becomes something like this
image

Compared to
image

Dapoulp:
Pushed fixes we discussed about + changing non-free icons

Dapoulp:
Pushed buttons recolor (and removed elements.css file)

PR #420, opened 2025-07-26
[Fix] GetDocFromElement Async

WBHarry:
GetDocFromElement was erroring out when used in compendiums as it can't be used async for embedded collections on compendiums. Fixed by making it async.

- GetDocFromElement is now async to work with compendiums.
- Added GetDocFromElementSync for use in contextMenu condition cases.
- Updated current uses to be async and await when needed.


cptn-cosmo (commented):
IDKWIAD

cptn-cosmo (approved):
IDKWIAD

PR #423, opened 2025-07-26
[Fix] Contexmenu Compendium Edit

WBHarry:
The edit option in the contextmenu is now always shown on compendiums


cptn-cosmo (commented):
ok i guess

cptn-cosmo (approved):

PR #439, opened 2025-07-28
[Feature] ActiveEffect Improvements

WBHarry:
- Fixed so genericStatuses aren't shown in activeEffectConfig if the system setting is turned off. 
- Fixed localization of genericStatuses in Inventory-ItemV2, they weren't before.
- Effects with statuses on them will show them on the token - they will also be shown as active in the Token HUD.


cptn-cosmo (approved):
lgtm

PR #448, opened 2025-07-28
[BUG] Levelup Proficiency Fix

WBHarry:
Fixed so that extra selected proficiciency is added correctly. It was being set, rather than added 😆


moliloo (approved):
Always that small bug XD

PR #450, opened 2025-07-29
Feat: add massive and powerful feature on weapon

dbyzero:
## Description

I rework as clean as possible the effect "massive" and "powerfull".

I think adding a dice then remove it is not a good approach for two reason :
- Foundry give us a native way to manage this
- Critical roll will check the final number of dice to apply max dice value to damage

It's for this reason that I place it directly in damageRoll.

I remove also the action on "Poweful" weapon because it's not an other action, it's now directly embedded in the default action.

## Type of Change

- [x] New feature
- [x] Code cleanup/refactor

## How Has This Been Tested?

- [x] Manual testing
- [ ] Other:

Peek 2025-07-29 03-32

## Checklist

### DIFF
--- a/module/config/itemConfig.mjs
+++ b/module/config/itemConfig.mjs
@@ -946,14 +946,9 @@ export const weaponFeatures = {
                         value: '-1'
                     },
                     {
-                        key: 'system.bonuses.damage.primaryWeapon.extraDice',
+                        key: 'system.bonuses.damage.primaryWeapon.addDiceAndDiscardLowest',
                         mode: 2,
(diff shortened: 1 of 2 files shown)


WBHarry:
We're aiming to do this kind of change, but this isn't the place in the code where we want to make it.
A lot of things are moving around, and we will be implementing it in a different place.

Thank you all the same. 🫡

dbyzero:
No problem I understand, I needed this code for my game this evening. Was very instructive to find a way to implement it on my side.

PR #492, opened 2025-07-31
[Feature] 479 - Check Loot/Consumables

### LINKED ISSUE #479: Check / re-add healing items in consumables and loot
They might be missing due to having been previously removed. Need to check if they have been re-added and add if not

WBHarry:
Closes #479 
Some corrections and adds to Consumables and Loot compendia


Psitacus (approved):
Image

PR #501, opened 2025-08-01
[Fix] Compendium Bug On CharacterCreation

WBHarry:
Made sure to load compendium document if it wasn't already


cptn-cosmo (approved):
Lgtm

PR #551, opened 2025-08-03
[Fix] CostSelection In Rollmessage

WBHarry:
Fixed so CostSelection is always in the RollSelection. Previously it was excluded in "lite" messages.


Psitacus (approved):
Image

PR #579, opened 2025-08-04
Domain cards cleanup pass with Icons and Descriptions

### LINKED ISSUE #527: Fix previously entered domains that need some improvements
Sage, Grace, and Valor were entered prior and need the following improvements after the chat rework and the new icons:

- Add correct icons everywhere
- Add descriptions to each of the actions/features within a card, as that is what gets sent to chat

Psitacus:
## Description

- Closes #527


WBHarry (approved):
:robot: :+1:

PR #582, opened 2025-08-05
[Fix] Subclass ActiveEffects

WBHarry:
Fixed so subclass features beyond foundation are only transferred when they should - IE, you have upgraded your subclass.


Ikraik (approved):
👍

PR #600, opened 2025-08-05
[Fix] 597/598 Armor Clearing

### LINKED ISSUE #597: [BUG] - Guardian; Frontline Tank ability has some problems when hitting the Apply Healing button
**Describe the bug**
Class - Guardian; Frontline Tank ability has some problems when hitting the Apply Healing button: 
1. It fails if you have an adversary targeted; 
2. It fails if you have nothing targeted; 
3. It fails if you have yourself targeted. 
4. Does remove the Hope though.

**To Reproduce**
1. Create a Guardian
2. Try the steps 1 to 3 above
3. Note the errors/message in the console
4. Note that  the Hope is removed and the Armor Slots are not Healed.

**Expected behavior**
This ability should work no matter what is targeted. It's a self only ability. It should also do the Armor Slot healing too. 

**Setup Information:**
 - OS: Linux
 - Browser: Chrome
 - Foundry Version: v13 b346
- System Version: commit - 66d3b114083b1a97b550f5cf111d2a92c448f89b

### LINKED ISSUE #598: swap evasion and proficiency
on the character sheet evasion is more important than proficiency, therefore their positions should be swapped

WBHarry:
Closes #597 
Closes #598 
- Fixed armor healing, it was a general problem because we had an old internal legacy name in a bunch of places.
- Swapped places of Proficiency and Evasion in character sheet.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -897,7 +897,7 @@
                     "name": "Hope",
                     "abbreviation": "HO"
                 },
-                "armorSlot": {
+                "armor": {
                     "name": "Armor Slot",
                     "abbreviation": "AS"
                 },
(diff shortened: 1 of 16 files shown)


cptn-cosmo (approved):
lgtm

PR #625, opened 2025-08-06
Fix the effect of rage up - Compendium only

Psitacus:
## Description

This is only the fix to the effect and cost that was incorrectly setup on rage-up. There is still the issue of the two effects weirdly being mutually exclusive.


cptn-cosmo (approved):
lgtm

PR #639, opened 2025-08-06
[Feature] Homebrew Domains

WBHarry:
- Added homebrew domains to Automation Settings
- Automation Settings now have tabs
- Fixed Tagify to properly distinguish between names and values. Previously it erroniously required name and value to be the same to work properly.


cptn-cosmo (approved):
lgtm

PR #708, opened 2025-08-07
[fix] SimpleReaction Roll Improvement

WBHarry:
Changing the title so it's more clear you've switched to a reaction roll


Ikraik (approved):
👍

PR #775, opened 2025-08-10
Add build to ignore for my linux dev

chrisryan10:
Need to lock out the 'build' directory so it doesn't get added to the repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,6 @@
 node_modules
 /packs
 Build
+build
 foundry
 styles/daggerheart.css


WBHarry:
Is it case sensitive on linux?
Just `Build` was blocking it for `build` on windows.

chrisryan10:
Yes, Linux is case sensitive.

WBHarry:
I think you should make a pull request of this branch against `hotfix` aswell in that case, so it gets added there too.

PR #832, opened 2025-08-11
Partial fix for #671

### LINKED ISSUE #671: [BUG] - Character Reaction Roll chat display is confusing
**Describe the bug**
When doing a Reaction Roll for a character, the Chat Card is titled Duality Roll and shows the roll as X with Fear/Hope.

Finding how to do a Reaction Roll on a character was not immediately clear, but once discovered, it becomes straightforward/understandable.

**To Reproduce**
1. Use a PC
2. Click on an Attribute to bring up the Duality Roll dialog
3. Select Reaction Roll
4. Roll the dice
5. See error

**Expected behavior**
1. Ideally when the little Reaction Roll icon is hit, the Duality Roll text changes to Reaction Roll
2. The Chat Card of the result has the title "Attribute" Reaction Roll where "Attribute" is changed to the name of the Attribute used.
3. The Chat Card shows the result without the: With Fear/Hope
4. Ideally, the card is a different color to the Hope/Fear cards to show Reaction Roll

**Setup Information:**
- System Version main commit a1a3ccc461f79e767b9eed28db46bffd0e7db275

chrisryan10:
## Description

Partial fix for #671. This reduces the potential confusion for Reaction Roll results and matches the rules more closely. See the screenshots below for more details.

- Fixes #671 

## Type of Change

- [x ] Bug fix

## How Has This Been Tested?

- [ x] Manual testing

Screenshot_20250811_183045

**Note the "Agility Reaction Check" title, and the result has no "with Hope", as per the rules.**

Screenshot_20250811_183251

**Note the "Agility Reaction Check" title, and the result has no "with Fear", as per the rules.**

The above 2 images were successful rolls against the Difficulty.

Screenshot_20250811_183429

**A failed Agility reaction roll.**

Screenshot_20250811_183858

**An Agility reaction roll that is a critical. Note that the "Critical" wording in the result remains, as per the rules, crits give extra benefits than a normal success (but no Hope, no clear a Stress).**

PROBLEM:
Screenshot_20250811_184444
Screenshot_20250811_184513

If I do a reaction roll by clicking an attribute on a character sheet, you get the DR dialog and have to hit the reaction icon to change its mode.  However the details in the chat don't get changed properly, because (I think) the title details are built before it knows it's a reaction. I have not worked out how to fix this.

## Checklist

- [ x] My code follows the project style guidelines
- [ x] I have performed a self-review of my code

### DIFF
--- a/daggerheart.mjs
+++ b/daggerheart.mjs
@@ -209,8 +209,10 @@ Hooks.on('chatMessage', (_, message) => {
         const difficulty = rollCommand.difficulty;
 
         const target = getCommandTarget({ allowNull: true });
+        const titleType = reaction ? 'DAGGERHEART.UI.Chat.dualityRoll.abilityReactionCheckTitle'
+                                   : 'DAGGERHEART.UI.Chat.dualityRoll.abilityCheckTitle';
         const title = traitValue
-            ? game.i18n.format('DAGGERHEART.UI.Chat.dualityRoll.abilityCheckTitle', {
+            ? game.i18n.format(titleType, {
                   ability: game.i18n.localize(SYSTEM.ACTOR.abilities[traitValue].label)
               })
             : game.i18n.localize('DAGGERHEART.GENERAL.duality');
--- a/lang/en.json
+++ b/lang/en.json
@@ -2265,7 +2265,8 @@
                     "title": "Domain Card"
                 },
                 "dualityRoll": {
-                    "abilityCheckTitle": "{ability} Check"
+                    "abilityCheckTitle": "{ability} Check",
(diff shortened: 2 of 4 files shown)


chrisryan10:
Going to get this through. It's not the full fix, but it's better than the current output. This at least reduces potential confusion.

Dapoulp:
As I said, I disagree on the fact to remove the "with Fear/Hope" info.
Even if there is no hope/stress/fear gain/lose, the roll is still made with fear or hope, which can be useful for GM

cptn-cosmo:
though i would argue that is still visible due to the color of the roll, which should be enough for reaction roils at least?

PR #883, opened 2025-08-12
[Fix] 846 Effect Item ModelData

WBHarry:
Fixed so effects supposed to use item data use the model directly, since items have no rolldata


chrisryan10 (approved):

chrisryan10:
Tested for both Tower and Round Shield. And going multiple tiers. Worked as expected.

PR #890, opened 2025-08-13
Change the critical damage max calculation

### LINKED ISSUE #889: [BUG] - critical maxing
Criticals are maxing the dice rather than rolling new ones and also including maximum as the modifier as well

chrisryan10:
## Description

Alternate attempt at calculating max damage portion of critical roll.

- Fixes #889
 
## Type of Change

- [ x] Bug fix

## How Has This Been Tested?

- [ x] Manual testing

When testing this, cover the following:
* different tier characters
* abilities that add to the damage, e.g. sneak attack
* check the modifiers are correct, e.g. weapon dam is 1d8 + 1, crit should appear as 1d8 + 1 + 8 (plus other bonuses).

### DIFF
--- a/module/dice/damageRoll.mjs
+++ b/module/dice/damageRoll.mjs
@@ -137,12 +137,7 @@ export default class DamageRoll extends DHRoll {
         }
 
         if (config.isCritical && part.applyTo === CONFIG.DH.GENERAL.healingTypes.hitPoints.id) {
-            let total = 0;
-            part.roll.terms.forEach(term => {
-                if (term._faces) {
-                  total += term._faces * term._number;
-                }
-            });
+            const total = part.roll.dice.reduce((acc, term) => acc + term._faces*term._number, 0);
             if (total > 0) {
                 part.roll.terms.push(...this.formatModifier(total));
             }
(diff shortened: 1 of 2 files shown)


module/dice/damageRoll.mjs
```diff
@@ -137,9 +137,15 @@ export default class DamageRoll extends DHRoll {
         }
 
         if (config.isCritical && part.applyTo === CONFIG.DH.GENERAL.healingTypes.hitPoints.id) {
-            const tmpRoll = Roll.fromTerms(part.roll.terms)._evaluateSync({ maximize: true }),
-                criticalBonus = tmpRoll.total - this.constructor.calculateTotalModifiers(tmpRoll);
-            part.roll.terms.push(...this.formatModifier(criticalBonus));
+            let total = 0;
+            part.roll.terms.forEach(term => {
```
WBHarry: This works fine in function. I think we could do it a little more neatly though:
```
let total = part.roll.dice.reduce((acc, term) => acc + term._faces*term._number, 0);
```

Checking dice instead of terms lets us skip verifying that it's a dice term, and reduce is tailor made to increase a number while looping through.

WBHarry (changes requested):

WBHarry (approved):
🤖 :+1:

PR #914, opened 2025-08-13
[Feature] 648 - Mark Defeated Actors

### LINKED ISSUE #648: [BUG] - Killing an Adversary issues
**Describe the bug**
When an Adversary is killed (i.e. all HP used) no Dead status is applied to the token.
A weird tooltip appears above the image in the Adversary's card - DAGGERHEART.UI.Tooltip.DeathMove - this is not clickable.

**To Reproduce**
1. Put an Adversary in a scene
2. Attack and kill it via damage, e.g. weapon hits
3. When the HP are all used, see the above issues.

**Expected behavior**
The Dead status should appear.
Not sure if Death Moves are applicable for Adversaries - I can't remember reading any rule about that. Either way, the tooltip is a bit confusing (and difficult to see)

**Setup Information:**
- System Version main commit 02958f9574981609ba9a4480ff538c4f5d5b3bcc

WBHarry:
Closes #648 
- Improved death marking styling on sheets.
- Added automation for defeated status
- Fixed so the tracker recognises and sets the correct defeated statuses based on actor type


Ikraik (approved):
👍

PR #937, opened 2025-08-14
[Fix] 924 - Damage Reduction Any

### LINKED ISSUE #924: [BUG] - Stress Damage Reduction: Any
**Describe the bug**
I was able to do minor,major,severe all the same way by doing 2 of the of each attribute but second input of each pair I deleted ".cost" and replaced it with ".enabled". This method does not work for "any"
Dwarf Ancestry also has the skill "Thick Skin" it was working until I added the 2nd Attribute Key ".enabled" for it to work.

**To Reproduce**
Use a skill that uses stress to reduce damage (example: Thick Skin) under dwarf
1. Go to 'effects'
2. Click on 'edit'
3. Click on Changes Tab
4. Click on the + icon by priority
5. drop down on attribute key to "stress damage reduction: any and select it
6. change mode to override
7. value of 2
8. repeats steps 4 to 6 but this time change value to 1
9. click on the 2nd attribute key made and all the way to the right delete ".cost" and replace with "enabled"
10. It will not work but if you do this for the other 3 stress damage reduction options they will work.

**Expected behavior**
I'm expecting "Stress Damage Reduction: Any" to follow the same logic as the other 3.
**Screenshots**

Image

**Setup**
- Os Windows 11
 - Browser no browser its the windows app
 - Foundry Version v13 b347
- System Version Foundryborne 1.0.2

**Additional context**

WBHarry:
Fixes #924 
- Fixed damage reduction any not working
- Removed unneccessary `damageReduction.<>.enabled` flag. It's enough to just check if there is a cost available.

### DIFF
--- a/module/applications/dialogs/damageReductionDialog.mjs
+++ b/module/applications/dialogs/damageReductionDialog.mjs
@@ -39,7 +39,7 @@ export default class DamageReductionDialog extends HandlebarsApplicationMixin(Ap
         this.availableStressReductions = Object.keys(actor.system.rules.damageReduction.stressDamageReduction).reduce(
             (acc, key) => {
                 const dr = actor.system.rules.damageReduction.stressDamageReduction[key];
-                if (dr.enabled) {
+                if (dr.cost) {
                     if (acc === null) acc = {};
 
                     const damage = damageKeyToNumber(key);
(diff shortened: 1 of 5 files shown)


Ikraik (approved):
Approved :D Tested and it works

PR #973, opened 2025-08-16
Fix/add translation key for Unarmed Attack

### LINKED ISSUE #953: Can't Translate Unarmed Attack
Can't Translate Unarmed Attack

Image

chrisryan10:
## Description

Fix/add translation key for Unarmed Attack. 

- Fixes #953 

## Type of Change

- [ x] Bug fix

## How Has This Been Tested?

- [x ] Manual testing

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1985,7 +1985,7 @@
             "true": "True",
             "type": "Type",
             "unarmed": "Unarmed",
-            "unarmedStrike": "Unarmed Strike",
+            "unarmedAttack": "Unarmed Attack",
             "unarmored": "Unarmored",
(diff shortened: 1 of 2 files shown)


chrisryan10:
Not sure if this will require some form of migration, because existing characters will still have "Unarmed Attack" even after this change, no matter the language used (I think).  And it does not seem possible to remove and replace the Unarmed Attack item on a character, i.e. the delete does not work.

WBHarry (approved):
Aslong as the `attack` hasn't been edited, it will get the update. 
And if it has been edited, then we shouldn't mess with it either way, so I think this is all good 👍

Issue #870, opened 2025-08-12
[Feature] Increase opacity of dark sheet background


CarlosFdez:
It is kinda hard to read character and adversary sheets on bright scenes. A friendo was trying to set it up and had this issue, so I started poking into it

<img width="642" height="753" alt="Image" src="https://github.com/user-attachments/assets/fe1434d9-dfa7-45b7-a3e7-6452ed150ebe" />

I tried to poke into it with the thought of making a PR, but it looks like you have a less variable per opacity. I'm not sure if adding a `@dark-blue-a0` is the right approach, or if I should consider using a less function (fade?). This is after I change the opacity from 60 to a0, which I think is clearer, but honestly even 90 would be fine.

<img width="778" height="739" alt="Image" src="https://github.com/user-attachments/assets/09ea9a46-dadc-4328-aeee-97dc879423b7" />

EDIT: it looks like for foundry, the transparency is %, not hex value. -For example, -color-cool-5-90 is 0.9 transparency, or e6. In the daggerheart system, its the hex value instead.

cptn-cosmo:
i like this, at least as an option

CarlosFdez:
Let me know the preferred way to do it and I can push something up. Its very simple though, so no worries.

I looked at the foundry css variable names for those with alpha and it looks like those are %s (90 is 0.9, not 0x90 aka 144), so the convention doesn't seem shared, but that's not necessarily a bad thing.

PR #974, opened 2025-08-16
Use more opaque color for sheet background

CarlosFdez:
## Description

Uses a more opaque color for sheet backgrounds to try to improve visibility on bright scenes. Closes https://github.com/Foundryborne/daggerheart/issues/870

I am uncertain about the less variable names, given that foundry uses a percentage for the opacity (90 = 90% aka e9). But I kept the convention rather than rocking the boat. Foundry uses e9 for their backgrounds, so the new value of a0 is still lower than what foundry uses. The daggerheart system has a higher blur value which helps, although all of that is disabled on lower performance modes.

I think the PR should probably be using an even more opaque value and reducing the blur a smidge. Let me know if I should do that, or if there are any adjustments you'd like me to do at the same time.

## Screenshots

I included the normal foundry application for comparison

image

In low performance mode, it looks like the following. Its still a bit hard to see imo, but better:

image


CarlosFdez:
Even though my preference is to go even more opaque, I dialed it back a bit to satisfy more preferences (A0 -> 90). I added an override specifically for low performance mode though, since the higher blur daggerheart uses for contrast doesn't help those guys.

styles/less/global/sheet.less
```diff
@@ -4,12 +4,23 @@
...
-    backdrop-filter: blur(10px);
+    background: @dark-blue-90;
+    backdrop-filter: blur(8px);
 }, {
     background: url('../assets/parchments/dh-parchment-light.png') no-repeat center;
 });
 
+// Override for dark mode with blur disabled to maintain visibility
+// This sheet uses significantly lower opacity than foundry and compensates with a higher blur
+// But if there is no blur, the sheet is very difficult to see otherwise
```
moliloo: Could you remove this comments? I think its not necessary a explanation about the code, you can change it to simply `// low perfomance background` or no add comments at all

CarlosFdez:
Done!

PR #1026, opened 2025-08-19
[Fix] DiceSoNice DamageRolls

WBHarry:
Fixed so the 3d damage dice can be seen by everyone when not whispered, previously only the one doing the roll got them.

### DIFF
--- a/module/dice/damageRoll.mjs
+++ b/module/dice/damageRoll.mjs
@@ -37,7 +37,13 @@ export default class DamageRoll extends DHRoll {
                     Object.values(config.damage).flatMap(r => r.parts.map(p => p.roll))
                 ),
                 diceRoll = Roll.fromTerms([pool]);
-            await game.dice3d.showForRoll(diceRoll, game.user, true, chatMessage.whisper, chatMessage.blind);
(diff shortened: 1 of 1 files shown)


module/dice/damageRoll.mjs
```diff
@@ -37,7 +37,13 @@ export default class DamageRoll extends DHRoll {
                     Object.values(config.damage).flatMap(r => r.parts.map(p => p.roll))
                 ),
                 diceRoll = Roll.fromTerms([pool]);
-            await game.dice3d.showForRoll(diceRoll, game.user, true, chatMessage.whisper, chatMessage.blind);
+            await game.dice3d.showForRoll(
+                diceRoll,
+                game.user,
+                true,
+                chatMessage.whisper?.length > 0 ? chatMessage.whipser : null,
```
chrisryan10: whipser --> whisper

WBHarry: Fixed 👍

WBHarry:
I'm thinking main. I think there's a fair chance we'll drop another patch before 1.1, and this is definately something I'd want fixed then.

PR #1039, opened 2025-08-21
[Feature] 1031 - Action Names In Chat

### LINKED ISSUE #1031: [Feature] Stop clubbing actor name with the attacks
**Description**
Do not club the name of the sidebar actor inside of the attack
We put the name of the token into the chat card (minor demon), as well as the name of the actor (this name is broadcast) inside of the attack. 

1 - the name of the sidebar actor should not show, as the GM might want to hide it
2 - by having it clubbed with the attack, modules that want to modify the token name or hide it, cannot do it since they miss the sidebar name.

**Screenshots**

Image

WBHarry:
Fixes #1031 
Fixed so that actions don't print the actor name in chat along with its own name

### DIFF
--- a/module/data/action/baseAction.mjs
+++ b/module/data/action/baseAction.mjs
@@ -163,7 +163,7 @@ export default class DHBaseAction extends ActionMixin(foundry.abstract.DataModel
         const hasRoll = this.getUseHasRoll(byPass);
         return {
             event,
-            title: `${this.item.name}: ${game.i18n.localize(this.name)}`,
+            title: game.i18n.localize(this.name),
             source: {
                 item: this.item._id,
                 action: this._id,


module/data/action/baseAction.mjs
```diff
@@ -163,7 +163,7 @@ export default class DHBaseAction extends ActionMixin(foundry.abstract.DataModel
         const hasRoll = this.getUseHasRoll(byPass);
         return {
             event,
-            title: `${this.item.name}: ${game.i18n.localize(this.name)}`,
+            title: game.i18n.localize(this.name),
```
Dapoulp: We should keep the Item name if it's an item but hide it if it's an Actor, so something like:
```js
title: `${this.item instanceof CONFIG.Actor.documentClass ? "" : `${this.item.name}: `}${game.i18n.localize(this.name)}`,
```

Dapoulp (changes requested):

module/data/action/baseAction.mjs
WBHarry: Added that change. Looks good :+1:

PR #1049, opened 2025-08-22
[Feature] 687 - Companion Improvements

### LINKED ISSUE #687: [BUG] - Companion Improvements
- Companions should not be shown in the Combat Tracker
- Companions should have a button for doing a general roll. Add a setting on the companion to select which trait on the Partner is used for it, defaulting to Instinct. If no partner is selected, then the button should be disabled.

WBHarry:
Closes #687 
Companions can't be put into CombatState anymore. Companions now have a Action Roll button

### DIFF
--- a/module/applications/hud/tokenHUD.mjs
+++ b/module/applications/hud/tokenHUD.mjs
@@ -13,6 +13,9 @@ export default class DHTokenHUD extends foundry.applications.hud.TokenHUD {
 
     async _prepareContext(options) {
         const context = await super._prepareContext(options);
+        const nonCombatTypes = ['environment', 'companion'];
+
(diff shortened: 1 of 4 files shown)


chrisryan10:
If a companion is in a group of selected tokens, and they're set to Combat State, the companion will appear in the tracker.

chrisryan10 (changes requested):
If a companion is in a group of selected tokens, and they're set to Combat State, the companion will appear in the tracker.

WBHarry:
> If a companion is in a group of selected tokens, and they're set to Combat State, the companion will appear in the tracker.

Added a fix for this :+1:

chrisryan10 (approved):

PR #1089, opened 2025-08-25
[Feature] 946 - Damage/Effect ChatMessages

### LINKED ISSUE #946: [Feature] Chat Damage/Effects Summary
When you press the "deal damage" button, the target's hit points are marked, but no chat message mentions this. 

This is annoying for two reasons:
- If the attack hits multiple enemies, you might forget who you've already dealt damage to.
- You can't see if resistances or other effects came into play during the process.

A message like "zombie marks 2 hit points" should be added.

WBHarry:
Closes #946 

**Remaining:**
- Not sure how to get an alternate title on the chat message atm. It's not applying.
- Do we want a setting that determines if these messages are created?
image

### DIFF
--- a/module/documents/chatMessage.mjs
+++ b/module/documents/chatMessage.mjs
@@ -208,32 +208,6 @@ export default class DhpChatMessage extends foundry.documents.ChatMessage {
                 )
             );
         }
-
-        Promise.all(damagePromises).then(async _ => {
-            const summaryMessageSettings = game.settings.get(
-                CONFIG.DH.id,
-                CONFIG.DH.SETTINGS.gameSettings.Automation
-            ).summaryMessages;
-            if (!summaryMessageSettings.damage) return;
-
-            const cls = getDocumentClass('ChatMessage');
-            const msg = {
-                type: 'systemMessage',
-                user: game.user.id,
-                speaker: cls.getSpeaker(),
-                title: game.i18n.localize(
-                    `DAGGERHEART.UI.Chat.damageSummary.${this.system.hasHealing ? 'healingTitle' : 'title'}`
-                ),
-                content: await foundry.applications.handlebars.renderTemplate(
-                    'systems/daggerheart/templates/ui/chat/damageSummary.hbs',
-                    {
-                        targets: targetDamage
-                    }
-                )
-            };
-
-            cls.create(msg);
-        });
     }
 
     async onRollAllSave(event) {
--- a/module/data/fields/action/damageField.mjs
+++ b/module/data/fields/action/damageField.mjs
@@ -100,13 +100,15 @@ export default class DamageField extends fields.SchemaField {
 
             if (config.hasHealing)
                 damagePromises.push(
-                    actor.takeHealing(config.damage).then(updates => targetDamage.push({ token: actor.token, updates }))
+                    actor
+                        .takeHealing(config.damage)
+                        .then(updates => targetDamage.push({ token: actor.token ?? actor.prototypeToken, updates }))
                 );
             else
                 damagePromises.push(
                     actor
                         .takeDamage(config.damage, config.isDirect)
-                        .then(updates => targetDamage.push({ token: actor.token, updates }))
+                        .then(updates => targetDamage.push({ token: actor.token ?? actor.prototypeToken, updates }))
                 );
         }
 


moliloo (commented):
I can't approve now because i working in some implementations in this PR

module/documents/actor.mjs
```diff
@@ -597,7 +597,9 @@ export default class DhpActor extends Actor {
 
         await this.modifyResource(updates);
 
-        if (Hooks.call(`${CONFIG.DH.id}.postTakeDamage`, this, updates) === false) return null;
+        Hooks.call(`${CONFIG.DH.id}.postTakeDamage`, this, updates);
```
Dapoulp: Don't remove Hook call return check, it's mandatory to handle the workflow properly

module/documents/actor.mjs
```diff
@@ -645,7 +647,8 @@ export default class DhpActor extends Actor {
 
         await this.modifyResource(updates);
 
-        if (Hooks.call(`${CONFIG.DH.id}.postTakeHealing`, this, updates) === false) return null;
+        Hooks.call(`${CONFIG.DH.id}.postTakeHealing`, this, updates);
```
Dapoulp: Don't remove Hook call return check, it's mandatory to handle the workflow properly

module/documents/chatMessage.mjs
```diff
@@ -131,6 +131,41 @@ export default class DhpChatMessage extends foundry.documents.ChatMessage {
...
+
+    async _onTokenTargetClick(event) {
+        const { tokenUuid } = event.target.dataset;
+        if (!canvas.ready || !tokenUuid) return;
+
+        const token = await foundry.utils.fromUuid(tokenUuid);
+        if (token?.object) {
+            const { _x: x, _y: y } = token.object.position;
+            game.canvas.pan({ x, y });
+        }
```
Dapoulp: Why recreate hovering/click target ? It already exists in that same file.
Better to use same methods, updating existing ones if necessary

module/documents/chatMessage.mjs
```diff
@@ -174,9 +210,28 @@ export default class DhpChatMessage extends foundry.documents.ChatMessage {
             }
 
             this.consumeOnSuccess();
-            if (this.system.hasHealing) target.actor.takeHealing(damages);
-            else target.actor.takeDamage(damages, this.system.isDirect);
+            const updates = this.system.hasHealing
+                ? await target.actor.takeHealing(damages)
+                : await target.actor.takeDamage(damages, this.system.isDirect);
+            targetDamage.push({ token: target, updates: updates });
```
Dapoulp: `await` on takeHealing/takeDamage MUST be removed or it will lock the whole process, and we really don't want that.
Instead, put those async method calls into an array and use Promise.all to then display the message

module/documents/chatMessage.mjs
```diff
@@ -202,6 +257,25 @@ export default class DhpChatMessage extends foundry.documents.ChatMessage {
                 ui.notifications.info(game.i18n.localize('DAGGERHEART.UI.Notifications.noTargetsSelected'));
             this.consumeOnSuccess();
             await action.applyEffects(event, this, targets);
+
+            const effects = this.system.action.item.effects.filter(effect =>
+                this.system.action.effects.some(x => x._id === effect.id)
+            );
```
Dapoulp: This won't display correct applied effects. A target can potentially succeed a Reaction Roll, some effects are not applied on a success, others are still applied. So list of effects may vary per targets

Dapoulp (changes requested):

Ikraik:
@WBHarry This PR is old and has conflicts. Do resolve them once the changes are made as Poulp mentioned.

module/documents/actor.mjs
WBHarry: Fixed :+1:

module/documents/actor.mjs
WBHarry: Fixed :+1:

module/documents/chatMessage.mjs
WBHarry: Just didn't check to see there was something like that 👅 

Removed these methods. Using the other ones now 👍

module/documents/chatMessage.mjs
WBHarry: Changed to use that solution :+1:

WBHarry: The actual message creation is now in `damageField` and `effectsField`, since the execute functionality has been moved to there.

module/documents/chatMessage.mjs
WBHarry: I changed so that targets that don't have any effects applied to them are excluded from the message.
I don't think we have individual saves for an action with multiple effects to apply, right? So I think that should solve it. (If multiple saves are desired, it will be multiple separate actions)

Ikraik:
Is this PR good enough to push in?

WBHarry:
If it looks good to you in testing, then I think so. Molilo changed up the styling and I fixed some issues.

Ikraik:
image
@WBHarry Is this resolved in code as well?

Ikraik (changes requested):
image

Target summary is accurate but target tokens are not matching creating console warning. 

Text on console: The target token no longer exists

Ikraik:
instead of cleared, having the word as marked would be better imo

Ikraik:
@WBHarry Once you are free from existing task, it looks like some issue here. I marked my review point above, lemme know if there's a way to resolve it. I can't seem to get my head around it.

WBHarry:
> image
> Target summary is accurate but target tokens are not matching creating console warning.
> 
> Text on console: The target token no longer exists

I'm assuming you were targeting tokens of actors you have deleted. I can't find an issue. 🤷‍♂️

WBHarry:
@Ikraik 
I've gone over the points raised and fixed.

- The alternative title is now getting in there (Damage Applied / Effects Applied)
- There is a setting to enable/disable showing the summaries for damage and effects respectively.
- The word `cleared` was being shown all the time because of a bug. When it's damage, it will now show `Marked`. `Cleared` is shown on healing.

PR #1090, opened 2025-08-25
[Feature] 942 - Limited Sheet Views

### LINKED ISSUE #942: [Feature] Change the way limited interacts with Actor sheets
A common place feature of Foundry modules is user configuration and how that effects what players can and cannot see. In the context of actors this usually boils down to;

None- Usually means no acess what so ever

Limited- Means players can see some information, usually a bio or maybe just the name and art of the Actor.

Observer- Can see the entire sheet, but cannot interact

Owner- Can see the sheet, edit it and use it.

Currently, Daggerheart treats limited the same as Observer. This means if I wanted to give my players acess to information on NPCs, I have to make individual Journal entires instead of just giving them acess to the actor and a public bio, which over complicates things.

Would it be possible to allow players to see a bio while given limited acess? This may require an extra text field so that "Notes" can still remain private for the GM.

WBHarry:
Closes #942 
Added stumps for all limited actor views. Still need to consolidate what content we want in the views, and the design.

### DIFF
--- a/styles/less/sheets/actors/actor-sheet-shared.less
+++ b/styles/less/sheets/actors/actor-sheet-shared.less
@@ -17,4 +17,40 @@
         font-family: @font-body;
         color: light-dark(@chat-blue-bg, @beige-50);
     }
+
+    &.limited {
+        &.character,
+        &.adversary,
+        &.environment,
+        &.companion {
+            width: 360px !important;
+
+            .window-content {
+                display: unset;
+            }
+
+            .limited-container {
+                width: 100%;
+                padding-top: var(--header-height);
+                display: flex;
+                flex-direction: column;
+                gap: 8px;
+
+                header {
+                    width: 100%;
+                    display: flex;
+                    flex-direction: column;
+                    align-items: center;
+
+                    .profile {
+                        width: 275px;
+                    }
+
+                    .title-name {
+                        text-align: center;
+                    }
+                }
+            }
+        }
+    }
 }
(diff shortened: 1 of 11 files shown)


moliloo:
I will look at the style of limited sheet in this weekedn

WBHarry:
Changed so that the sheets are no resizable when they are shown as `limited`

Ikraik:
https://github.com/user-attachments/assets/50266203-b374-437f-ad1f-c77925941c48

hmm, resize is still there in my branch. I'll test it on a new world just in case and update the comment here.

Update: No resize in new world. Invisible bounds still there

Ikraik (changes requested):
https://github.com/user-attachments/assets/7dece5d8-d84d-4619-baf4-2bb02099fd97

Issue still persists where there seems to be invisible bounds blocking the sheet to be dragged around. This is tested on a new world. My guess is that the size is based on the full sheet and not on the size of the limited sheet hence setting that bounds to the container.

WBHarry:
Hm..I think I should be able to change that. Will give it a shot

styles/less/sheets/actors/actor-sheet-shared.less
```diff
@@ -17,4 +36,191 @@
...
+                    width: 100%;
+                    display: flex;
+                    flex-direction: column;
+                    align-items: center;
+                    top: -36px;
+                    margin-bottom: -30px;
+
+                    .profile {
+                        width: 100%;
+                        max-height: 275px;
```
Ikraik: add 'max-width: 275px' as well

Ikraik, line 69:
```suggestion
                        max-height: 275px;
                        max-width: fit-content;
```

This seems to resolve the resizing issue on the profile image.

Ikraik (changes requested):
Added suggestion for improving profile image scaling

Ikraik (approved):
LGTM

PR #1106, opened 2025-08-28
[Feature] 1033 - Consume Quantity On Use

### LINKED ISSUE #1033: [Feature] Consuming consumable on use
**Description**
The feature for consuming the consumable items on use is not implemented.

**Screenshots**

Image

WBHarry:
Closes #1033 
- Added Quantity as an available cost in ActionConfig for Consumables and Loot items.
- Updated compendium Consumables to consume quantity.
- Reworked CostField so it's easier and more logical to consume properties from the parent item.

### DIFF
--- a/module/data/fields/action/costField.mjs
+++ b/module/data/fields/action/costField.mjs
@@ -46,16 +46,11 @@ export default class CostField extends fields.ArrayField {
 
         if (this.parent?.parent) {
             for (var cost of config.costs) {
-                if (cost.keyIsID) {
+                if (cost.itemId) {
                     usefulResources[cost.key] = {
                         value: cost.value,
                         target: this.parent.parent,
-                        keyIsID: true
-                    };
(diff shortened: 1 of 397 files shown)


module/data/fields/action/costField.mjs
```diff
@@ -176,10 +178,44 @@ export default class CostField extends fields.ArrayField {
...
+                };
+            default:
+                return { path: '', value: undefined };
+        }
+    }
+
     /**
-     * 
-     * @param {*} costs 
-     * @returns
```
Ikraik, line 182: Sigh I don't see why comments get part of diff XD

Ikraik (dismissed):
We didn't add the functionality towards the consumables where if pressed on an action it would have an item cost and reduces that. It seems to be missing. Maybe we can add that now instead of new PR to focus on this. Probably using this PR branch would help.

PR #1103, opened 2025-08-27
Compendium browser per type

Dapoulp:
## Description

Compendium Browser will now automatically display custom compendiums too.

- Fixes:
  - https://github.com/Foundryborne/daggerheart/issues/1100
  - https://github.com/Foundryborne/daggerheart/issues/1101
  - https://github.com/Foundryborne/daggerheart/issues/743
- Closes:
   - https://github.com/Foundryborne/daggerheart/issues/909

## Type of Change

Please check the relevant options:

- [x] Bug fix
- [x] New feature
- [x] Code cleanup/refactor
- [ ] Documentation update
- [ ] Test coverage
- [ ] Dependency update
- [ ] Configuration change
- [ ] Other (please describe):

## How Has This Been Tested?

Please describe the tests you ran to verify your changes:

- [ ] Manual testing
- [ ] Other:

## Screenshots (if applicable)

Include screenshots or GIFs to help explain your changes visually.

## Checklist

- [ ] My code follows the project style guidelines
- [ ] I have performed a self-review of my code
- [ ] I have commented my code where necessary
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings or errors
- [ ] I have added tests that prove my fix or feature works
- [ ] New and existing tests pass locally with my changes

## Additional Comments

Add any other context or questions here.

---

> Thank you for your contribution! 🎉

PR #1108, opened 2025-08-28
Use less ambiguous sort direction symbols

chrisryan10:
## Description

From #1103, I mentioned in the PR that the sort arrows were in the wrong direction. During a discord discussion I discovered that the sort arrows mean different things to different people. This change is an attempt to use icons that are less ambiguous for everyone.

There is no issue associated with this change (except the PR mentioned above). 

## How Has This Been Tested?

- [x ] Manual testing

Open the Compendium Browser and open Domain Cards (for example).  Sort the columns via the header bar. Note the different sort direction icons.

### DIFF
--- a/styles/less/ui/item-browser/item-browser.less
+++ b/styles/less/ui/item-browser/item-browser.less
@@ -279,11 +279,11 @@
                 }
 
                 &[data-sort-type='ASC']:after {
-                    content: '\f0d7';
+                    content: '\f884';
                 }
 
                 &[data-sort-type='DESC']:after {
-                    content: '\f0d8';
+                    content: '\f160';
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


styles/less/ui/item-browser/item-browser.less
```diff
@@ -279,11 +279,11 @@
                 }
 
                 &[data-sort-type='ASC']:after {
-                    content: '\f0d7';
+                    content: '\f884';
```
Dapoulp, line 282: It feels wierd to have 2 times an arrow down, this should be: `\f161`

Dapoulp (changes requested):

chrisryan10:
Changed the icons to do the following:

Up arrow for ascending (A-Z, 1-10), small bar at top, large bar at bottom.
Down arrow for descending (Z-A,10-1), large bar at top, small bar at bottom.

chrisryan10:
Ok, that makes sense. I've never used the Font Awesome stuff before, so I was just searching for whatever worked. :)

Although, I wish we could split the fields into Alpha vs Numeric to make it even more clear.

PR #1148, opened 2025-09-04
This is psuedo fix to the incorrect resource usage when using multipl…

### LINKED ISSUE #1144: [Bug] Using 2 or more experiences in a roll does not handle Hope correctly.
**Describe the bug**
If you make a roll using 2 or more experiences, the handling of the amount hope consumed and potentially recovered by the result is incorrect. Both roll with Fear and roll with Hope do not give the correct Hope value.

Note: Using 1 experience works as expected.

**To Reproduce**

Create a character with at least 2 experiences.
Set your Hope to 4.
Do a duality roll (e.g. weapon attack) and select 2 experiences.
If you get a roll with Fear, 3 Hope will be removed.
If you get a roll with Hope, 3 Hope will be removed and no Hope added.
If you get a critical, 3 Hope will be removed and no Hope added.

**Expected behavior**
2 Hope should be removed on a roll with Fear.
For a roll with Hope or a critical, 2 Hope should be removed and 1 Hope added OR just 1 Hope removed as net result.

chrisryan10:
…e experiences.

## Description

CostField.getRealCosts is being called multiple times - after each experience is selected. The totals were being added/merged incorrectly. This psuedo-fix resets the totals before doing the merged total costs. Note that the values for when a crit/with hope/with fear roll being pushed into the updates needed changing as well.

Ideally this draft PR is used to sort out a better solution.  

- Fixes #1144 

## Type of Change

- [ x] Bug fix
- [x ] Other (please describe): See comments above

## How Has This Been Tested?

- [ x] Manual testing

Tested for 0,1,2,3 experiences used, and results of crit, with Hope, with Fear.

### DIFF
--- a/module/data/fields/action/costField.mjs
+++ b/module/data/fields/action/costField.mjs
@@ -95,6 +95,14 @@ export default class CostField extends fields.ArrayField {
 
     static getRealCosts(costs) {
         const realCosts = costs?.length ? costs.filter(c => c.enabled) : [];
+
+        // Reset totals in case of repeat calls.
+        realCosts.forEach(c => {
+            c.scale = c.scale ?? 0;
+            c.step = c.step ?? 1;
+            c.total = c.value + c.scale * c.step;
+        });
+
         let mergedCosts = [];
         realCosts.forEach(c => {
             const getCost = Object.values(mergedCosts).find(gc => gc.key === c.key);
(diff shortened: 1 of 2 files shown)


chrisryan10:
Rerolls might be broken with this "fix".

Dapoulp:
I don't think the problem is where you modified files.
I'll take a look

PR #1166, opened 2025-09-06
[Feature] 1110 - Armor/Weapon Custom Features

### LINKED ISSUE #1110: [Feature] Add Custom Weapon Features
**Description**
Currently can't create weapon features, only select pre-configured ones.

This breaks the promise that everything that the devs can do in the system, world item wise, can be done without having to code or changing the system from a dev perspective. 

There should be an option for users to add custom features through their in-world settings like there is for domains.

WBHarry:
Closes #1110 
Added homebrew for armor and weapon features


chrisryan10 (approved):
When released, this should be documented (ideally).

PR #1174, opened 2025-09-09
[PR] Fix/filter error item browser config

chrisryan10:
## Description

No issue created. I noticed a problem with the subclasses display in the character creation. Unfortunately in commit f1b6d3851dbed1559d88ecebc71d1bb537fd8b8b an error was made in merging? changes and we ended up with the subclass filter in the wrong location.

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

- [x] Manual testing

### DIFF
--- a/module/config/itemBrowserConfig.mjs
+++ b/module/config/itemBrowserConfig.mjs
@@ -370,19 +370,6 @@ export const typeConfig = {
                 label: 'DAGGERHEART.ITEMS.Subclass.spellcastingTrait'
             }
         ],
-        filters: []
-    },
-    beastforms: {
-        columns: [
-            {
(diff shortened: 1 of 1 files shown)


module/config/itemBrowserConfig.mjs
```diff
@@ -370,19 +370,6 @@ export const typeConfig = {
                 label: 'DAGGERHEART.ITEMS.Subclass.spellcastingTrait'
             }
         ],
-        filters: []
-    },
-    beastforms: {
```
chrisryan10, line 375: Note that beastforms isn't lost (expand below to see), but it had been duplicated.

PR #1041, opened 2025-08-21
Fix/Bug-with-rendering-inventory-descriptions

joaquinpereyra98:
## Description

- Performs enrichment and injection of descriptions at an earlier stage of rendering.
- `extended` class are applied to accordions prior to synchronization.

## Type of Change

Please check the relevant options:

- [X] Bug fix
- [ ] New feature
- [X] Code cleanup/refactor
- [ ] Documentation update
- [ ] Test coverage
- [ ] Dependency update
- [ ] Configuration change
- [ ] Other (please describe):

## How Has This Been Tested?

Please describe the tests you ran to verify your changes:

- [X] Manual testing
- [ ] Other:

PR #1176, opened 2025-09-09
[Community PR] Fix expanded item summary behavior when re-rendering

### LINKED ISSUE #1173: [Bug] Item Summary expand animations play on every rerender
**Describe the bug**
On an actor sheet, if you have items expanded (manually or automatic) then any sheet rerender will cause the expand animation to play again.

**To Reproduce**
1) Create a character, with items in it
2) Expand some items, make sure you're looking at them
3) Change something on the actor like HP or Hope
4) Witness the wobble

**Expected behavior**
I expect the animation to be skipped if its a re-render

**Screenshots**

https://github.com/user-attachments/assets/3b65a95a-327c-49c7-b618-4168e32f54ac

**Setup Information:**
 - OS: Windows
 - Browser: Chrome
 - Foundry Version 13.348
 - System Version Development Branch

CarlosFdez:
## Description

This PR prevents item description expand animations from replaying between re-renders, and also fixes scroll position restoration between re-renders when expanded items are present.

Internally it moves the preservation of extend state to _preSyncPartState(), which occurs before the new element is mounted. It also adds the description enrichment to _renderHTML(), which also occurs before mounting but is an async function. Having the descriptions ready and expanded before mounting skips the animation entirely and allows scroll restoration to work.

- Closes #1173 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Code cleanup/refactor
- [ ] Documentation update
- [ ] Test coverage
- [ ] Dependency update
- [ ] Configuration change

## How Has This Been Tested?

- [ ] Manual testing
- [ ] Other:

https://github.com/user-attachments/assets/f6841686-7f45-4fdc-8326-69a7c5dd6f24

## Checklist

- [x] I have made corresponding changes to the documentation
- [x] New and existing tests pass locally with my changes

### DIFF
--- a/module/applications/sheets/api/application-mixin.mjs
+++ b/module/applications/sheets/api/application-mixin.mjs
@@ -203,11 +203,19 @@ export default function DHApplicationMixin(Base) {
             this.relatedDocs.filter(doc => doc).map(doc => delete doc.apps[this.id]);
         }
 
+        /** @inheritdoc */
+        async _renderHTML(context, options) {
+            const rendered = await super._renderHTML(context, options);
+            for (const result of Object.values(rendered)) {
+                await this.#prepareInventoryDescription(result);
+            }
+            return rendered;
+        }
+
         /**@inheritdoc */
         async _onRender(context, options) {
             await super._onRender(context, options);
             this._createTagifyElements(this.options.tagifyConfigs);
-            await this.#prepareInventoryDescription(context);
         }
 
         /* -------------------------------------------- */
@@ -215,8 +223,8 @@ export default function DHApplicationMixin(Base) {
         /* -------------------------------------------- */
 
         /**@inheritdoc */
-        _syncPartState(partId, newElement, priorElement, state) {
-            super._syncPartState(partId, newElement, priorElement, state);
+        _preSyncPartState(partId, newElement, priorElement, state) {
+            super._preSyncPartState(partId, newElement, priorElement, state);
             for (const el of priorElement.querySelectorAll('.extensible.extended')) {
                 const { actionId, itemUuid } = el.parentElement.dataset;
                 const selector = `${actionId ? `[data-action-id="${actionId}"]` : `[data-item-uuid="${itemUuid}"]`} .extensible`;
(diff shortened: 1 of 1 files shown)


joaquinpereyra98:
I would recommend you check out my draft #1041. Above all, to achieve all the promises in the same amount of time.
But my recommendation for the active team was to review the entire rendering process to simplify all methods and reduce processing time.

CarlosFdez:
They look to be about the same. Seems we came to the same approach! What's wrong with the draft, other than changing the selector may cause a regression?

The only other change that I notice is that you used Promise.all() instead of a loop. I would have to benchmark the difference. I imagine there is no difference for world actors, but I'd have to see how fromUuid behaves in compendiums when you hit the same pack or actor multiple times. Looking at foundry code (`CompendiumCollection#getDocuments()`), I suspect it may actually circumvent caching. I think you need a smarter batch retrieval (that groups by pack then pack.getDocuments()) in order to avoid those issues.

While I suspect the actor to already be cached, Promise.all() *may* still run into performance impacts with embedded links coming from the same pack. I don't expect these performance impacts to keep happening though due to said caching.

> Above all, to achieve all the promises in the same amount of time.

If you mean relative to live, 10 linear promises is equivalent to two sets of 5 linear promises. There should be minimal to no performance impacts unless the overhead of querySelectorAll() dominates even on smaller DOM hierarchies. If you meant equal to yours, we'd have to benchmark, caching makes things complicated. Honestly sheet rendering is so snappy already that I didn't benchmark.

An easier improvement may be to make a more narrow selector with tag and/or class names in the querySelectorAll(). I remember reading that query selectors on only attributes are relatively slow.

> But my recommendation for the active team was to review the entire rendering process to simplify all methods and reduce processing time.

It wouldn't be a simplification, but only rendering summaries that are open would actually improve rendering time! That said, the sheet rerender is already very snappy, and the list of tasks look enormous. I think its reasonable for them to prioritize.

I really only set out to fix a bug. I would be down to do this but I'd like permission, since its more involved and I don't have that level of trust.

WBHarry (approved):
Works great. Thank you for the fix 🤖 :+1:

PR #1184, opened 2025-09-14
[Feature] 613 - Countdown Improvements

### LINKED ISSUE #613: [Feature] Countdown Improvements
The countdown window should go transparent like the fear tracker, at least when in the minimized view.

WBHarry:
Closes #613

### DIFF
--- a/module/applications/ui/countdownEdit.mjs
+++ b/module/applications/ui/countdownEdit.mjs
@@ -13,15 +13,14 @@ export default class CountdownEdit extends HandlebarsApplicationMixin(Applicatio
         this.hideNewCountdowns = false;
     }
 
-    get title() {
-        return game.i18n.localize('DAGGERHEART.APPLICATIONS.CountdownEdit.title');
-    }
-
     static DEFAULT_OPTIONS = {
         classes: ['daggerheart', 'dialog', 'dh-style', 'countdown-edit'],
         tag: 'form',
         position: { width: 600 },
-        window: { icon: 'fa-solid fa-clock-rotate-left' },
+        window: {
+            title: 'DAGGERHEART.APPLICATIONS.CountdownEdit.title',
+            icon: 'fa-solid fa-clock-rotate-left'
+        },
         actions: {
             addCountdown: CountdownEdit.#addCountdown,
             toggleCountdownEdit: CountdownEdit.#toggleCountdownEdit,
@@ -83,8 +82,7 @@ export default class CountdownEdit extends HandlebarsApplicationMixin(Applicatio
     canPerformEdit() {
         if (game.user.isGM) return true;
 
-        const noGM = !game.users.find(x => x.isGM && x.active);
-        if (noGM) {
+        if (!game.users.activeGM) {
             ui.notifications.warn(game.i18n.localize('DAGGERHEART.UI.Notifications.gmRequired'));
             return false;
         }
@@ -109,7 +107,7 @@ export default class CountdownEdit extends HandlebarsApplicationMixin(Applicatio
 
     static async updateData(_event, _, formData) {
         const { hideNewCountdowns, ...settingsData } = foundry.utils.expandObject(formData.object);
-        
+
         // Sync current and max if max is changing and they were equal before
         for (const [id, countdown] of Object.entries(settingsData.countdowns ?? {})) {
             const existing = this.data.countdowns[id];
@@ -119,7 +117,7 @@ export default class CountdownEdit extends HandlebarsApplicationMixin(Applicatio
             } else {
                 countdown.progress.current = Math.min(countdown.progress.current, countdown.progress.max);
             }
-        }       
+        }
 
         this.hideNewCountdowns = hideNewCountdowns;
         this.updateSetting(settingsData);
--- a/module/applications/ui/countdowns.mjs
+++ b/module/applications/ui/countdowns.mjs
@@ -25,7 +25,7 @@ export default class DhCountdowns extends HandlebarsApplicationMixin(Application
         window: {
             icon: 'fa-solid fa-clock-rotate-left',
             frame: true,
-            title: 'Fear',
+            title: 'DAGGERHEART.UI.Countdowns.title',
             positioned: false,
             resizable: false,
             minimizable: false
@@ -49,10 +49,6 @@ export default class DhCountdowns extends HandlebarsApplicationMixin(Application
             template: 'systems/daggerheart/templates/ui/countdowns.hbs'
         }
     };
-
-    get title() {
-        return game.i18n.localize('DAGGERHEART.UI.Countdowns.title');
-    }
 
     get element() {
         return document.body.querySelector('.daggerheart.dh-style.countdowns');


moliloo:
I still need to look at the style here and check if i need make some improvements, i will do this in the weekend

CarlosFdez:
I got a bit curious since I worked on global progress clocks, so I tried it out. Its pretty neato!

Personal opinion,, and this was a call I made in the module that I felt really bad about, but I think the default foundry UI settings is too faded to use faded-ui for an informational panel. If its possible to make a custom variant of faded-ui with a minimum opacity that is more readable, it might be easier to use while still feeling like it belongs.

Other than that, maybe it'd be handy if new countdown were created in edit mode with the name auto focused? That way they'd be faster to create during gameplay.

Ikraik:
Woah, is this the fated feature :D

CarlosFdez:
Thoughts on "New Countdown" for a just newly created countdown being a placeholder until you save it, when it becomes real? I suppose for that to work it'd have to not persist the countdown until you save it for the first time, which could be out of scope? That would allow you to omit "current' when first creating it though, which might further increase speed of creation mid improv.

Are encounter cooldowns intended to be short term ones, and narrative cooldowns intended to be long term ones? I recall having a problem separating the two in global progress clocks, and while I had a solution in mind I never had time to actually test it.

WBHarry:
We do have a design for separating Encounter and Narrative countdowns in the UI Application, but we haven't gone with it yet.
Personally I'm not sure it's neccessary, but we'll see.

For new Countdowns, you're thinking of having a 'Save' button and only updating a local copy of the Countdowns until the form is submitted via this?

CarlosFdez:
> We do have a design for separating Encounter and Narrative countdowns in the UI Application, but we haven't gone with it yet. Personally I'm not sure it's neccessary, but we'll see.

Good to hear! I imagine it might become necessary in the future tho. One of the most common requests I get is people who have a lot of countdowns who have trouble managing them. Most of the bloat seems to be long term clocks (project clocks/faction/etc). Though perhaps people rarely have as many ongoing countdowns in daggerheart?

My thought was to have short term stuff in the "widget", and all of them including long term ones in the sidebar. But I haven't properly tested that.

> For new Countdowns, you're thinking of having a 'Save' button and only updating a local copy of the Countdowns until the form is submitted via this?

Something like that yeah, but only because I imagine that implementation would be the best suited to have the initial create form be slightly different from the edit form, though it would also have the additional benefit of keeping it hidden from players until its properly initialized, so they don't see "New Countdown" on screen. You might have something else in mind. I could see an alternative solution where its hidden until first save.

From what I tested, the areas where I imagine it might cause a slight slowdown mid-session were:
1) Switching to the daggerheart sidebar
2) Having to delete the current name in order to set the new one
3) Setting min equal to max for a decreasing countdown. While you can maybe make it sync when creating a new one, I don't know if you're intending to support both increasing and decreasing. I can't imagine it being too easy to distinguish decreasing/increasing/loop without something like a font awesome icon, and even then it might be clunky.

CarlosFdez:
It looks like the way hiding is implemented doesn't account for new players getting added mid campaign. I hid the newly created countdown, then I added a user to the world and it wasn't hidden to the new player.

WBHarry:
Mm. That's true. Any new player would default to `inherit` as their ownership level until it's changed.
I don't really like the idea of adding a ''default' on the individual Countdown level tho. I feel that'll be one bridge too far with people not understanding anymore. It'd become:
1) Global Default
2) Specific Countdown Default
3) Player Specific Ownership

With the standard becoming
1) Observer
2) Inherit 
3) Inherit

WBHarry:
I have added so that if a Countdown is visible to -some- players, but not all of them, then the player colour dots are shown.
That should make it immediately obvious to the GM that the new player can see the countdown after they've added the new User. Not to mention that the `fa-eye-slash` icon would have disappeared aswell, since that's only shown if no players can see it.

image

CarlosFdez:
I guess my opinion is that a global default is more of an world setting you won't want to often change. Perhaps a button that links to the setting like what we did in the PF2e party sheet could exist but I can't imagine someone changing it all that much. Maybe if instead it modified what the next countdown was? Maybe there's some use case I haven't realized yet that the global default solves.

Anyways to check if it was reasonable I double checked foundry source and it seems `document.ownership` is a `{ default: number } & Record<string, number | undefined>`, with inherit not being stored. But I can see how it'd be a bit unwieldy with both that and a changeable global default. You'd need to setup the ui such that `ownership.default` can only be none or inherit, perhaps reframing as a hide/reveal. If you think the mid join campaign is a rare enough case, then that's fine. I was more concerned with the manual hide via permissions edit than the accidental showing (due to the pips), and even that can be made easier with a context menu or something.

chrisryan10:
Just tried this out - not extensive testing, seems fine. Not a fan of the faded Countdowns panel. 

Also, the panel needs to relocatable.

Can we have a way to add a journal link here as well?

WBHarry:
I don't think relocating the panel is in the cards. One of the biggest deals of the remake was to make it as part of the UI - IE, -not- having it movable.

I don't particularly care for the fade effect either. Atm it's tied to the same fade as the general V13 UI (so you can just remove or lessen that). Possible we just don't want it at all for it though.

chrisryan10:
I disagree with the panel not being movable. I bet it will be become a feature request pretty quick.

CarlosFdez:
It was a pretty common feature request for global progress clocks that I kinda ignored. I can agree it will happen, but there are cons to doing that, particularly with moving windows/full screen mode/etc. The most I conceded was an offset and a configurable anchor point (to work around modules), but in this case it is a system and modules should work around systems usually.

chrisryan10:
Iirc in Foundry v14, we're getting native pop out windows, so maybe that will be enough, assuming it works with that.

CarlosFdez:
Native pop out windows are a different thing for character/journal sheets on another monitor I think.

I think Foundry's default fade setting is way too low even for non-informational panels. Its also a client setting every player has to change.

module/applications/ui/countdownEdit.mjs
```diff
@@ -0,0 +1,190 @@
...
+    /** @override */
+    async _postRender(_context, _options) {
+        if (this.currentEditCountdown) {
+            setTimeout(() => {
+                const input = this.element.querySelector(
+                    `.countdown-edit-container[data-id="${this.currentEditCountdown}"] input`
+                );
+                if (input) {
+                    input.focus();
+                    input.selectionStart = input.value.length + 1;
```
CarlosFdez:
```suggestion
                    input.select();
```

Try this out. I gave it a shot locally and the workflow seemed to work pretty nicely.

WBHarry: 👍 
Good way to do it, yup.

module/data/countdowns.mjs
```diff
@@ -1,25 +1,28 @@
...
+            })
         };
     }
-
-    static CountdownCategories = { narrative: 'narrative', combat: 'combat' };
 }
 
+/* Outdated and unused. Needed for migration. Remove in next minor version. (1.3) */
 class DhCountdownData extends foundry.abstract.DataModel {
-    static LOCALIZATION_PREFIXES = ['DAGGERHEART.APPLICATIONS.Countdown']; // Nots ure why this won't work. Setting labels manually for now
```
CarlosFdez, line 17: The reason this didn't work is probably because you need to call foundry.helpers.Localization.localizeDataModel() on the data model I think. The `I18nInit` hook would be a suitable place. I think foundry calls it automatically for type data models but not for random data models here and there.

Though its fine to refer to the fields directly.

module/applications/ui/countdownEdit.mjs
```diff
@@ -0,0 +1,201 @@
...
+        super();
+
+        this.data = game.settings.get(CONFIG.DH.id, CONFIG.DH.SETTINGS.gameSettings.Countdowns);
+        this.editingCountdowns = new Set();
+        this.currentEditCountdown = null;
+        this.hideNewCountdowns = false;
+    }
+
+    get title() {
+        return game.i18n.localize('DAGGERHEART.APPLICATIONS.CountdownEdit.title');
```
CarlosFdez: This can be part of DEFAULT_OPTIONS's window prop I think.

CarlosFdez:
The stuff I mentioned are minor "if you'd like to" things, and while I still think the faded ui and the presentation of default ownership can be improved, they're minor enough. I think its better to get this in the hands of more people, from my limited testing before it seems to work pretty well. Its a neat feature.

module/applications/ui/countdownEdit.mjs
```diff
@@ -0,0 +1,201 @@
...
+                }
+            }, 100);
+        }
+    }
+
+    canPerformEdit() {
+        if (game.user.isGM) return true;
+
+        const noGM = !game.users.find(x => x.isGM && x.active);
+        if (noGM) {
```
CarlosFdez:
```suggestion
        if (!game.users.activeGM) {
```

module/applications/ui/countdowns.mjs
```diff
@@ -1,355 +1,222 @@
...
-            game.user.getFlag(CONFIG.DH.id, CONFIG.DH.FLAGS[`${this.basePath}Countdown`].position) ??
-            Countdowns.DEFAULT_OPTIONS.position;
-
-        const viewSetting =
-            game.user.getFlag(CONFIG.DH.id, CONFIG.DH.FLAGS[`${this.basePath}Countdown`].simple) ?? !game.user.isGM;
-        this.simpleView =
-            game.user.isGM || !this.testUserPermission(CONST.DOCUMENT_OWNERSHIP_LEVELS.OBSERVER) ? viewSetting : true;
-        context.simple = this.simpleView;
+    get title() {
+        return game.i18n.localize('DAGGERHEART.UI.Countdowns.title');
```
CarlosFdez: Wouldn't this replace the "Fear" title in the DEFAULT_OPTIONS?

module/data/countdowns.mjs
WBHarry: Ah, no it's actually because it doesn't work for `DataModel` type. It works for `foundry.abstract.TypeDataModel` type.
Figures that out a while ago 👅

module/applications/ui/countdownEdit.mjs
WBHarry: True. I changed it now 👍 
I've generally used the `title` getter for components. Probably something I picked up during App V1 dev or something.

module/applications/ui/countdownEdit.mjs
WBHarry: :+1: Added.

module/applications/ui/countdowns.mjs
WBHarry: Ah. Copy paste stuff that lingered.
I removed the getter and just set that translation in the `DEFAULT_OPTIONS` instead.

WBHarry:
> The stuff I mentioned are minor "if you'd like to" things, and while I still think the faded ui and the presentation of default ownership can be improved, they're minor enough. I think its better to get this in the hands of more people, from my limited testing before it seems to work pretty well. Its a neat feature.

Yah, I figure if we do change stuff there it'll probably be to just not have the fade at all. But I'm happy to leave that be for public critcism 🙃

CarlosFdez:
lgtm

PR #1196, opened 2025-09-24
[Community PR] Implement @Lookup enricher

### LINKED ISSUE #1096: [Feature] Creation of inline enrichers for common actor elements such as [@name]
**Description**
Requesting the addition of inline enrichers similar to [[lookup]] from 5e system that would query and return data when used in text fields. Ex: entering the following in a feature text box for an actor called "Capybara" would produce the output results shown below.

Text box: "The [[lookup @name]] decides to support this feature request and makes me very happy"

Output result: "The Capybara decides to support this feature request and makes me very happy"

This would allow features to include the enricher and have the name change automatically instead of requiring manual adjustments when creating homebrew new adversaries or PCs.

**Screenshots**
Image

[https://github.com/foundryvtt/dnd5e/wiki/Enrichers#lookup-enrichers](Link to 5e example)

CarlosFdez:
Implements a lookup inline enricher that can resolve formulas as inline text. For example, `@Lookup[@name]` returns the actor name of the current actor, or `@name` if there isn't one. It also refactors existing inline functions to use a new inline parser function. Closes #1096 

## Additional Comments

Although the feature was approved, there wasn't any discussion to its exact implementation. I'm happy to pivot based on the project needs. So far these are the points of contention, issues, and areas of possible improvement I've detected.

* 5e uses `[[lookup @name]]`. While the issue also requested that, 5e in general prefers [[blah]] based inlines, whereas this project does not. I called it `@Lookup` for project consistency, but let me know if this assumption is incorrect.
* I don't know how well it solves the problem, sometimes adversaries are referred to in plural such as in Group Attack. I wonder if optimizations in other areas, such as reusing the description in action, might be a better approach to solving the problem?
* An enrich parameter could facilitate description reuse, ex `@Lookup[@system.description|enrich]` in an adversary action. We would need special handling to dodge `<p>` nesting, an issue the PF2e system has to deal with on the `@Localize` inline

It didn't take me very long to implement, so no worries about anything.

### DIFF
--- a/module/documents/actor.mjs
+++ b/module/documents/actor.mjs
@@ -501,6 +501,7 @@ export default class DhpActor extends Actor {
     /**@inheritdoc */
     getRollData() {
         const rollData = super.getRollData();
+        rollData.name = this.name;
         rollData.system = this.system.getRollData();
         rollData.prof = this.system.proficiency ?? 1;
         rollData.cast = this.system.spellcastModifier ?? 1;
--- a/module/enrichers/DamageEnricher.mjs
+++ b/module/enrichers/DamageEnricher.mjs
@@ -1,29 +1,8 @@
+import { parseInlineParams } from './parser.mjs';
+
 export default function DhDamageEnricher(match, _options) {
-    const parts = match[1].split('|').map(x => x.trim());
-
-    let value = null,
-        type = null,
-        inline = false;
-
-    parts.forEach(part => {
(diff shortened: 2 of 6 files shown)


chrisryan10 (approved):
Tested, works for me. Thanks also for the code refactor/cleanup.

CarlosFdez:
Awesome! Feel free to message me on the Discord (Supe) if any followup becomes required.

Issue #1008, opened 2025-08-18
[Task][Enhancement] update default roll d20 icon with custom 2d12 roll icon


cptn-cosmo:
Currently all actions and items show the default d20 icon when hovering over the icon. The default icon should be the custom 2d12 icon we got instead.

![Image](https://github.com/user-attachments/assets/0b72f6c8-d1b9-4b76-a20f-bf49e6ef66ca)

chrisryan10:
This has been released.

PR #888, opened 2025-08-13
Fix the missing ancestries in the SRD

chrisryan10:
## Description

Fixed the missing ancestries in SRD.

- Fixes #843

## Type of Change

Please check the relevant options:

- [ x] Bug fix

## How Has This Been Tested?

Please describe the tests you ran to verify your changes:

- [x ] Manual testing

PR #777, opened 2025-08-10
Updated .gitignore and author details

chrisryan10:
Add the build directory to .gitignore

And lazjen author details.

PR #1202, opened 2025-09-29
[Community PR] Use 2d12 as hover icon for PCs and omit when feature has no uses

CarlosFdez:
## Description

Replaces the d20 svg hover icon with 2d12 on non-npc actors (everything except adversaries and environments). It also skips showing the hover icon if there are no usable actions.

Closes [#1008 ](https://github.com/Foundryborne/daggerheart/issues/1008)

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Code cleanup/refactor
- [ ] Documentation update
- [ ] Test coverage
- [ ] Dependency update
- [ ] Configuration change

## How Has This Been Tested?

- [ ] Manual testing
- [ ] Other:

image

image

image

image

### DIFF
--- a/assets/icons/dice/duality/DualityBW.svg
+++ b/assets/icons/dice/duality/DualityBW.svg
@@ -6,7 +6,7 @@
      width="42.0746mm" height="50.8071mm"
      viewBox="0 0 159 192">
   <path id="Fear"
-        fill="none" stroke="black" stroke-width="1"
+        fill="#808080" stroke="black" stroke-width="1"
         d="M 107.14,62.84
            C 107.14,62.84 52.70,137.55 52.70,137.55
              52.70,137.55 110.33,137.55 110.33,137.55
@@ -32,7 +32,7 @@
              131.77,75.21 157.51,67.23 157.51,67.23
              157.25,67.32 130.42,30.45 130.43,30.45 Z" />
   <path id="Hope"
-        fill="none" stroke="black" stroke-width="1"
+        fill="white" stroke="black" stroke-width="1"
         d="M 143.84,1.80
            C 143.84,1.80 105.59,54.92 105.59,54.92
              105.59,54.92 83.04,39.72 83.04,39.72
--- a/templates/sheets/global/partials/inventory-item-V2.hbs
+++ b/templates/sheets/global/partials/inventory-item-V2.hbs
@@ -27,7 +27,7 @@ Parameters:
         {{#if @root.isNPC}}
           <img class="roll-img d20" src="systems/daggerheart/assets/icons/dice/default/d20.svg" alt="d20">
         {{else}}
-          <img class="roll-img duality" src="systems/daggerheart/assets/icons/dice/duality/DualityBW.png" alt="2d12">
+          <img class="roll-img duality" src="systems/daggerheart/assets/icons/dice/duality/DualityBW.svg" alt="2d12">
         {{/if}}
       {{/if}}
     </div>


WBHarry:
We did look at this some time ago. Honestly, I never thought it look good though 🤔

Ikraik:
I'm not sure if that svg was planned for that... I thought we were thinking on our custom 2d12 that we did for foundryborne menu but black and white... or am I confusing it with something else?

Psitacus:
I agree with Ikraik, we just never got around to it. The one that is essentially a single d12 but split into two and with like “spiky bits”. Like we have in the server for supporters

Psitacus:
Don’t have my pc right now but this 
IMG_0348

CarlosFdez:
Well, if you have the asset send it over. That one is quite crunchy! I was just looking over the list of stuff in the milestone and saw something I could knock out before bed lmao.

WBHarry:
It's already included in the project actually
image
(Yes, the empty space in those filenames is causing me physical pain)

Psitacus:
Might need to be converted to SVG before though, I believe, and edited with the right colours. That’s what we were waiting on I think

CarlosFdez:
The image is cached and the space is small, so using the png should be fine (although webp would be better). It can always be swapped out with an SVG later.

I'm currently trying to work on the SF2e system, but once I get a moment I'll do the swapparoonie and the styling we'll likely need for it. The spiky 2d12 actually does look quite nicer.

CarlosFdez:
Updated. This could be done purely in CSS, but if we did I'd prefer to use a CSS variable rather than a LESS variable for the image used. I added a padding to the image since it looks nicer with it there imo.

cptn-cosmo:
Adversaries roll a d20 and not 2d12 however. For environments it usually depends on the action what to roll.

CarlosFdez:
@cptn-cosmo adversaries rolling a d20 is handled (its the second screenshot), but environments being weird is interesting. Are there examples you have in mind? It might be fine leaving those to be a d20 icon but if there's an easy way to check what it is, we might as well do it.

EDIT: I did notice that the description is vague if skimmed though, so I made it clearer.

cptn-cosmo:
Environment Features don't usually have a roll. It might be best to not show any icons on hover there or use a chat bubble icon instead to send the feature to chat, as no roll i s being made by default.

Ikraik:
hmm, it is using the colored variant.. to match it with other's it needed to be black and white. Lemme see if I can add that to this branch for a test.

Ikraik:
Added the SVG files for manipulation. Let me know if you need anything else.

CarlosFdez:
I think while colored looks nice (and imo its fine if PCs aren't consistent with adversaries), I tried it out. Unfortunately the SVG isn't filled in: 

image

Looks like this with the png

image

> Environment Features don't usually have a roll. It might be best to not show any icons on hover there or use a chat bubble icon instead to send the feature to chat, as no roll i s being made by default.

At first I thought that could apply to every feature except attacks, but it looks like most actions pop up a selection for the real action you want to do which may or may not require a roll. Chat bubble isn't quite the right icon there (there is a separate chat bubble), but I think it'd be worthwhile to not show any hover icon if there isn't an action to perform (EX: Haunted City > Buried Knowledge). I'll work on that later!

CarlosFdez:
Changed out the SVG. @cptn-cosmo I added a check to see if the action is usable, so that the icon won't change on hover if clicking on it won't do anything. Unfortunately actionsList is sometimes an array and sometimes a collection, and sometimes the item isn't even an item but an action instead, so the `or` in the handlebars is a bit gnarly.

image

@Ikraik I used the PNG, but if you want me to I can set the fill values of the svg to be black and white and then use that. You said you added it for manipulation though, so I don't know if you want me to do that.

Ikraik:
@CarlosFdez yes that's the intention with a good SVG. I had added vector data for both fear and hope side for manipulation and only the attributes needs to be changed (which I think is fill, stroke and stroke-width) This would make it versatile down the line but yeah that's the intention. Manipulate it within code however you see fit and let us know if it worked better.

CarlosFdez:
I changed the colors and pushed it up but it still leaves something to be desired

image

How about using grey for fear instead of black, like so?

image

CarlosFdez:
I swapped the fear to grey, it also contrasts better in light mode.

Ikraik:
Great, is it similar in value to other grey icons?

CarlosFdez:
Not sure what you mean by grey icons, but if you mean the grey text nearby, that has alpha so I can't check directly, but I did use a screen color picker and the grey im using is a bit brighter than the other greys. The brightest they go is around `#777` and the grey I used was `#888`, so I can maybe go a smidge darker.

here is #808080
image

Ikraik (approved):
LGTM 👍 
Tested the feature across and works in areas where expected.

Issue #1120, opened 2025-08-29
[Feature] Improve Loadout card view and open it when clicking the name of a card in the sidebar


cptn-cosmo:
**Description**
Currently hovering the cursor over any sidebar item changes it to be interactable, but only the specific buttons actually do something. It would be nice if clicking the name (anything in a loadout item that is not a button) would open the "edit" view (which should be slightly improved, see below). This is easier than going through the menu and a bit more intuitive to reference an action before using it, without sending it to chat.
It is not very discoverable that the descriptions are currently only available in the loadout tab *if* you toggle it to the list view and expand the entries.

**View improvements**
Ideally we add a "read-only" mode to the view, but even the current one would work with some improvements:

- The Description Field should use the full available space to display more text
- The text formatting in the Description field should actually work (with spaces, etc.)
- If possible the frame around the description field could also be removed, you can see from the tab bar already that you are in the Description and it would leave even more space for content.

PR #1214, opened 2025-10-12
[Community PR] Remove item description fieldset

CarlosFdez:
## Description

Removes the description fieldset from item descriptions, to satisfy one of the "prerequisites" of https://github.com/Foundryborne/daggerheart/issues/1120. Also incidentally fixes showing artist attribution.

image

image

image

### DIFF
--- a/module/applications/sheets/items/domainCard.mjs
+++ b/module/applications/sheets/items/domainCard.mjs
@@ -39,6 +39,7 @@ export default class DomainCardSheet extends DHBaseItemSheet {
         const context = await super._prepareContext(options);
         context.domain = CONFIG.DH.DOMAIN.allDomains()[this.document.system.domain];
         context.domainChoices = CONFIG.DH.DOMAIN.orderedDomains();
+
         return context;
     }
 }
--- a/styles/less/global/prose-mirror.less
+++ b/styles/less/global/prose-mirror.less
@@ -4,13 +4,10 @@
 .application.daggerheart {
     prose-mirror {
         height: 100% !important;
-        flex: 1;
 
         .editor-menu {
             background-color: transparent;
-            padding: 0 0 8px 0;
         }
-
         .editor-content {
             scrollbar-width: thin;
             scrollbar-color: light-dark(@dark-blue, @golden) transparent;
--- a/styles/less/global/tab-description.less
+++ b/styles/less/global/tab-description.less
@@ -7,7 +7,7 @@
         flex-direction: column;
         height: -webkit-fill-available !important;
         overflow-y: hidden !important;
-        padding: 10px 20px 0 20px;
+        padding-top: 10px;
 
         prose-mirror.active + .artist-attribution {
             display: none;


CarlosFdez:
There are further improvements possible, but they're a bit more involved. My immediate thought is making the description content (when inactive) flex instead of absolute position so that the artist attribution would be part of the scrollable content. That would clean up the book of grynn example.

styles/less/sheets/items/item-sheet-shared.less
```diff
@@ -10,4 +10,19 @@
...
+        padding: 0 20px;
+        .prosemirror {
+            flex: 1;
+            &.inactive .editor-content {
+                padding-top: 10px;
+            }
+            &.active + .artist-attribution {
+                display: none;
+            }
+        }
```
moliloo: We have a file to handle prosemirror styles, you can find it at `styles/less/global/prose-mirror.less`, i suggest you to move this style to that file

styles/less/sheets/items/item-sheet-shared.less
```diff
@@ -10,4 +10,19 @@
         font-family: @font-body;
         color: light-dark(@chat-blue-bg, @beige-50);
     }
+
+    .tab.description.active {
+        display: flex;
+        flex-direction: column;
+        padding: 0 20px;
```
moliloo: We have a file to description tab, you can find it at `styles/less/global/tab-description.less`, you can move this style to there

styles/less/sheets/items/item-sheet-shared.less
CarlosFdez: These prosemirror styles aren't general ones, but ones quite specific to the description tab. Changing it everywhere might break everywhere else. flex: 1 and the artist attribution ones might be safe to move, but the padding-top can't really be something we move to there.

I could move that to the tab description scss though. I'll also see how that ends up looking elsewhere just in case that's somehow an improvement though.

styles/less/sheets/items/item-sheet-shared.less
CarlosFdez: Done

styles/less/sheets/items/item-sheet-shared.less
CarlosFdez: I think I found a way to reduce tab specific prosemirror overrides with less unintentional breakage. It did have effects on other sheets though, but I tried to reduce their impact. What's your opinion on this?

image

I kept the artist attribution in the description tab styles though since its not actually a prosemirror style.

moliloo: > It did have effects on other sheets though, but I tried to reduce their impact. What's your opinion on this?
Can you tell to me what side effects are you talking about? All i track at first here is Class item having two attributions labels, but i think its not a behavior made by you

CarlosFdez: I only see a single attribution label on class, where's the second?

Originally I only edited the prosemirror style on the description tab to compensate for the fieldset removal. Moving it to `prose-mirror.less` as is will affect other uses such as the biography or action descriptions. 

I added 10 top padding to the description tab instead. However I noticed while the text has no top margin, the menu did, so I removed the editor's top margin so that it lines up. But that means that editing in a fieldset is a bit more flush than it was before.

I can instead go the other way and add the top padding to the editor content globally like you originally suggested. Other uses will look like this if I do:

image

image

Let me know what direction you prefer! The difference is the padding between the fieldset label and the content.

moliloo: > I only see a single attribution label on class, where's the second?
I checked again and its just me missunderstanding the thing, ignore this for now

> Let me know what direction you prefer! The difference is the padding between the fieldset label and the content.
The padding top its something i dont like, cheking your pr i think your first idea to avoid this style in a global scope seems to be the best so far, when i suggest to move your style, it was more to move to sort into files then move to change scope

CarlosFdez: Alrighty! Doing it like this then. There are some changes I want to do to the editor menu, but I'll save that for a future PR to not poison this one with extra details.

moliloo: Okay, you can open a issue about this changes in the editor to talk about, because we just make a basic style into the buttons

PR #1232, opened 2025-10-25
[Feature] 1220 - CharacterCreation Order

### LINKED ISSUE #1220: [Feature] Character Creation could be improved
After running three session zero's with Foundrybourne i unfortunately noticed a pattern during character creation where they would go through the process but because of the way its laid out everyone would end up doing it on their own rather than in the way the rules intend.

Image

As you can see in Foundrybourne it goes Ancestry, Community, Class, Traits, Experience, Domain Cards, and Equipment.

This isn't bad but in Daggerheart it already has its own order laid out in the Book and Free content which clearly goes Class, Ancestry, Community, Traits, Equipment, Background, Experiences, Domain Cards, Connections

Image

I think following the rules would probably be the best way of doing it as they had their reasoning's for making it this order.

Overall I think that Character Creation should have some way of being connected as a big part of Daggerheart is the group Character Creation where everyone discusses their choices, I think this could be improved by making a system that allows all the Players and GM to see a screen where Players must select their options before advancing together.

I'm sure that is a little too ambitious for Foundry, but i think at minimum making Character Creation align more with rules would be a great benefit.

WBHarry:
Closes #1220 

Improved order of steps so class is now first.

### DIFF
--- a/module/applications/characterCreation/characterCreation.mjs
+++ b/module/applications/characterCreation/characterCreation.mjs
@@ -83,9 +83,9 @@ export default class DhCharacterCreation extends HandlebarsApplicationMixin(Appl
 
     static PARTS = {
         tabs: { template: 'systems/daggerheart/templates/characterCreation/tabs.hbs' },
+        class: { template: 'systems/daggerheart/templates/characterCreation/tabs/class.hbs' },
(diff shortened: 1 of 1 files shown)


chrisryan10 (approved):
Yep, works as expected!

Issue #1237, opened 2025-11-04
[Feature] Improve appearance of wrapping labels in sheet sidebar when the attack is long


CarlosFdez:
**Describe the bug**
When an attack includes modifiers and perhaps a very close or very far range, the labels wrap in a pretty unappealing way.

<img width="287" height="165" alt="Image" src="https://github.com/user-attachments/assets/a11c43fa-7677-4dc5-aa00-e499ea688a28" />

<img width="279" height="80" alt="Image" src="https://github.com/user-attachments/assets/86d4cd85-596b-43e1-bdf1-68356b29c6d6" />

In development it seems to be a bit better, but I think we can maybe get better still:

<img width="276" height="198" alt="Image" src="https://github.com/user-attachments/assets/369f939f-6737-4f14-9e2c-910ab041ff03" />

CarlosFdez:
I initially started creating this as a bug and then I realized bug wasn't quite the right word. Can someone change the PR type for me pretty please?

PR #1238, opened 2025-11-04
[Feature] Adjust sidebar appearance when labels are long

CarlosFdez:
I split the template into a compact variant, separated out the styling for it, and made it a grid instead. I also fixed some visual quirks with hovering that were in the same parts of code, now the hover image and the cursor pointer only apply when you're actually hovering over the clickable thing as well.

Closes https://github.com/Foundryborne/daggerheart/issues/1237

image


Ikraik:
LGTM but adding Molilo to approval check as well.

PR #1254, opened 2025-11-10
[Fix] LightMode Touchups

WBHarry:
Some fixes to places I found where the colour was too bright in lightmode.

In CharacterCreation & Levelup:
**Before**
image
**After**
image

DaggerheartMenu
**Before**
image
**After**
image

### DIFF
--- a/styles/less/dialog/character-creation/selections-container.less
+++ b/styles/less/dialog/character-creation/selections-container.less
@@ -448,7 +448,7 @@
                         white-space: nowrap;
                         border: 1px solid light-dark(@dark-blue, @golden);
                         border-radius: 6px;
-                        color: light-dark(@beige, @dark);
+                        color: @dark;
                         background-image: url('../assets/parchments/dh-parchment-light.png');
                         padding: 0 2px;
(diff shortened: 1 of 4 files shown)


styles/less/dialog/character-creation/selections-container.less
```diff
@@ -448,7 +448,7 @@
                         white-space: nowrap;
                         border: 1px solid light-dark(@dark-blue, @golden);
                         border-radius: 6px;
-                        color: light-dark(@beige, @dark);
+                        color: @dark;
```
Ikraik, line 451: Is there a reason why you went with dark instead of light-dark?

WBHarry: Yeah, the colour was good in both cases 😆

WBHarry: The alternative would be writing `light-dark(@dark, @dark)`

Ikraik: Then that would resolve it :D

PR #1283, opened 2025-11-14
Add minor styling fixes

CarlosFdez:
Fixes item descriptions on firefox, and makes the biography outer scroll not happen on the default size. The side padding for the biography is 10px, so I made the bottom match.


WBHarry (approved):
🤖 :+1:

PR #1284, opened 2025-11-14
[Fix] Party Fixes

WBHarry:
- Fixed so that deleting actors in the world also removes them from any parties they are part of.
- Changed so that Companions and Adversaries can be added to the party. They are not able to be part of Group Rolls or Team Tag Rolls. Their resource displays have been fixed.
- Fixed so Party.Inventory has the edit and delete button properly
- Fixed so inventory-item gives the right right-padding when the context menu isn't present
- Fixed so that the leader of a group roll generates hope/fear when they make their roll.

I haven't fixed so reroll correctly adjusts resources on a leader's group roll yet. Realisticly, we'll have to redo some design and offer rerolling of individual dice to satisfy the system's mechanics, and I'm not doing that just now.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -2624,7 +2624,6 @@
                 "duplicateCard": "You cannot select the same card more than once.",
                 "duplicateCharacter": "This actor is already registered in the party members list.",
                 "onlyCharactersInPartySheet": "You can only drag characters, companions and adverasries to the party sheet.",
-                "onlyLinkedActorsInPartySheet": "Actors in the ",
                 "notPrimary": "The weapon is not a primary weapon!",
                 "notSecondary": "The weapon is not a secondary weapon!",
                 "itemTooHighTier": "The item must be from Tier1",
--- a/module/applications/sheets/actors/party.mjs
+++ b/module/applications/sheets/actors/party.mjs
@@ -280,18 +280,18 @@ export default class Party extends DHBaseActorSheet {
         });
     }
 
-    get partyMembersForRoll() {
-        return this.document.system.partyMembers.filter(x => Party.DICE_ROLL_ACTOR_TYPES.includes(x.type));
-    }
-
     static async #tagTeamRoll() {
-        new game.system.api.applications.dialogs.TagTeamDialog(this.partyMembersForRoll).render({
+        new game.system.api.applications.dialogs.TagTeamDialog(
+            this.document.system.partyMembers.filter(x => Party.DICE_ROLL_ACTOR_TYPES.includes(x.type))
+        ).render({
             force: true
         });
     }
 
-    static async #groupRoll(params) {
-        new GroupRollDialog(this.partyMembersForRoll).render({ force: true });
+    static async #groupRoll(_params) {
+        new GroupRollDialog(
+            this.document.system.partyMembers.filter(x => Party.DICE_ROLL_ACTOR_TYPES.includes(x.type))
+        ).render({ force: true });
     }
 
     /**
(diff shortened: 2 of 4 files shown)


CarlosFdez:
I think we can maybe avoid said migration and prune. We can maybe add something to ForeignDocumentUUIDArrayField (an optional prop) so that it filters out any creature that don't exist in the initialize(). Because of the way the update works, that means they'll be stripped out next party update.

WBHarry:
If that works, then that's good.
I seem to recall there being an issue with the actors not being available during initialize, making it impossible.

CarlosFdez:
Let me try that in a PR myself. I remember seeing it was possible but I wasn't sure if I should do it globally or just to party. Just to party for now is my thought.

CarlosFdez:
I gave the prune a shot in https://github.com/Foundryborne/daggerheart/pull/1285. Turns out it was already pruned though, just elsewhere. I had already handled the ghost characters during development, unless there's a case I missed.

WBHarry:
The issue occured when you deleted a character from the world while it was present in the party, then you opened the party to use it. 
I'll check the PR later.

WBHarry:
Pulled in main that now contains your pruning fix. I removed the migration since it works :+1:

lang/en.json
```diff
@@ -2623,7 +2623,8 @@
                 "cardTooHighLevel": "The card is too high level!",
                 "duplicateCard": "You cannot select the same card more than once.",
                 "duplicateCharacter": "This actor is already registered in the party members list.",
-                "onlyCharactersInPartySheet": "You can drag only characters to a party sheet.",
+                "onlyCharactersInPartySheet": "You can only drag characters, companions and adverasries to the party sheet.",
+                "onlyLinkedActorsInPartySheet": "Actors in the ",
```
CarlosFdez: wat is dis?

module/applications/sheets/actors/party.mjs
```diff
@@ -455,7 +463,7 @@ export default class Party extends DHBaseActorSheet {
         const data = foundry.applications.ux.TextEditor.implementation.getDragEventData(event);
         const item = await foundry.utils.fromUuid(data.uuid);
 
-        if (item instanceof DhpActor) {
+        if (item instanceof DhpActor && Party.ALLOWED_ACTOR_TYPES.includes(item.type)) {
```
CarlosFdez: Maybe rename item to document? I was legitimately confused here.

module/applications/sheets/actors/party.mjs
```diff
@@ -276,14 +280,18 @@ export default class Party extends DHBaseActorSheet {
         });
     }
 
+    get partyMembersForRoll() {
```
CarlosFdez: Design wise I'd rather filter in the two uses you call this in or at least make this # private, but its not really a big deal..

module/data/actor/party.mjs
```diff
@@ -31,7 +31,7 @@ export default class DhParty extends BaseDataActor {
         // Register this party to all members
         if (game.actors.get(this.parent.id) === this.parent) {
             for (const member of this.partyMembers) {
-                member.parties?.add(this.parent);
+                member?.parties?.add(this.parent);
```
CarlosFdez: as of a different change, member cannot ever be falsey anymore. But I guess in javascript the extra caution can't hurt?

module/documents/actor.mjs
```diff
@@ -85,6 +85,17 @@ export default class DhpActor extends Actor {
         this.updateSource({ prototypeToken });
     }
 
+    /**@inheritdoc */
+    async _preDelete() {
```
CarlosFdez: I think we can do without this now (there's already a re-render in on-delete, and pruned stuff wont' show between re-renders), but if you want to keep it for data sanity, it might be good to move it to _onUpdate() and check if the current user matches document.primaryUpdater

CarlosFdez: Updated the solution, thought of a better one in the interim if we're keeping the data deletion.

WBHarry: Removed :+1:

module/data/actor/party.mjs
WBHarry: Yah, used to cause an issue, but should be fine now. Removed :+1:

module/applications/sheets/actors/party.mjs
WBHarry: Moved it into the two places.

module/applications/sheets/actors/party.mjs
WBHarry: Better naming yeah. Changed it. :+1:

lang/en.json
WBHarry: My OC translation. (Don't steal)

---
I removed it tho 👅

PR #1285, opened 2025-11-14
[PR] Prune missing party members

CarlosFdez:
Prunes invalid party members in the data model during data prep. Since updates check prepared data, this means that changing party members will remove them from the source data when that occurs.


CarlosFdez:
~~I'm realizing I had already pruned them, but I guess its now pruned one level deeper.~~ actually didn't work before.

CarlosFdez:
The previous pruning didn't work actually. Looks like its because it was a getter or something maybe. Should work now though.

WBHarry (approved):
🤖 :+1:

PR #1286, opened 2025-11-15
Hide countdown panel for players when empty

CarlosFdez:
When the list of countdowns is empty and the current viewer is a player, hides the countdown panel.

https://github.com/user-attachments/assets/819cb252-7368-42be-9d1c-668b52f50f38


CarlosFdez:
I agree that it's preferable not to hide for gms. It's only really important for players cause they can't interact with it if it's empty, and without any buttons it looks weird.

For GMs we can always find tune the empty appearance.

PR #1288, opened 2025-11-15
[Fix] Beastform Fixes

WBHarry:
Patched up beastform.
- Should have fixed the issue with it getting stuck (updates were trying to run on tokens lacking id..dunno what's up with that, but I just filtered them out)
- Removed the Cancel Beastform from the context menu. Instead I'm swapping out the button for the beastform while it's active to make it really obvious.
- Scroll on the wildcard image select


Ikraik (approved):
LGTM 👍

PR #1290, opened 2025-11-15
[Fix] Item Browser Corrections

WBHarry:
- Fixed so that the lists do not fail to show because the search is empty (the common problem)
- Fixed so that hitting the clear button actually loads all the items again instead of an empty list


CarlosFdez:
lgtm!

CarlosFdez:
The HTML in the localization is a bit odd but im blanking on better alternatives right now.

PR #1292, opened 2025-11-16
[Fix] Beastform Actor Wildcard Image

### LINKED ISSUE #1129: [Bug] Beastform Breaking Wildcard on Druid Character Sheet
**Describe the bug**
I suspect that the last Foundry update damaged a fix your team had previously implemented, as this was not occurring in the past. This is a new problem as of Daggerheart build 1.1.2 and Foundry build 348 (v13). When I ran a game on August 16th, this was not a problem. As of August 30th (when I also updated both Daggerheart and Foundry prior to running a game) this became a problem. The fix in question is in regards to using Evolution or Beastform with a Druid's whose prototype token is set up with a wildcard image. The beastform transformation works just fine. However, when the druid reverts their transformation, their token is not reverted to the same setup as it had prior to the beastform transformation. Instead, the game chooses one of the images in the folder that was setup to be a wildcard image folder and then alters their prototype token making that image their prototype token image. For example, a token set up with assets/art/characters/Rowan/* as their token image (with the Randomized Wildcard Images feature clicked off) will have that changed to something like assets/art/characters/Rowan/Rowan.png. It becomes necessary to reset their wildcard token setup on the prototype token after every use of a beastform.

**To Reproduce**
1. Go to a Druid's character sheet. 
2. Click on Evolution or Beastform
3. Click on Evolution or Beastform again to undo the effect.
4. See error where their wildcard token setup has been undone and a specific image for their wildcard folder has been permanently set up as their new prototype token image.

**Expected behavior**
I expected the character sheet to return to its original setup (which it was doing in the past prior to something breaking your feature). 

**Screenshots**
Rowan's setup prior to using a beastform: 

Image

Rowan's setup after revering from a use of Beastform: 

Image

**Setup Information:**
 - OS: node.js
 - Browser Chrome
 - Foundry Version v13 b348
- System Version v.1.1.2

**Additional context**
None

WBHarry:
Closes #1129
- Fixed so that the BeastformEffect datamodel stores the PrototypeToken images.
- The individual tokens belonging to the actor store a flag of their original images that can then be restored when the beastform ends.

With this you can have three tokens out on the canvas from an Actor that has a wildcard image path, with three different pictures.
-->
Transform to a beast
--> 
Turn Back
-->
The tokens have the individual pictures they had before and the ProtoType token has the path with the wildcard.

### DIFF
--- a/module/helpers/utils.mjs
+++ b/module/helpers/utils.mjs
@@ -251,6 +251,12 @@ export const adjustRange = (rangeVal, decrease) => {
     return range[rangeKeys[newIndex]];
 };
 
+/**
+ *
+ * @param {DhActor} actor - The actor for which all tokens will run a data update.
+ * @param {string} update - The data update to be applied to all tokens.
+ * @param {func} updateToken - Optional, specific data update for the non-prototype tokens as a function using the token data. Useful to handle wildcard images where each token has a different image but the prototype has a wildcard path.
+ */
 export const updateActorTokens = async (actor, update, updateToken) => {
     await actor.prototypeToken.update({ ...update });
 
--- a/module/data/activeEffect/beastformEffect.mjs
+++ b/module/data/activeEffect/beastformEffect.mjs
@@ -6,6 +6,7 @@ export default class BeastformEffect extends BaseEffect {
         const fields = foundry.data.fields;
         return {
             characterTokenData: new fields.SchemaField({
+                usesDynamicToken: new fields.BooleanField({ initial: false }),
                 tokenImg: new fields.FilePathField({
                     categories: ['IMAGE'],
                     base64: false,
@@ -49,6 +50,7 @@ export default class BeastformEffect extends BaseEffect {
                     src: this.characterTokenData.tokenImg
                 },
                 ring: {
+                    enabled: this.characterTokenData.usesDynamicToken,
                     subject: {
                         texture: this.characterTokenData.tokenRingImg
                     }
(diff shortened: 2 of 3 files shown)


module/helpers/utils.mjs
```diff
@@ -251,15 +251,15 @@ export const adjustRange = (rangeVal, decrease) => {
     return range[rangeKeys[newIndex]];
 };
 
-export const updateActorTokens = async (actor, update) => {
+export const updateActorTokens = async (actor, update, updateToken) => {
```
CarlosFdez, line 260: Would it be possible to add a comment to this function? It was actually a bit hard for me to parse on review what this was doing. I think update is the prototype token update, and updateToken is the per token one?

CarlosFdez:
It looks good to me as far as I can read it, but I haven't tested so I'm not certain.

module/helpers/utils.mjs
WBHarry: That's fair! Added.
```
/**
 * 
 * @param {DhActor} actor - The actor for which all tokens will run a data update.
 * @param {string} update - The data update to be applied to all tokens.
 * @param {func} updateToken - Optional, specific data update for the non-prototype tokens as a function using the token data. Useful to handle wildcard images where each token has a different image but the prototype has a wildcard path.
 */
```

WBHarry:
Did the change where the original state of wether the Token has `Dynamic Token` active is also saved. The state is restored upon ending the beastform to ensure consistant restoral to the image that was before.

CarlosFdez (approved):

PR #1293, opened 2025-11-16
[Fix] Status Effect Error

WBHarry:
Fixed so that status effects are shown on ActiveEffectConfig again when useGeneric is not active.
`CONFIG.DH.GENERAL.conditions` had been changed from a constant to a function in a previous change, but it was still being used as a value here, making it fail.

### DIFF
--- a/module/applications/sheets-configs/activeEffectConfig.mjs
+++ b/module/applications/sheets-configs/activeEffectConfig.mjs
@@ -112,7 +112,7 @@ export default class DhActiveEffectConfig extends foundry.applications.sheets.Ac
                     CONFIG.DH.SETTINGS.gameSettings.appearance
                 ).showGenericStatusEffects;
                 if (!useGeneric) {
-                    partContext.statuses = Object.values(CONFIG.DH.GENERAL.conditions).map(status => ({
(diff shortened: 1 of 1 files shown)


module/applications/sheets-configs/activeEffectConfig.mjs
```diff
@@ -112,7 +112,7 @@ export default class DhActiveEffectConfig extends foundry.applications.sheets.Ac
                     CONFIG.DH.SETTINGS.gameSettings.appearance
                 ).showGenericStatusEffects;
                 if (!useGeneric) {
-                    partContext.statuses = Object.values(CONFIG.DH.GENERAL.conditions).map(status => ({
+                    partContext.statuses = Object.values(CONFIG.DH.GENERAL.conditions()).map(status => ({
```
Ikraik, line 115: I see 4 instances where it needs to be checked. Check for all references to it maybe and see it follows it well?

So far it is in actor.mjs (this follows the format above but you can check nonetheless), 
module\applications\sheets-configs\setting-active-effect-config.mjs (Line 141)
module\data\combatant.mjs (line 14)
and the current one you have. Maybe check if their implementation is correct as well? Just to make sure if it is correct.

PR #1295, opened 2025-11-16
[Fix] Downtime Actions

WBHarry:
Fixed so that downtime actions can be applied again. They were bugging out because the workflow was looking for them on the Actor, when they are in a game setting.

### DIFF
--- a/module/applications/ui/chatLog.mjs
+++ b/module/applications/ui/chatLog.mjs
@@ -141,8 +141,8 @@ export default class DhpChatLog extends foundry.applications.sidebar.tabs.ChatLo
                 ...actionType,
                 _id: foundry.utils.randomID(),
                 name: game.i18n.localize(actionType.name),
-                itemSource: {
-                    type: CONFIG.DH.ITEM.itemSourceType.restMove,
+                originItem: {
+                    type: CONFIG.DH.ITEM.originItemType.restMove,
                     itemPath: movePath,
                     actionIndex: actionIndex
                 }
--- a/module/data/chat-message/actorRoll.mjs
+++ b/module/data/chat-message/actorRoll.mjs
@@ -17,11 +17,11 @@ const targetsField = () =>
         })
     );
 
-export const itemSourceField = () =>
+export const originItemField = () =>
     new fields.SchemaField({
         type: new fields.StringField({
-            choices: CONFIG.DH.ITEM.itemSourceType,
-            initial: CONFIG.DH.ITEM.itemSourceType.itemCollection
+            choices: CONFIG.DH.ITEM.originItemType,
+            initial: CONFIG.DH.ITEM.originItemType.itemCollection
         }),
         itemPath: new fields.StringField(),
         actionIndex: new fields.StringField()
@@ -45,7 +45,7 @@ export default class DHActorRoll extends foundry.abstract.TypeDataModel {
             source: new fields.SchemaField({
                 actor: new fields.StringField(),
                 item: new fields.StringField(),
-                itemSource: itemSourceField(),
+                originItem: originItemField(),
                 action: new fields.StringField()
             }),
             damage: new fields.ObjectField(),
@@ -59,15 +59,15 @@ export default class DHActorRoll extends foundry.abstract.TypeDataModel {
         return fromUuidSync(this.source.actor);
     }
 
-    get itemAction() {
+    get actionItem() {
         const actionActor = this.actionActor;
         if (!actionActor || !this.source.item) return null;
 
-        switch (this.source.itemSource.type) {
-            case CONFIG.DH.ITEM.itemSourceType.restMove:
+        switch (this.source.originItem.type) {
+            case CONFIG.DH.ITEM.originItemType.restMove:
                 const restMoves = game.settings.get(CONFIG.DH.id, CONFIG.DH.SETTINGS.gameSettings.Homebrew).restMoves;
-                return Array.from(foundry.utils.getProperty(restMoves, `${this.source.itemSource.itemPath}`).actions)[
-                    this.source.itemSource.actionIndex
+                return Array.from(foundry.utils.getProperty(restMoves, `${this.source.originItem.itemPath}`).actions)[
+                    this.source.originItem.actionIndex
                 ];
             default:
                 const item = actionActor.items.get(this.source.item);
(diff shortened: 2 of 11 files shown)


module/applications/dialogs/downtime.mjs
```diff
@@ -178,10 +178,17 @@ export default class DhpDowntime extends HandlebarsApplicationMixin(ApplicationV
...
+            const category = this.moveData[categoryKey];
+            return Object.keys(category.moves)
+                .filter(x => category.moves[x].selected)
+                .flatMap(key => {
+                    const move = category.moves[key];
+                    return [...Array(move.selected).keys()].map(_ => ({
+                        ...move,
+                        movePath: `${categoryKey}.moves.${key}`
+                    }));
+                });
```
WBHarry, line 191: The change here is just mapping on the keys instead of values to let us add the `movePath` property which will let us easily grab the right move from the game setting later.

module/applications/ui/chatLog.mjs
```diff
@@ -132,12 +132,21 @@ export default class DhpChatLog extends foundry.applications.sidebar.tabs.ChatLo
...
+            {
+                ...actionType,
+                _id: foundry.utils.randomID(),
+                name: game.i18n.localize(actionType.name),
+                itemSource: {
+                    type: CONFIG.DH.ITEM.itemSourceType.restMove,
+                    itemPath: movePath,
+                    actionIndex: actionIndex
+                }
+            },
```
WBHarry, line 149: Sending along the new `itemSource` property into the action config so that it can ultimately be set on the chat Message.

module/data/chat-message/actorRoll.mjs
```diff
@@ -17,6 +17,16 @@ const targetsField = () =>
...
+export const itemSourceField = () =>
+    new fields.SchemaField({
+        type: new fields.StringField({
+            choices: CONFIG.DH.ITEM.itemSourceType,
+            initial: CONFIG.DH.ITEM.itemSourceType.itemCollection
+        }),
+        itemPath: new fields.StringField(),
+        actionIndex: new fields.StringField()
+    });
+
```
WBHarry, line 29: `itemPath` points the way to the downtime item in the game setting in this case, but it's intended to be generic if we expand `type` to more places where actions reside.
`actionIndex` is used, because the actual `id` of the downtime moves are changed on every reload.

Ikraik:
I feel this is a feature implement than a fix? I think there was an issue for downtime actions not fully implemented

WBHarry:
Nah. The actions have been implemented since 1.0 or sometime. They just became broken recently.
So I fixed so that the healing/restore armor etc works again.

In main currently you can't apply the healing/restore armor etc via the button because of faulty handling.

WBHarry:
I don't think there's any problem with adding features either, really. No reason we can't add random improvements. I don't want a long release cycle again.

module/data/fields/action/targetField.mjs
```diff
@@ -72,17 +72,17 @@ export default class TargetField extends fields.SchemaField {
 
     /**
      * Format actor to useful datas for Action roll workflow.
-     * @param {*} actor     Actor object to format.
+     * @param {*} token     Actor object to format.
```
CarlosFdez:
```suggestion
     * @param {*} token     Token object to format.
```

module/data/fields/action/targetField.mjs
```diff
@@ -24,9 +24,9 @@ export default class TargetField extends fields.SchemaField {
         if (!this.target?.type) return (config.targets = []);
         config.hasTarget = true;
         let targets;
-        // If the Action is configured as self-targeted, set targets as the owner.
+        // If the Action is configured as self-targeted, set targets as the owner. Probably better way than to fallback to getDependentTokens
```
CarlosFdez, line 27: there's getActiveTokens() for the current scene only, but I get the feeling that the issue is that you want to select actors, not tokens somehow.  Since theoretically a downtime action can work without any actor on any scene ever. Maybe easier said than done.

module/data/chat-message/actorRoll.mjs
```diff
@@ -48,17 +59,27 @@ export default class DHActorRoll extends foundry.abstract.TypeDataModel {
...
+            default:
+                const item = actionActor.items.get(this.source.item);
+                return item ? item.system.actionsList?.find(a => a.id === this.source.action) : null;
+        }
     }
 
     get action() {
         const actionActor = this.actionActor,
-            actionItem = this.actionItem;
+            itemAction = this.itemAction;
```
CarlosFdez: maybe `const { actionActor, itemAction } = this;` would be simpler to read. Or separate const.

module/data/chat-message/actorRoll.mjs
```diff
@@ -48,17 +59,27 @@ export default class DHActorRoll extends foundry.abstract.TypeDataModel {
         return fromUuidSync(this.source.actor);
     }
 
-    get actionItem() {
+    get itemAction() {
```
CarlosFdez: If its that you're trying to maintain convention, wouldn't this be `actionItem`? item and actor would be my preference but that seems to be the convention here.

CarlosFdez:
Is itemSource intended to be the item it came from? TBH I mixed that up with the foundry meaning of source and was confused there. How about `origin`, or `originItem` if origin can be ambiguous with the origin of an effect.

module/data/fields/action/targetField.mjs
WBHarry: After some looking,
The Token was really only needed for the "click-the-chat-message-to-pan-to-the-token-on-the-canvas" functionality. I just changed that to only be available if the action in question -is- towards a token on the canvas (which this isn't.)
So I restored the `PrototypeToken` fallback, which is now alright since we don't except there to neccessarily be a `token.id`

module/data/chat-message/actorRoll.mjs
WBHarry: Doneso :+1:

module/data/chat-message/actorRoll.mjs
WBHarry: Changed it to actionItem for now.

WBHarry:
I swapped the naming over to `originItem`, works for me :+1:

PR #1297, opened 2025-11-17
Fix styling of used action tokens for players

CarlosFdez:
Fixes the styling of used action tokens of actors not owned by the current viewer.

Before:
image

After:
image


WBHarry (approved):
🤖 :+1:

PR #1299, opened 2025-11-18
Add passive effects for each trait for use in the Druid Evolution fea…

chrisryan10:
…ture.

## Description

Added passive effects for each trait for the Druid Evolution feature.

## Type of Change

- [x] New feature

## How Has This Been Tested?

- [x] Manual testing

Create a Druid and examine the effects tab to see passive effects. Toggle to see the appropriate trait increase by 1.


WBHarry:
Hm. I'm not sure I like this, even as an interim solution. It's a -lot- of effects to make the list of them more annoying to parse.
image

chrisryan10:
Maybe. Certainly not a long term solution. However it does give an easy way to do the trait change and to track it.

What's the preference? Interim solutions like this that require no coding, or wait for a programmed fix in the future?

WBHarry:
Hm..yeah, I'm not 100% on how I -do- want the solution to be, as it will involve some new things.
Lets bring this in as a useful thing for now then :+1:

PR #1302, opened 2025-11-18
[Feature] Countdown Actions

WBHarry:
- Added Countdown Actions
- Added a System Setting for CountdownAutomation. When it's off, any countdown automation is ignored.
- Added Looping to countdowns.
- Added CharacterSpotlight automation, so a countdown can automatically progress whenever a Character takes the spotlight.
- Updated Adversaries and Environments with countdowns to use these actions.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -78,7 +78,8 @@
                     "label": "Beastform"
                 },
                 "countdown": {
-                    "defaultOwnership": "Default Ownership"
+                    "defaultOwnership": "Default Ownership",
+                    "startCountdown": "Start Countdown"
                 },
                 "damage": {
                     "multiplier": "Multiplier",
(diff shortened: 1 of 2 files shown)


CarlosFdez:
Is there a use case for a single button to spin up multiple countdowns?

image

WBHarry:
> Is there a use case for a single button to spin up multiple countdowns?

Yup. There are Dynamic Countdowns that start a Progress Countdown and a Consequence Countdown. An example in the SRD that I use it for is the Bustling Market Environment.

WBHarry:
We should add a general component to visually separate multiple Countdowns, Damage stuff and whatnot in the Action Config - I just wasn't inspired on how to do it. 🤔

CarlosFdez:
No worries. I was just thinking that by default the action should start with one countdown in the list but then I was wondering why we even needed a list in the first place.

WBHarry:
> No worries. I was just thinking that by default the action should start with one countdown in the list but then I was wondering why we even needed a list in the first place.

Went ahead and added an initial countdown when a CountdownAction is created :+1:

PR #1306, opened 2025-11-19
[Fix] RCT Countdown Positioning

### LINKED ISSUE #1277: [Bug] Countdown Dialog position conflicts with webcams and cannot be moved
Describe the bug 
When the game countdown dialog appears, it is displayed in a fixed position. This position overlaps with the player webcam feeds, obstructing them from view. The dialog window cannot be clicked and dragged to a new position.

To Reproduce Steps to reproduce the behavior:
- Go to 'Game Settings' and enable A/V (webcams).
- Ensure multiple player webcams are active and visible on the UI.

See error: The countdown dialog appears on the screen, often directly on top of the webcam feeds. It is not possible to move the dialog.

Expected behavior 
The countdown dialog should be draggable, allowing users to move it to a different position on the screen so it does not block webcams or other important UI elements.

**Screenshots**
Image

**Setup Information:**
 - OS: Windows
 - Browser chrome
 - Foundry Version last 
- System Version 121

Additional context
 This is primarily a UI/UX issue where a modal dialog's fixed position conflicts with other core UI features like the A/V feeds.

WBHarry:
Closes #1277 
Fixed so that the countdown Ui positions itself well when Foundry RCT video is enabled

https://github.com/user-attachments/assets/a5fbfc86-5da8-4903-8b50-982b887b317f

### DIFF
--- a/module/applications/ui/countdowns.mjs
+++ b/module/applications/ui/countdowns.mjs
@@ -81,14 +81,6 @@ export default class DhCountdowns extends HandlebarsApplicationMixin(Application
         return frame;
     }
 
-    /**@inheritdoc */
-    async _onFirstRender(context, options) {
-        await super._onFirstRender(context, options);
-
-        if (ui.webrtc.rendered) this.toggleRctPosition();
-        this.toggleCollapsedPosition(undefined, !ui.sidebar.expanded);
-    }
-
     /** Returns countdown data filtered by ownership */
     #getCountdowns() {
         const setting = game.settings.get(CONFIG.DH.id, CONFIG.DH.SETTINGS.gameSettings.Countdowns);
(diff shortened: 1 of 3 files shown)


CarlosFdez:
out of curiousity, since you have it working, how does the pf2e effects panel look like when web video is on? Do I have to fix it there myself?

WBHarry:
It handles it right on both Top position and Right position
**Top**
image
**Right**
image

CarlosFdez:
Huh, maybe a consequence of where its placed? I think if you put it in ui-right you might be able to avoid the rtc specific code....unless you prefer the end result of course. That's fine too.

WBHarry:
> Huh, maybe a consequence of where its placed? I think if you put it in ui-right you might be able to avoid the rtc specific code....unless you prefer the end result of course. That's fine too.

Nah, definately better to have it move naturally with Foundry's elements. I moved it over there and removed all hooks/classes that handled absolute positioning. :+1:

CarlosFdez (approved):

Issue #1198, opened 2025-09-28
[Feature] Add way to reuse text between the description and an action


CarlosFdez:
**Description**
I often see that text is duplicated between the descriptions and the action, with the exception being grimoires where the description has multiple abilities in its description. When changes are made to an existing creature's ability, the GM must be aware that both places need to be changed. This came up during a one shot my GM was hosting.

Some way to facilitate text reuse could be handy, but I don't know the best way. The things that come to mind are:
1) An empty action description reuses the item's description
2) A checkbox in the action to reuse the item's description
3) Enriching the results of a lookup enricher (#1096) or a new enricher type similar to `@Embed` to refer to a description. It could be the default value of the action description, or used for a grimoire's description to refer to the action description. We would need to avoid cyclical references though.

WBHarry:
The thing is that the text isn't always duplicated. What'll often be the case is that a Feature will have text describing two things, or a happening that is better handled via 2 actions. So in those cases the two actions will reference part of the original description respectively.

I don't really see it as a big issue. Changing descriptions on mass doesn't really seem like an issue the system needs to provide solutions for.

CarlosFdez:
If its true that most are different in subtle ways, I can see that, but in most adversaries I tried editing (grain of salt, tier 1 one shots), I noticed the exact same text in the action and the description. My GM tonight missed a spot when reflavoring a jagged knife bandit, which luckily I avoided with extra prep work when I GM'd. As a player I only ran into it reflavoring a grimoire card to have essentia flavor, but imo its not a big deal for players. It was more awkward for adversaries.

While lookup name and a feature compendium can also solve the core issue, there are times when the name is plural and I suspect those will need manual editing.

Is it the case that an action description may be left completely empty but not the description? I'm wondering if deduping is good as a fallback, since it'd be the simplest to an end user, and not require building anything complicated.

CarlosFdez:
> Is it the case that an action description may be left completely empty but not the description?

I wrote a macro to check and the answer is unfortunately yes. Fire Jar's Pour is blank. Hobbling Strike's mark stress has no description. If its intentional that they're blank, then I suppose nothing can be done, but perhaps they *should* fall back? It looks like missing action text is quite common. Its kinda weird that this description is blank in the chat.
<img width="676" height="421" alt="Image" src="https://github.com/user-attachments/assets/da1f6337-b938-4427-8131-e0976e3cf6e6" />

CarlosFdez:
[...]

PR #1310, opened 2025-11-20
[Feature] Fallback action description to item description when posted in chat

CarlosFdez:
Closes https://github.com/Foundryborne/daggerheart/issues/1198

When an action with no description is posted to chat, it will post the item's description. This does not give the action a description in the actions tab proper. 

You can test with Fire Jar (there seems to be an issue with uses right now, but its unrelated)

image


WBHarry (approved):
🤖 :+1:

PR #1316, opened 2025-11-20
[Fix] Inventory Item Transfer

WBHarry:
- Fixed so items from the inventory tab of the Party sheet can be dragged out to characters etc.

Changed so that InventoryItems (weapon,armor,consumable...) are transfered when dragged from Character to Character or Character to Party and vice versa.
If the item has a quantity we show a dialog to select how much to transfer over. The quantity is added to any existing item, otherwise a new one is made with the selected quantity. The selected quantity is detracted from the original item that was dragged.

https://github.com/user-attachments/assets/5c4c4ebe-753f-473c-8a07-5b85bac9483d

### DIFF
--- a/module/applications/sheets/api/base-item.mjs
+++ b/module/applications/sheets/api/base-item.mjs
@@ -33,9 +33,9 @@ export default class DHBaseItemSheet extends DHApplicationMixin(ItemSheetV2) {
             removeResource: DHBaseItemSheet.#removeResource
         },
         dragDrop: [
-            { dragSelector: null, dropSelector: '.tab.features .drop-section' },
+            { dragSelector: null, dropSelector: '.drop-section' },
             { dragSelector: '.feature-item', dropSelector: null },
-            { dragSelector: '.action-item', dropSelector: null }
+            { dragSelector: '.inventory-item', dropSelector: null }
         ],
         contextMenus: [
             {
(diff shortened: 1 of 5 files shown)


CarlosFdez:
Why not a single template with root: true and tag: "form" in the default options?

module/applications/sheets/api/base-actor.mjs
```diff
@@ -217,6 +217,72 @@ export default class DHBaseActorSheet extends DHApplicationMixin(ActorSheetV2) {
...
+                            if (quantityTransfered === actorItem.system.quantity) {
+                                await originActor.deleteEmbeddedDocuments('Item', [data.originId]);
+                            } else {
+                                cancel = true;
+                                await actorItem.update({
+                                    'system.quantity': actorItem.system.quantity - quantityTransfered
+                                });
+                            }
+
+                            const existingItem = this.document.items.find(
```
CarlosFdez: might be handy to have a more involved check, perhaps in a helper. I think the compendium source is probably safe but it might be also good to compare some additional data like name/description/number of actions, in case someone started with an item but then modified it.

module/applications/sheets/api/base-actor.mjs
```diff
@@ -217,6 +217,72 @@ export default class DHBaseActorSheet extends DHApplicationMixin(ActorSheetV2) {
...
+                                            quantity: quantityTransfered
+                                        }
+                                    }
+                                ]);
+                            }
+                        } else {
+                            cancel = true;
+                        }
+                    } else {
+                        await originActor.deleteEmbeddedDocuments('Item', [data.originId]);
```
CarlosFdez: If you add render: false here you'll skip a re-render on the create and it'll feel a wee bit faster.

module/applications/sheets/api/base-actor.mjs
WBHarry: Added a method for comparing items in `utils.mjs`. Using compendium source, description and name.
Number of actions will have to stay in PF2e land 😉

module/applications/sheets/api/base-actor.mjs
WBHarry: Changed it, and it does feel a bit snappier :+1:

PR #1318, opened 2025-11-21
[Feature] DiceSoNice Fonts

WBHarry:
Added a font choice to system diceSoNice settings


chrisryan10 (approved):
Approved, but I'll let you do the actual merge.

Only comment: the Preview button should probably be centered or maybe right aligned. It seems a bit awkward where it is now. If it's not an issue - merge away.

WBHarry:
Centered it :+1:

PR #1329, opened 2025-11-24
[Feature] Countdown Rols And Randomzie

WBHarry:
Fixed random starting values for Countdowns via formula.

### DIFF
--- a/system.json
+++ b/system.json
@@ -2,7 +2,7 @@
     "id": "daggerheart",
     "title": "Daggerheart",
     "description": "An unofficial implementation of the Daggerheart system",
-    "version": "1.2.5",
+    "version": "1.2.6",
     "compatibility": {
         "minimum": "13",
         "verified": "13.351",


module/applications/ui/countdownEdit.mjs
```diff
@@ -57,6 +59,9 @@ export default class CountdownEdit extends HandlebarsApplicationMixin(Applicatio
                       ? 'DAGGERHEART.UI.Countdowns.decreasingLoop'
                       : 'DAGGERHEART.UI.Countdowns.loop'
                 : null;
+            const randomizeValid = Roll.parse(countdown.progress.startFormula ?? '').some(
```
CarlosFdez: I think rolls have an isDeterministic getter you can use.

module/data/action/countdownAction.mjs
```diff
@@ -21,10 +24,26 @@ export default class DhCountdownAction extends DHBaseAction {
...
+                }
             }
         ];
 
         return updateSource;
     }
+
+    /** @inheritDoc */
+    static migrateData(source) {
+        for (let countdown of source.countdown) {
```
CarlosFdez:
```suggestion
        for (const countdown of source.countdown) {
```
🌞

module/data/fields/action/countdownField.mjs
```diff
@@ -40,40 +41,69 @@ export default class CountdownField extends fields.ArrayField {
...
+                data,
+                null,
+                {
+                    refreshType: RefreshType.Countdown
+                }
+            );
+        };
+
+        if (game.modules.get('dice-so-nice')?.active) {
+            Promise.all(
```
CarlosFdez: If you use await, then you won't need to wrap update in a method. You could await on dice so nice then do the update stuff after, since its happening with or without either way.

module/helpers/utils.mjs
```diff
@@ -445,3 +445,11 @@ export function itemIsIdentical(a, b) {
 
     return compendiumSource && name & description;
 }
+
+export function waitForDiceSoNice(message, update) {
```
CarlosFdez: Is there a reason you can't make this an async function you can await on before doing the update, that doesn't involve passing the update here?

CarlosFdez: Something like:
```js
export async function waitForDiceSoNice(message) {
    if (game.modules.get('dice-so-nice')?.active) {
        await game.dice3d.waitFor3DAnimationByMessageID(message.id);
    }
}
```

Then you can do

```js
await waitForDiceSoNice(message);
await this.updateSetting({ blah });
```

Even if you can't await in the method for whatever reason, you could still do `waitForDiceSoNice(message).then(update)`

module/applications/ui/countdownEdit.mjs
WBHarry: Changed over to this. Better to leave it toFoundry core logic yup :+1:

module/data/action/countdownAction.mjs
WBHarry: Fixed 👅

module/data/fields/action/countdownField.mjs
WBHarry: Much better yeah. My brain told me I couldn't await `Promise.All`. It lied.
Changed it :+1:

module/helpers/utils.mjs
WBHarry: Brainrot is the only reason.
Much better this way yup. Swapped over to just awaiting 🥳

Issue #1195, opened 2025-09-23
[Feature] Add labels to actors in the sidebar


CarlosFdez:
**Description**
A subtitle for actors in the actor directory that denotes what it is. Perhaps `Level 3` for a player character, `Tier 3 Solo` and such for an adversary, and `Tier 2 Event` or `Tier 3 Social` or what have you for an environment?

**Screenshots**
PF2e does this in the sidebar:

<img width="309" height="171" alt="Image" src="https://github.com/user-attachments/assets/a193f09f-c246-479f-b9de-e0fc6f707b8b" />

CarlosFdez:
If approved I'm down to implement it. Let me know.

Ikraik:
So I kept new feature requests in backlog to discuss internally first. We have a decent number to go through so will let you know if approved on our end. But you can always setup a module for now that implements this feature and once I approve it here, you can then make a community PR to merge that logic within the code 👍

CarlosFdez:
No worries, just doing a sanity check, I know feature PRs without discussion/approval cause more trouble than they help. Take your time, I see there's a lot on the backlog.

PR #1332, opened 2025-11-25
[Feature] Actor Directory subtitles

CarlosFdez:
Shows labels such as "Tier 3 Adversary" below actor names in the actor tab. 

Closes https://github.com/Foundryborne/daggerheart/issues/1195

image


WBHarry (commented):
Looks great.
One thing is that when a player has Limited access to an Adversary/Environment sheet, then I think we should omit the `Tier` information. Simply writing `Adversary` or `Environment` should do in that case.

CarlosFdez:
Hmm, I didn't consider visibility. Good point. I'll address that later.

CarlosFdez:
Turns out those things are shown in limited actors anyways. The default is none, players shouldn't be seeing those things in most cases.

I changed the label to include the adversary and environment type, however that required me to add a helper function to the context since normally the entry partial only gets the document. 
image

I'm blanking on a better name for the localization key for Tier X Y labels..

PR #1333, opened 2025-11-25
[Fix] Companion direct damage

CarlosFdez:
Makes the little checkbox show up in the companion attack configuration, and removes the console error from the formGroup helper.


WBHarry (approved):
🤖 :+1:

PR #1340, opened 2025-11-25
[Feature] EffectsDisplay

WBHarry:
Added an EffectsDisplay in the style of the PF2e system.

### DIFF
--- a/styles/less/ux/tooltip/bordered-tooltip.less
+++ b/styles/less/ux/tooltip/bordered-tooltip.less
@@ -15,7 +15,6 @@
             display: flex;
             flex-direction: column;
             text-align: start;
-            width: 100%;
             padding: 5px;
             gap: 0px;
 


WBHarry:
I upped the EffectsDisplay icon size just a bit from 40 to 46 at least. Felt it got a bit too itty bitty tiny 🙃

WBHarry:
Hm. Doesn't it look better to have the title centered?
image

image

CarlosFdez:
Looks good to me, though try to use `inert` on `<i>` used for icons, to prevent screenreaders from reading the unicode text used by font awesome.

moliloo:
> Hm. Doesn't it look better to have the title centered? image
> image

Yeah, i think its the best solution for now, its because i was having some troubles trying to style this tooltip, you can leave the title and the effect by in the cernter

moliloo:
> I upped the EffectsDisplay icon size just a bit from 40 to 46 at least. Felt it got a bit too itty bitty tiny 🙃

No problem, i reduce the size because we dont have large screen like you, so 100px is kinda big

WBHarry:
> Yeah, i think its the best solution for now, its because i was having some troubles trying to style this tooltip, you can leave the title and the effect by in the cernter

Good trick to mess with the styling live for tooltips is to just put a breakpoint in the `TooltipManager` class. Then you can hover an Effect display --> Resume past the breakpoint --> Quickly hover it again. You'll get to the breakpoint again with the tooltip on screen to mess with 😎

WBHarry:
> Looks good to me, though try to use `inert` on `<i>` used for icons, to prevent screenreaders from reading the unicode text used by font awesome.

This is something that should be done as a general sweep, as we have hundreds of `<i>` elements, and none have the global property set.

PR #1342, opened 2025-11-26
[Fix] Levelup/Delevel fixes

WBHarry:
- The delevel was insistant on deleting the picked domain cards during levelup, which failed if they were already deleted. Fixed.
- Found that classes/subclasses weren't setting their uuids correctly on levelup, which made the subclass fail to be set properly.


moliloo (approved):
Its good to merge

PR #1348, opened 2025-11-29
[Feature] 1344 - Daggerheart Menu Render Hook

### LINKED ISSUE #1344: [Feature Request] API Hook to inject custom buttons into the Daggerheart Directory
I would like to request a dedicated API command or a specific Hook that allows modules to easily add custom buttons to the Daggerheart directory UI.
Having a standardized hook would provide a clean way for module developers to integrate new tools and features directly into the system's interface.
Proposed Implementation / Usage Example:
Ideally, I would like to be able to listen to a specific hook to inject the button and define its behavior.
Here is an example of how I would like to use this feature:

/**
 * Proposed usage for the requested feature
 */
Hooks.on('daggerheart.renderDirectory', (app, html, data) => {
    
    // 1. Create the button element
    const myCustomButton = $(`
        <button class="create-custom-btn">
            <i class="fas fa-dice-d20"></i> My Feature
        </button>
    `);

    // 2. Define the code/action generated by the button
    myCustomButton.click((ev) => {
        ev.preventDefault();
        console.log("Custom Daggerheart button clicked!");
        // My module logic goes here
        new MyCustomApplication().render(true);
    });

    // 3. Inject the button into the directory
    html.append(myCustomButton);
});

WBHarry:
Closes #1344 
Added a render hook for the daggerheartMenu. This lets modules easily extend the menu for custom functionality.
Ex:
```
Hooks.on(CONFIG.DH.HOOKS.sidebarMenuRender, (_app, contentElement, _context) => {
    const testFeature = document.createElement('button');
    testFeature.innerHTML = '<i class="fas fa-dice-d20"></i> My Feature';
    contentElement.appendChild(testFeature);
    testFeature.addEventListener('click', _event => {
        console.log('TestFeature Test');
    });
});
```
image

### DIFF
--- a/module/applications/sidebar/tabs/daggerheartMenu.mjs
+++ b/module/applications/sidebar/tabs/daggerheartMenu.mjs
@@ -44,6 +44,18 @@ export default class DaggerheartMenu extends HandlebarsApplicationMixin(Abstract
     /* -------------------------------------------- */
 
     /** @inheritDoc */
+    async _onRender(context, options) {
+        super._onRender(context, options);
+
+        Hooks.callAll(
+            CONFIG.DH.HOOKS.sidebarMenuRender,
+            this,
+            this.element.querySelector('[data-application-part="main"]'),
+            context
+        );
+    }
+
+    /** @inheritDoc */
     async _prepareContext(options) {
         const context = await super._prepareContext(options);
         context.refreshables = this.refreshSelections;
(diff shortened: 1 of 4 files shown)


CarlosFdez:
I just double checked and there's already a render hook. `renderDaggerheartMenu`.

You can see it if you do `CONFIG.debug.hooks = true`

WBHarry:
Huh, I'll be darned. So Foundry actually sends out specific, named hook calls for each sidebar tab render. Cool :+1:

PR #1349, opened 2025-11-29
[Fix] 1327 - Transfer Item Bug

### LINKED ISSUE #1327: [Feature] Party Actor and character items
**Description**
When transferring an item to the party actor or vice versa it seems like it duplicates the weapon or armor (this does not happen with consumables or loot). 
If I close the actor and open it again the "duplicated" weapon or armor is gone. I am guessing it does not update correctly for those two cases.

- Another thing would be if it would be good to implement a way to rearrange items. Right now they appear as they are added, with no way to order them.

**Screenshots**

Image

WBHarry:
Closes #1327 
We weren't rerendering the inventory where the item was dragged from, causing it to still appear to be there, eventhough it had been removed. Fixed.

### DIFF
--- a/module/applications/sheets/api/base-actor.mjs
+++ b/module/applications/sheets/api/base-actor.mjs
@@ -271,7 +271,7 @@ export default class DHBaseActorSheet extends DHApplicationMixin(ActorSheetV2) {
                             cancel = true;
                         }
                     } else {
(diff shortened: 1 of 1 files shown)


CarlosFdez:
Looks like some other branch stuff slipped in..

WBHarry:
Ah dang it. Didn't start from main. Ah well. I'll bring it out clean

PR #1352, opened 2025-11-29
[Feature] Fill entire tab when editing the biography and tweak rendering when not

CarlosFdez:
What happens while editing is obvious from the video, but for viewing this attempts to do something more similar to the inventory where you have to scroll down to see each section. Something more similar to how in a longer form page you'd keep scrolling down to read everything, rather than having a bunch of small sections with scrollbars.

If this turns out to be meh, we can roll back the min height and position relative tweaks and add flex: 1.

https://github.com/user-attachments/assets/0fc6b7a3-88f8-4e70-95c2-747565caa94d


WBHarry (approved):
🤖 :+1:

PR #1355, opened 2025-12-01
[Fix] 1325 - Weapon Tooltip Damage Formula

### LINKED ISSUE #1325: [Bug] Weapon Preview does not take proficiency into account
**Describe the bug**
The preview that shows when hovering over an equipped weapon always displays damage formula [1dX + bonus + (proficiency-1)] instead of [(proficiency)dX + bonus].

**To Reproduce**
1. Equip a weapon on a character with a proficiency of 2 or greater

**Expected behavior**
Preview should show the correct damage formula.

**Screenshots**
Image

WBHarry:
Closes #1325 
- Fixed so the damage calculation is correct. 
- Fixed so the right tooltip is shown for attacks on sheets. They were broken.

### DIFF
--- a/module/data/action/damageAction.mjs
+++ b/module/data/action/damageAction.mjs
@@ -2,4 +2,17 @@ import DHBaseAction from './baseAction.mjs';
 
 export default class DHDamageAction extends DHBaseAction {
     static extraSchemas = [...super.extraSchemas, 'damage', 'target', 'effects'];
+
+    /**
+     * Return a display ready damage formula string
+     * @returns Formula string
+     */
+    getDamageFormula() {
+        const strings = [];
+        for (const { value } of this.damage.parts) {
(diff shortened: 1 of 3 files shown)


CarlosFdez:
The way getFormulaString() is set up as a static method intended to be bound is very strange. Importing a function helper or passing in an action sounds more appealing immediately, but let me see if I can think of an alternative.

CarlosFdez:
Alrighty, took a quick poke. While I've learned that binding static methods is already common for DamageField, what are your thoughts on adding a `getDamageFormula()` method to `DHDamageAction` that fetches the parts from this.damage.parts, and then calling `attack.getDamageFormula()` in the handlebars helper?

If its a getter than no handlebars helper is needed, though I'm uncertain if its a good idea to make that sort of computation a getter. Maybe?

Ikraik (approved):
LGTM 👍

WBHarry:
I moved the damage string logic over to DhDamageAction as getDamageFormula. Less confusing for sure :+1:

PR #1357, opened 2025-12-02
[Feature] Sortable inventories and adversary/environment drag/drop

CarlosFdez:
This adds support for drag resort within actor and party sheets, as well as dragging and dropping features between adversary and environment sheets. It is also a major refactor of the sheet drop handlers to take better advantage of core foundry drop functionality, so I would like to ask people familiar with system functionality to test this very very thoroughly. Enabling ignore whitespace when viewing the diff would make the actual changes easier to spot, since I reduced indentation levels while I was at it.

This should fully close https://github.com/Foundryborne/daggerheart/issues/1327

### DIFF
--- a/module/applications/sheets/actors/character.mjs
+++ b/module/applications/sheets/actors/character.mjs
@@ -882,11 +882,7 @@ export default class CharacterSheet extends DHBaseActorSheet {
             return super._onDropItem(event, item);
         }
 
-        if (item.type === 'domainCard' && !this.document.system.loadoutSlot.available) {
-            const itemData = item.toObject();
-            itemData.system.inVault = true;
-            return await this._onDropItemCreate(itemData);
-        } else if (item.type === 'beastform') {
+        if (item.type === 'beastform') {
             if (this.document.effects.find(x => x.type === 'beastform')) {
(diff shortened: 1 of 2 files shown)


CarlosFdez:
After some limited testing, adversary sheets now allow you to drag a feature from the compendium browser/sidebar into the adversary. However you still cannot drag features from the adversary sheet to another adversary or to the sidebar, so its pretty inconsistent right now.

What is the intended or preferred behavior for adversary sheet drag/drop? I can block attempts to drop items onto adversary sheets entirely if that's what we want, or open up the drag handler.

CarlosFdez:
I included adversary/environment feature drag/drop support. That depended on this and it'd probably be better not to retest everything twice.

WBHarry (changes requested):
Everything seems to be working well, except the handling of things being dropped onto the Character sheet.
The `character._onDropItem` never uses the `baseActor._onDropItem`, and therefore always just makes a copy of anything dragged into it rather than potentially doing a transfer.

CarlosFdez:
Alrighty, it should be good now. I also moved handling for the auto vaulting on create to domain card while I was at it.

PR #1359, opened 2025-12-03
[Feature] Prune stats when extracting

CarlosFdez:
Prunes stats (except compendium source for embedded items) when extracting from the database, so that version changes and modified by dates don't register as diffs. These stats are stamped with the correct version numbers during the build.


WBHarry (approved):
:robot: :+1:

PR #1365, opened 2025-12-05
[Fix] Homebrew Item Features

WBHarry:
Fixed so that effects can be used for Homebrew Features, on actions and stand-alone.
Also some other minor fixes

I extracted the majority of `ActionConfig` into `DHActionBaseConfig`.
Now we have `DHActionConfig` and `DHActionSettingsConfig` that extend it.

We could have probably made it work as a single config, but I think it's cleaner this way - and it's actually easy to manage, so I think it's a win. 🥇  
The reason for the split is because Actions typically always belong to an Item, such as a Feature. But for the homebrew setup for  Weapon/Armor Features, Downtime Moves and others, that is not the case as the action is standalone.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -2385,7 +2385,7 @@
                 "downtimeMove": "Downtime Move",
                 "armorFeature": "Armor Feature",
                 "weaponFeature": "Weapon Feature",
-                "newFeature": "New ItemFeature",
+                "newFeature": "New Item Feature",
                 "downtimeMoves": "Downtime Moves",
                 "itemFeatures": "Item Features",
                 "nrChoices": "# Moves Per Rest",
--- a/module/config/itemConfig.mjs
+++ b/module/config/itemConfig.mjs
@@ -459,16 +459,13 @@ export const allArmorFeatures = () => {
         ...armorFeatures,
         ...Object.keys(homebrewFeatures).reduce((acc, key) => {
             const feature = homebrewFeatures[key];
-            const actionEffects = [];
-            const actions = feature.actions.map(action => {
-                const effects = action.effects.map(effect => feature.effects.find(x => x.id === effect._id));
-                actionEffects.push(...effects);
-                return {
-                    ...action,
-                    effects: effects,
-                    type: action.type
-                };
-            });
+            const actions = feature.actions.map(action => ({
+                ...action,
+                effects: action.effects.map(effect => feature.effects.find(x => x.id === effect._id)),
+                type: action.type
+            }));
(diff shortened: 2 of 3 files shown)


lang/en.json
```diff
@@ -2383,7 +2384,7 @@
                 "newDowntimeMove": "Downtime Move",
                 "downtimeMove": "Downtime Move",
                 "armorFeature": "Armor Feature",
-                "weaponFeature": "Weapon Feaure",
+                "weaponFeature": "Weapon Feature",
                 "newFeature": "New ItemFeature",
```
Ikraik:
```suggestion
                "newFeature": "New Item Feature",
```

module/config/itemConfig.mjs
```diff
@@ -1414,11 +1426,24 @@ export const weaponFeatures = {
...
+            const actionEffects = [];
+            const actions = feature.actions.map(action => {
+                const effects = action.effects.map(effect => feature.effects.find(x => x.id === effect._id));
+                actionEffects.push(...effects);
+                return {
+                    ...action,
+                    effects: effects,
+                    type: action.type
+                };
+            });
```
CarlosFdez:
```suggestion
            const actions = feature.actions.map(action => ({
                ...action,
                effects: action.effects.map(effect => feature.effects.find(x => x.id === effect._id)),
                type: action.type
            }));
            const actionEffects = actions.flatMap((a) => a.effects);
```

That said from my testing it looks like effects is blank even if I add an effect to it, because its not in feature.effects, with or without this change. How do I make it have something?

module/config/itemConfig.mjs
```diff
@@ -459,7 +459,19 @@ export const allArmorFeatures = () => {
...
```
CarlosFdez:
```suggestion
            const actions = feature.actions.map(action => ({
                ...action,
                effects: action.effects.map(effect => feature.effects.find(x => x.id === effect._id)),
                type: action.type
            }));
            const actionEffects = actions.flatMap((a) => a.effects);
```

lang/en.json
WBHarry: Corrected it :+1:

module/config/itemConfig.mjs
WBHarry: Changed to the flatmap for actionEffects :+1:

module/config/itemConfig.mjs
WBHarry: Changed to the flatmap for actionEffects :+1:

PR #1374, opened 2025-12-06
[Fix] 1363 Refresh Standardization

### LINKED ISSUE #1363: [Bug] Long rest inconsistently refreshes use tokens
**Describe the bug**
The long rest on the character sheet will refresh uses of things like piercing arrows and dedicated. However, the "long rest" in the "GM Tools" side bar will not. The "short rest" in both "GM Tools" and on the character sheet will refresh these.

**To Reproduce**
1. Go to the character sheet to use some refreshable item or action (eg. piercing arrows)
2. Click on use item
3. Go to the "Daggerheart menu" in the right sidebar
4. Click on "GM Tools" -> "Long rest"
5. Check whether the item has refreshed

**Expected behavior**
Consistent rest refresh behavior.

**Setup Information:**
 - OS: Window
 - Browser Chrome
 - Foundry Version v13 b351

WBHarry:
Closes #1363 
Standardized what gets refreshed by which refreshType using a utils method. It's now used in the DaggerheartMenu and the Downtime menu to determine if a resource/action should get it's resources/uses refreshed.

### DIFF
--- a/module/applications/dialogs/tagTeamDialog.mjs
+++ b/module/applications/dialogs/tagTeamDialog.mjs
@@ -1,3 +1,4 @@
+import { getCritDamageBonus } from '../../helpers/utils.mjs';
 import { GMUpdateEvent, RefreshType, socketEvent } from '../../systemRegistration/socket.mjs';
 
 const { HandlebarsApplicationMixin, ApplicationV2 } = foundry.applications.api;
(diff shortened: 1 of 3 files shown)


CarlosFdez:
CONFLICTS

WBHarry:
https://media.tenor.com/c1P7qk-sMqkAAAAM/kermit.gif

WBHarry:
The world is safe once more. Merge in 😎

CarlosFdez:
The function may lead to some confusion in the future and could probably justify a comment. I could also see refreshIsAllowed("shortRest", { allowed: ["longRest"] }) as a more intuitive API maybe, but that's small beans.

Otherwise LGTM.

PR #1381, opened 2025-12-07
[Fix] Allow disabling coins

CarlosFdez:
This adds the ability to disable coins like the other currency types. Personally, I think coins should default to disabled for new worlds given that tracking it is a variant rule, but I left it as is for compatibility.

### DIFF
--- a/module/applications/sheets/api/base-actor.mjs
+++ b/module/applications/sheets/api/base-actor.mjs
@@ -89,6 +89,7 @@ export default class DHBaseActorSheet extends DHApplicationMixin(ActorSheetV2) {
                     value: context.source.system.gold[key]
                 };
             }
+            context.inventory.hasCurrency = Object.values(context.inventory.currencies).some((c) => c.enabled);
         }
 
         return context;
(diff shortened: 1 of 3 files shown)


CarlosFdez:
In general the blocking is a bit suspicious to me. I suspect its because there used to be some other way to enable/disable coins. Please correct me if I'm wrong about this.

WBHarry:
There used to be, yeah.

The reason I changed it was beacause I added the ability to disable each denomination. It didn't make sense to me to disable -every- denomination, so I made the top one not able to be disabled.

CarlosFdez:
I mean, if a game doesn't use gold at all, disabling every single one makes perfect sense to me, though maybe I should test for that scenario soon.

Coins is the one the rules disable by default.

WBHarry:
Yah, I'm aware coins are technically an optional rule. It's just quite akward since it's logically the choice that's first when listing them, which makes it strange UI:wise

WBHarry:
But maybe it is a thing that you could run a game without any coinage at all. We could just allow disabling all of them. Would just have to verify that there's no logic expecting there to be some demonination present (I doubt it)

CarlosFdez:
omitted from the sheets at least, anything else I think maybe best to wait for a bug report. I'd surprised if anything breaks though.

PR #1382, opened 2025-12-07
[PR] [Feature] Refresh actors when a homebrew setting is changed

CarlosFdez:
Small QoL to trigger trigger a reset and re-render (non-forced, so open sheets only) for all actors whenever a homebrew setting is updated. This means that changing coins will immediately show the change in open actor sheets.

### DIFF
--- a/module/systemRegistration/settings.mjs
+++ b/module/systemRegistration/settings.mjs
@@ -134,17 +134,23 @@ const registerNonConfigSettings = () => {
     });
 };
 
-/** 
+/**
  * Triggers a reset and non-forced re-render on all given actors (if given)
  * or all world actors and actors in all scenes to show immediate results for a changed setting.
  */
 function resetActors(actors) {
     actors ??= [
         game.actors.contents,
-        game.scenes.contents.flatMap((s) => s.tokens.contents).flatMap((t) => t.actor ?? []),
+        game.scenes.contents.flatMap(s => s.tokens.contents).flatMap(t => t.actor ?? [])
     ].flat();
     actors = new Set(actors);
     for (const actor of actors) {
+        for (const app of Object.values(actor.apps)) {
+            for (const element of app.element?.querySelectorAll('prose-mirror.active')) {
+                element.open = false; // This triggers a save
+            }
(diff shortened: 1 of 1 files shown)


module/systemRegistration/settings.mjs
```diff
@@ -130,3 +133,19 @@ const registerNonConfigSettings = () => {
...
+ * or all world actors and actors in all scenes to show immediate results for a changed setting.
+ */
+function resetActors(actors) {
+    actors ??= [
+        game.actors.contents,
+        game.scenes.contents.flatMap((s) => s.tokens.contents).flatMap((t) => t.actor ?? []),
+    ].flat();
+    actors = new Set(actors);
+    for (const actor of actors) {
+        actor.reset();
```
WBHarry, line 154: Hm, does reinitializing the actor potentially wipe out some changes that were currently underway on the character?
IE - player was writing their bio or something like that. 🤔

CarlosFdez: Oh definitely. Maybe good to limit the settings affected.

Ikraik: @CarlosFdez is this resolved?

CarlosFdez: No. Actually I'm remembering now that its hard to get what props have actually changed, so it requires storing a global variable and honestly its not worth narrowing. Changes to the homebrew object are rarely ever done during play anyways, and every single prop that I can see is worth refreshing for.

Ikraik: @CarlosFdez shall I close it for now or planning to leave it draft?

CarlosFdez: Sorry, I mean that I don't think it needs resolving. Correct me if I'm wrong though.

Ikraik:
BTW @CarlosFdez shall I make this into a draft PR?

CarlosFdez:
No. It was complete, it just needed to decide what compromise we wanted, and that required feedback. We got that feedback and I decided on saving any open prosemirror windows.

PR #1401, opened 2025-12-09
[PR] [Feature] Refocus consumable quantity and simple resource fields on change

CarlosFdez:
Assigning an id allows the core handlebars helper to refocus it on re-render. However upstream doesn't maintain cursor position on refocus, so this only ends up improving the use of arrow keys the increment/decrement the numbers.

### DIFF
--- a/templates/sheets/global/partials/inventory-item-V2.hbs
+++ b/templates/sheets/global/partials/inventory-item-V2.hbs
@@ -61,7 +61,7 @@ Parameters:
     {{/if}}
     {{#if (and (not hideResources) (gte item.system.quantity 0))}}
     <div class="item-resource">
-      <input type="number" id="{{@root.rootId}}-Item-{{item.id}}-Quantity" class="inventory-item-quantity" value="{{item.system.quantity}}" min="0" />
+      <input type="number" id="{{item.uuid}}-quantity" class="inventory-item-quantity" value="{{item.system.quantity}}" min="0" />
     </div>
     {{/if}}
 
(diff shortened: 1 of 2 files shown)


CarlosFdez:
I'm considering replacing with `{{item.uuid}}-Quantity` and such. Thoughts?

the name and id props are what HandlebarsApplicationMixin#_preSyncPartState() uses to refocus an element after a re-render. You'll see that while it works fine for currencies right now, if you use the arrow keys on the consumable quantities it unfocuses immediately. I'm thinking this might actually be a (very minor) bugfix, rather than a feature, actually.

WBHarry:
Ah, right. Refocusing when you hit `enter`. I never do that, so I didn't get what the functional change was at first. I always just enter something and click outside. 😆

WBHarry:
> I'm considering replacing with `{{item.uuid}}-Quantity` and such. Thoughts?
> 
> the name and id props are what HandlebarsApplicationMixin#_preSyncPartState() uses to refocus an element after a re-render. You'll see that while it works fine for currencies right now, if you use the arrow keys on the consumable quantities it unfocuses immediately. I'm thinking this might actually be a (very minor) bugfix, rather than a feature, actually.

It sounds sensible to me. Especially since you set the id to be `{root}.item.{item.id}` regardless, which is essentially recreating the uuid.

Curiously, I tested this, and using `{{item.uuid}}-Quantity` makes the arrow-key useage work without losing focus. Unsure why really, as I feel the previous one should have also been stable. 🤔 
Seems like two birds with one stone though.

CarlosFdez:
Wait, it doesn't work without using the item uuid? How strange. I was mostly considering simplicity.

WBHarry:
What I noticed was that writing a number and hitting enter will keep the focus with the code that's currently in. But using the up/down arrows loses focus.

If I changed it to the `uuid` solution, then both worked.

CarlosFdez:
I double checked and the focus is kept with the current solution. I'm swapping to uuid anyways though, since its simpler and I don't think we're ever gonna multiple of the same sheet open.

Ikraik:
Hmm the refocus with multiple sheet happened when I was messing with names XD. Going with UUID is a good approach and unique enough I feel. 

(Edit: I think I used document.id to resolve it for actor sheets)

PR #1414, opened 2025-12-11
[Fix] EffectHud standardized

WBHarry:
Made the effect-hud more foundry standard to work better for modules.
It changes nothing for the system standalone as it's equivalent.

### DIFF
--- a/styles/less/hud/token-hud/token-hud.less
+++ b/styles/less/hud/token-hud/token-hud.less
@@ -56,7 +56,7 @@
             .effect-locked {
                 position: absolute;
                 bottom: 2px;
-                right: 2px;
+                left: 11.5px;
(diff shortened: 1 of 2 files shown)


styles/less/hud/token-hud/token-hud.less
```diff
@@ -56,7 +56,7 @@
             .effect-locked {
                 position: absolute;
                 bottom: 2px;
-                right: 2px;
```
CarlosFdez, line 59: does this work with interface scaling?

Ikraik (approved):
LGTM 👍

WBHarry:
Hm. I'm suprised, but it does actually work.
1.0
image

1.5
image

PR #1417, opened 2025-12-12
[Community PR] Pulling out hope method and always updating the resources.

### LINKED ISSUE #1130: [Bug] /dr command rolls no longer automated
**Describe the bug**
Rolls made with the /dr command no longer add fear/hope 

**To Reproduce**
1. Go to chat
2. Type /dr & press enter
3. Complete Roll from popup
4. Fear/Hope (Stress for crits) are not changing, even with automation enabled

**Expected behavior**
Fear/Hope/Stress should be updated with automation enabled. (Alternatively make this an additional automation setting for /dr command specifically)

nsalyzyn:
## Description

Updating the dice automation to actually work again. I did some bisecting of the repo and it hasn't worked for some time. I pulled out the hope/fear automation into it's own private method, which makes this a bit harder to review. However, that was necessary to fix another obvious bug - the code below the hope/fear automation would not be run if hope/fear automation was off. Also, while testing this, I discovered that stress was _increased_ instead of decreased on a critical, so I fixed that too.

- Fixes #1130 

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Code cleanup/refactor
- [ ] Documentation update
- [ ] Test coverage
- [ ] Dependency update
- [ ] Configuration change

## How Has This Been Tested?

I rolled a bunch of times with automation on and off as the GM.

- [X] Manual testing
- [ ] Other:

## Checklist

- [X] I have made corresponding changes to the documentation

## Additional Comments

If you want this as 3 different PRs to fix the 3 things I changed, let me know. Also, let me know if you prefer a different layout for private methods, since I couldn't find any great examples of private function use or naming convention in the code base.

### DIFF
--- a/module/dice/dhRoll.mjs
+++ b/module/dice/dhRoll.mjs
@@ -237,6 +237,51 @@ export default class DHRoll extends Roll {
     }
 }
 
+async function automateHopeFear(config) {
+    const automationSettings = game.settings.get(CONFIG.DH.id, CONFIG.DH.SETTINGS.gameSettings.Automation);
+    const hopeFearAutomation = automationSettings.hopeFear;
+    if (!config.source?.actor ||
+        (game.user.isGM ? !hopeFearAutomation.gm : !hopeFearAutomation.players) ||
+        config.actionType === 'reaction' ||
+        config.tagTeamSelected ||
+        config.skips?.resources)
+        return;
+    const actor = await fromUuid(config.source.actor);
+    let updates = [];
+    if (!actor) return;
+
+    if (config.rerolledRoll) {
+        if (config.roll.result.duality != config.rerolledRoll.result.duality) {
+            const hope = (config.roll.isCritical || config.roll.result.duality === 1 ? 1 : 0)
+                - (config.rerolledRoll.isCritical || config.rerolledRoll.result.duality === 1 ? 1 : 0);
+            const stress = (config.roll.isCritical ? 1 : 0) - (config.rerolledRoll.isCritical ? 1 : 0);
+            const fear = (config.roll.result.duality === -1 ? 1 : 0)
+                - (config.rerolledRoll.result.duality === -1 ? 1 : 0)
+
+            if (hope !== 0)
+                updates.push({ key: 'hope', value: hope, total: -1 * hope, enabled: true });
+            if (stress !== 0)
+                updates.push({ key: 'stress', value: -1 * stress, total: stress, enabled: true });
+            if (fear !== 0)
+                updates.push({ key: 'fear', value: fear, total: -1 * fear, enabled: true });
+        }
+    } else {
+        if (config.roll.isCritical || config.roll.result.duality === 1)
+            updates.push({ key: 'hope', value: 1, total: -1, enabled: true });
+        if (config.roll.isCritical)
+            updates.push({ key: 'stress', value: -1, total: 1, enabled: true });
+        if (config.roll.result.duality === -1)
+            updates.push({ key: 'fear', value: 1, total: -1, enabled: true });
+    }
+
+    if (updates.length) {
+        const target = actor.system.partner ?? actor;
+        if (!['dead', 'defeated', 'unconscious'].some(x => actor.statuses.has(x))) {
+            await target.modifyResource(updates);
+        }
+    }
+}
+
 export const registerRollDiceHooks = () => {
     Hooks.on(`${CONFIG.DH.id}.postRollDuality`, async (config, message) => {
         const automationSettings = game.settings.get(CONFIG.DH.id, CONFIG.DH.SETTINGS.gameSettings.Automation);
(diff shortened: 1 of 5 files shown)


module/dice/dhRoll.mjs
```diff
@@ -237,6 +237,40 @@ export default class DHRoll extends Roll {
...
+        config.actionType === 'reaction' ||
+        config.tagTeamSelected ||
+        config.skips?.resources)
+        return;
+    const actor = await fromUuid(config.source.actor);
+    let updates = [];
+    if (!actor) return;
+    if (config.roll.isCritical || config.roll.result.duality === 1)
+        updates.push({ key: 'hope', value: 1, total: -1, enabled: true });
+    if (config.roll.isCritical) updates.push({ key: 'stress', value: -1, total: 1, enabled: true });
```
WBHarry: This is incorrect. Stress is an `inverted` resource, so it's supposed to be value=1, total=-1.
This change makes stress increase on critical :tongue:

module/dice/dhRoll.mjs
```diff
@@ -237,6 +237,40 @@ export default class DHRoll extends Roll {
...
+    if (config.roll.result.duality === -1) updates.push({ key: 'fear', value: 1, total: -1, enabled: true });
+
+    if (config.rerolledRoll) {
+        if (config.rerolledRoll.isCritical || config.rerolledRoll.result.duality === 1)
+            updates.push({ key: 'hope', value: -1, total: 1, enabled: true });
+        if (config.rerolledRoll.isCritical) updates.push({ key: 'stress', value: 1, total: -1, enabled: true });
```
WBHarry: This is incorrect. Stress is an `inverted` resource, so it's supposed to be value=-1, total=1 for the reroll.

WBHarry:
Ah, there's actually something funky here. For 'normal' rolls, the previous logic produces a correct result. 
But for `/dr` specifically, the invertion you did produces a correct result. There's something wrong in the later stages for `/dr` then. Having a look

WBHarry:
Alright. We can keep the changed logic you introduced here. There are just two lines that have to go away to make it work for Trait Rolls and Group Rolls.

`sheets/actors/character.mjs`
Line 689 (`game.system.api.fields.ActionFields.CostField.execute`) has to go away. The cost is already handled.

Same for:
`chatLog.mjs` 
Line 248 (`game.system.api.fields.ActionFields.CostField.execute`)

Tested that the PR does fix using the `/dr` command with a selected character, so good :+1:

nsalyzyn:
Actually, this code still doesn't work when I run the game as a player, so I'll need to keep digging.

EDIT: Oh, I just didn't have a Gamemaster logged in. The resources don't update if the gamemaster isn't present.

module/dice/dhRoll.mjs
nsalyzyn: As you noted, we can keep this logic - even though it shouldn't be working the way it is.

WBHarry:
Ah, yeah. Since we're just running `emitAsGM` at the end to do the update, it will always defer to the GM to do the update.
This should be changed, as there is no need to do that if the player already has ownership of the entity to update.

WBHarry:
Seems like a good way to do it would be to add a check in `emitAsGM` to check if the user has permissions. Flipping the logic something like:
```
export const emitAsGM = async (eventName, callback, update, uuid = null, refresh = null) => {
    if (game.user.isGM || entity.testUserPermission(game.user, 'OWNER')) {
        return callback(update);
    } else {
        return await game.socket.emit(`system.${CONFIG.DH.id}`, {
            action: socketEvent.GMUpdate,
            data: {
                action: eventName,
                uuid,
                update,
                refresh
            }
        });
    }
};
```

You don't have to do that add in this PR if you don't want to though, as it would probably entail changing the call signature of the function to get `entity`, and changing the places where it is called to accomodate.

WBHarry (dismissed):
Going to call this finished as the original issue is resolved.
Changing so that users can do the update without a GM is a separate issue :+1:

nsalyzyn:
I'm going to call that out of scope. It doesn't look like the player _has_ permission to do things like edit fear anyway, so your suggestion would only work for their hope and stress. 

Of note, I figured out how to reroll dice in the UI! My new code has a regression here. Back to the drawing board. Sorry about that. Still learning your UI _and_ code base it seems.

nsalyzyn:
I had to make multiple more changes to make rerolls work. The code should make a bit more sense now.

This fixes a pre-existing bug with critical hits subtracting fear on re-roll.

There still exists corner cases where you have full hope, "gain hope" on a roll, then reroll it to gain fear. Similar with zero stress and re-rolling a critical. Both of these should rarely happen, since these are rolls you're least likely to reroll. I'm not sure how the code as written could easily address these corner cases.

Tested by using d4 rolls on both player and GM and re-rolling them many times to see if the behavior is consistent.

WBHarry:
It's strange that there's no commit history. It makes it a lot harder to verify what changes are being done gradually. 🤔

nsalyzyn:
Sorry, I've been using force push and amend to keep it as one commit. This is my first time contributing to a project via Github, so I don't know what is polite. Nor was I sure how it would show up. I'll do multiple commits in the future for my PRs. (My past experience is 10+ years in FAANG, so I'm not a total noob :) - although very little of that was in Javascript).

I do see my force-pushed actions highlighted in this thread, if you need to see a breakdown.

WBHarry:
It's not a massive PR by any stretch, so it's not so bad. But yes, it is a lot nicer to keep a commit history so it's easier to see what exactly was done for the `pre-existing bug with critical` 😉 

We do squash commits for merging, so it'll just be one commit in main regardless.

nsalyzyn:
Ok. Good to know it'll be squashed later. I wasn't sure.

Fixing the pre-existing bug with critical is the bit where I say `const hope = ...` and go from there. Otherwise, you'd do two updates to hope back-to-back when you call modifyResources, which doesn't work.

Edit: IMO, that improves readability on what actually is going into the updates, which is why I added the comment on modifyResources.

CarlosFdez:
> Sorry, I've been using force push and amend to keep it as one commit. This is my first time contributing to a project via Github, so I don't know what is polite. Nor was I sure how it would show up. I'll do multiple commits in the future for my PRs. (My past experience is 10+ years in FAANG, so I'm not a total noob :) - although very little of that was in Javascript).
> 
> I do see my force-pushed actions highlighted in this thread, if you need to see a breakdown.

It's very project dependent. In this project people have a preference for separate prs and since they like local pulling stuff, it helps not to break history.

I've worked on projects where we do prefer force push, and force pushing is my default.

PR #1418, opened 2025-12-12
[Feature] Make the existing spotlight queue toggleable (default off)

CarlosFdez:
This makes the spotlight ordering feature an opt-in feature rather than the default. The setting exists at the top level, but let me know if there is a better place for this. It doesn't seem like an automation setting and barely even an appearance settings, but my rubric might not be fine tuned.

I could also add a Enable/Disable context menu option to the encounter tracker itself, instead or in addition to the visible settings. Let me know what you think about that.

image

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -2504,6 +2504,11 @@
                     "hint": "Apply variant rules from the Daggerheart system",
                     "name": "Variant Rules",
                     "actionTokens": "Action Tokens"
+                },
+                "SpotlightRequestQueue": {
+                    "name": "Spotlight Request Queue",
+                    "label": "Spotlight Request Queue",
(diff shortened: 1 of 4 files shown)


Ikraik:
@CarlosFdez it didn't fit in with any other headers? Feels like it comes under either Homebrew or Automation as it is simplifying an exisiting implement. But that's just personal opinion here.

CarlosFdez:
Can't be homebrew, these aren't campaign specific specific settings or a houserule. Its not automation either, since automation is about the software automatically handling certain rules (such as auto granting hope/fear or auto ticking an action token). At best its appearance, but its a loose fit.

I'm in agreement in Harry that it doesn't fit any existing category. If we move it elsewhere, it may be as a button in a context menu in the encounter tracker (but then we lose the hint, but maybe that's fine).

PR #1422, opened 2025-12-13
[Feature] Support for configurable currency icons

CarlosFdez:
Adds icons for each currency and a way to override them. Its a little simple looking in the inventory, but maybe someone can give it a bit more flair so it looks a little more like an item.

Note that until the refresh on homebrew changes PR is merged in, changes are not reflected in the character sheet until closed and reopened.

https://github.com/user-attachments/assets/831032b7-ddeb-4718-aeea-19b45096f249

### DIFF
--- a/module/data/settings/Homebrew.mjs
+++ b/module/data/settings/Homebrew.mjs
@@ -9,7 +9,7 @@ const currencyField = (initial, label, icon) =>
             initial,
             label
         }),
-        icon: new foundry.data.fields.StringField({ required: true, nullable: false, blank: false, initial: icon }),
+        icon: new foundry.data.fields.StringField({ required: true, nullable: false, blank: true, initial: icon })
     });
 
 export default class DhHomebrew extends foundry.abstract.DataModel {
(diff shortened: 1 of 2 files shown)


module/data/settings/Homebrew.mjs
```diff
@@ -1,14 +1,15 @@
...
+const currencyField = (initial, label, icon) =>
     new foundry.data.fields.SchemaField({
         enabled: new foundry.data.fields.BooleanField({ required: true, initial: true }),
         label: new foundry.data.fields.StringField({
             required: true,
             initial,
             label
-        })
+        }),
+        icon: new foundry.data.fields.StringField({ required: true, nullable: false, blank: false, initial: icon }),
```
WBHarry: The icon should be nullable. Currently you can't empty the icon field and save it (because you don't want to display an icon for your homebrew currency for instance). You have to enter a bogus string, which isn't ideal.

WBHarry: I set the icon to be `blank: true` to allow an empty icon. :+1:

PR #1431, opened 2025-12-16
[PR] [Feature] Support drag dropping currencies to actor sheets

CarlosFdez:
Adds support for currency drag/drop and reworks the styling of the item transfer dialog. Unfortunately the drag handle is a bit hidden and non-obvious. Please give it a try though.

image

### DIFF
--- a/styles/less/dialog/item-transfer/sheet.less
+++ b/styles/less/dialog/item-transfer/sheet.less
@@ -3,12 +3,12 @@
         display: flex;
         align-items: center;
         justify-content: center;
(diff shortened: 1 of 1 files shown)


Ikraik:
Just to confirm this is just to transfer items or just currency?

CarlosFdez:
Regular item transferring is already supported, and I'm adding the ability to do the same for currency. However the visual dialog update is for both. I had to generalize the dialog some more to support both and I had the opportunity to update its styling (thanks molilo).

PR #1429, opened 2025-12-15
[PR] [Feature] Feature form labels + SRD Update (Adversaries & Environments)

Ikraik:
## Description

Feature Form label added in features and only applies to Adversary and Environment. Added name field changes to this PR which now handles text wrapping under a label.

- Fixes #1027 
- Closes #1027

## Type of Change

Please check the relevant options:

- [ ] Bug fix
- [x] New feature
- [ ] Code cleanup/refactor
- [x] Documentation update
- [ ] Test coverage
- [ ] Dependency update
- [ ] Configuration change
- [ ] Other (please describe):

## How Has This Been Tested?

Please describe the tests you ran to verify your changes:

- [x] Manual testing
- [ ] Other:

## Screenshots (if applicable)
The Settings:
image

Updated Adversary:
image

Updated Environment:
image

Text wrapped Name Field on character:

https://github.com/user-attachments/assets/d54ad9b7-b2be-4983-a256-1f29fcd7618c

## Checklist

- [x] My code follows the project style guidelines
- [x] I have performed a self-review of my code
- [x] I have commented my code where necessary
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings or errors
- [ ] I have added tests that prove my fix or feature works
- [ ] New and existing tests pass locally with my changes

## Additional Comments

By default, all features (for adversary and environment) will have initial value of "passive". This is to support custom adversary changes that are not there in SRD. Issue to note: Sorting is now possible within adversary and environment using these labels but it is to be handled in a separate PR. Name field is a stop gap solution to handle input text wrap. 

---

> Thank you for your contribution! 🎉

PR #1433, opened 2025-12-17
[PR][Feature] Actor Sizes

WBHarry:
- The system is implemented.
- Compendium adversaries updated.

### DIFF
--- a/module/data/item/beastform.mjs
+++ b/module/data/item/beastform.mjs
@@ -196,14 +196,14 @@ export default class DHBeastform extends BaseDataItem {
 
         await this.parent.parent.createEmbeddedDocuments('ActiveEffect', [beastformEffect.toObject()]);
 
-        let width = this.tokenSize.width;
-        let height = this.tokenSize.height;
-        if (this.tokenSize.size !== 'custom') {
-            const tokenSizes = game.settings.get(CONFIG.DH.id, CONFIG.DH.SETTINGS.gameSettings.Homebrew).tokenSizes;
-            const tokenSize = tokenSizes[this.tokenSize.size];
-            width = tokenSize;
-            height = tokenSize;
-        }
+        const autoTokenSize =
+            this.tokenSize.size !== 'custom'
+                ? game.settings.get(CONFIG.DH.id, CONFIG.DH.SETTINGS.gameSettings.Homebrew).tokenSizes[
+                      this.tokenSize.size
+                  ]
+                : null;
+        const width = autoTokenSize ?? this.tokenSize.width;
+        const height = autoTokenSize ?? this.tokenSize.height;
 
         const prototypeTokenUpdate = {
             height,


Ikraik:
Do note that #1429 has compendium changes so if you are planning for one I suggest doing it after that PR is merged.

module/applications/settings/homebrewSettings.mjs
```diff
@@ -424,6 +425,17 @@ export default class DhHomebrewSettings extends HandlebarsApplicationMixin(Appli
         this.close();
     }
 
+    static async resetTokenSizes() {
+        await this.settings.updateSource({
+            tokenSizes: Object.values(this.settings.schema.fields.tokenSizes.fields).reduce(
```
CarlosFdez: Does `this.settings.schema.fields.tokenSizes.getInitialValue()` work? Seemed to when I had to do a similar thing for currency.

WBHarry: Hm. Possible it would. I assumed it wouldn't because I didn't explicitly set `initial` for the schemaField. But would be good to try.

WBHarry: Worked :+1:

module/data/item/beastform.mjs
```diff
@@ -190,9 +196,18 @@ export default class DHBeastform extends BaseDataItem {
...
         await this.parent.parent.createEmbeddedDocuments('ActiveEffect', [beastformEffect.toObject()]);
 
+        let width = this.tokenSize.width;
+        let height = this.tokenSize.height;
+        if (this.tokenSize.size !== 'custom') {
+            const tokenSizes = game.settings.get(CONFIG.DH.id, CONFIG.DH.SETTINGS.gameSettings.Homebrew).tokenSizes;
+            const tokenSize = tokenSizes[this.tokenSize.size];
+            width = tokenSize;
+            height = tokenSize;
+        }
```
CarlosFdez:
```suggestion
        const autoTokenSize = this.tokenSize.size !== 'custom' ? game.settings.get(CONFIG.DH.id, CONFIG.DH.SETTINGS.gameSettings.Homebrew).tokenSizes[this.tokenSize.size] : null;
        const width = autoTokenSize ?? this.tokenSize.width;
        const height = autoTokenSize ?? this.tokenSize.height;
```

Maybe try something like that? Formatting the top line of course.

WBHarry: Neater, yup. Changed it :+1:

CarlosFdez: Ended up being about the same number of lines but at least the intent is clearer.

CarlosFdez:
I took another glance at the code, does this work if dragging from a compendium instead?

WBHarry:
Yup! Can try it in main.

PR #1435, opened 2025-12-18
[Community PR] Preventing getRollData from returning infinite recursion on the system property

nsalyzyn:
## Description

Adds system explicitly to getRollData's output so that it doesn't infinitely recurse.

- Fixes #867

## Type of Change

Please check the relevant options:

- [X] Bug fix
- [ ] New feature
- [ ] Code cleanup/refactor
- [ ] Documentation update
- [ ] Test coverage
- [ ] Dependency update
- [ ] Configuration change
- [ ] Other (please describe):

## How Has This Been Tested?

* I tested in the console whether items and actors have infinite recursion.
* I checked a few different ways of doing rolls - attacking, clicking on traits and /dr.
* I made sure the Seraph prayer dice behaved correctly (this one didn't work unless there was at least 1 layer of system depth for some reason)

- [X] Manual testing
- [ ] Other:

## Checklist

- [X] My code follows the project style guidelines
- [X] I have performed a self-review of my code
- [ ] I have commented my code where necessary
- [ ] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings or errors
- [ ] I have added tests that prove my fix or feature works
- [X] New and existing tests pass locally with my changes

## Additional Comments

I had a bit of trouble verifying that the `item/base.mjs` change is necessary. Triggering it in the console was tricky. I think I did it, but if you have suggestions of more things to test, that would be helpful.

PR #1436, opened 2025-12-18
[PR] Attempt #2 preventing getRollData from returning infinite recursion on the system property

### LINKED ISSUE #867: [BUG] - getRollData returns object with infinitely nested .system property
**Describe the bug**
When calling getRollData(), the returned object contains a .system property that itself contains another .system property, which contains another .system, and so on recursively without end.

**To Reproduce**
1. Open the console in Foundry.
2. Run `game.actors.contents[0].getRollData()`
3. Expand the .system property in the console.
4. Observe that .system.system.system... repeats indefinitely.

**Expected behavior**
getRollData() should return a finite object, with .system containing actual system data only once, not recursively.

**Notes**
Apparently this happens with Actors and Items.

nsalyzyn:
## Description

Prevents the DhpActor class from munging its system with a self-reference every time getRollData() is called. Uses the clone() method.

- Fixes #867

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Code cleanup/refactor
- [ ] Documentation update
- [ ] Test coverage
- [ ] Dependency update
- [ ] Configuration change

## How Has This Been Tested?

* I tested in the console whether items and actors have infinite recursion.
* I checked a few different ways of doing rolls - attacking, clicking on traits and /dr. Attacking with items required clone() rather than just a plain old Object.
* I made sure the Seraph prayer dice behaved correctly (this one required the system property to be set to begin with)

- [X] Manual testing
- [ ] Other:

## Checklist

## Additional Comments

See #1435 for a prior attempt to fix this same issue.

### DIFF
--- a/module/documents/actor.mjs
+++ b/module/documents/actor.mjs
@@ -526,7 +526,7 @@ export default class DhpActor extends Actor {
 
     /**@inheritdoc */
     getRollData() {
-        const rollData = { ...super.getRollData() };
+        const rollData = super.getRollData().clone();
         rollData.name = this.name;
         rollData.system = this.system.getRollData();
         rollData.prof = this.system.proficiency ?? 1;


CarlosFdez:
The infinite recursion doesn't seem a big deal, however the side effects are. This will break anything that relies on the those side effects, but they really shouldn't have existed in the first place.

CarlosFdez (approved):

CarlosFdez:
I don't like this solution but I don't think better exists, at least not without major breaking changes.

PR #1437, opened 2025-12-18
[PR][Feature] Support multiline adversary and character names

CarlosFdez:
Made the adversary name an editable header element that saves on blur or when pressing the enter key.

image

image

### DIFF
--- a/styles/less/sheets/actors/adversary/header.less
+++ b/styles/less/sheets/actors/adversary/header.less
@@ -24,6 +24,7 @@
                 border: 1px solid transparent;
                 outline: 2px solid transparent;
                 transition: all 0.3s ease;
+                word-break: break-word;
 
                 &:hover {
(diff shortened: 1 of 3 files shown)


Ikraik:
So apart from adversaries, characters also are affected but it looks like this will solve the issue well since it is global and doesn't affect styling too much as well. I'll tag molilo for review in terms of styling expertise but apart from it, it seems fine to me.

CarlosFdez:
Characters are actually unaffected. Did we want to rope them in too? Characters have a lot more room so it didn't seem necessary.

Ikraik:
Yeah. Because the field size is too small for moderately sized name. Less lenient than adversary in that regard. Might as well add support for Character actor as well.

moliloo (approved):
Its good to merge, nice work!

PR #1439, opened 2025-12-18
[PR] [Fix] 1432 - Battlepoints Effect Application

### LINKED ISSUE #1432: [Bug] Players receiving bonus damage from BP 1d4 damage bonus effect
The Players can receive the effect to do 1d4 bonus damage from the BP option to increase difficulty under certain circumstances.

To reproduce:
1. Create an encounter with adversaries but no players.
2. Add the +1d4 damage option
3. Add players to the encounter
4. Players will receive the buff

**Expected behavior**
Adversaries should receive the effect but not players.

**Setup Information:**
 - OS: Windows
 - Browser: Chrome
 - Foundry Version v13 b351
- System Version v1.3.2

WBHarry:
Fixes #1432 
Fixed so effects are only applied to correct tokens.

### DIFF
--- a/module/applications/ui/combatTracker.mjs
+++ b/module/applications/ui/combatTracker.mjs
@@ -5,8 +5,7 @@ export default class DhCombatTracker extends foundry.applications.sidebar.tabs.C
         actions: {
             requestSpotlight: this.requestSpotlight,
             toggleSpotlight: this.toggleSpotlight,
-            setActionTokens: this.setActionTokens,
-            openCountdowns: this.openCountdowns
+            setActionTokens: this.setActionTokens
         }
     };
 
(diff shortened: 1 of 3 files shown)


module/applications/ui/combatTracker.mjs
```diff
@@ -5,8 +5,7 @@ export default class DhCombatTracker extends foundry.applications.sidebar.tabs.C
         actions: {
             requestSpotlight: this.requestSpotlight,
             toggleSpotlight: this.toggleSpotlight,
-            setActionTokens: this.setActionTokens,
-            openCountdowns: this.openCountdowns
```
Ikraik, line 9: Is there a reason why openCountdowns was removed?

WBHarry: Function doesn't exist. It was just an empty declaration left behind

Ikraik (approved):
To note: +1d4 adds twice if both magical and physical damage exists.

CarlosFdez:
This looks fine to me. The 1d4 adding twice thing is a problem but unrelated to what you're fixing at least.

PR #1443, opened 2025-12-19
[PR] Changing the drag targets in the inventory item v2 template

### LINKED ISSUE #990: [BUG] - Drag to select resources doesn't work on sheet
**Describe the bug**
Dragging to select the resource amount drags the entire button

**To Reproduce**
1. Have a feature with a resource
2. Try to drag on top of the number to select and delete it
3. the entire button gets pulled and you cant select it

**Expected behavior**
Should be able to pull to select resource text

**Screenshots**

Image

**Setup Information:**
 - OS: Windows
 - Browser Chrome
 - Foundry Version 347
- System Version 1.0.5

**Additional context**

nsalyzyn:
## Description

In order to fix the issue, this PR moves the draggable locations inside of `class="inventory-item"` to some children elements. There are still other templates that use `class="inventory-item" draggable="true"`, so the dragDrop selections for sheets are left as being versatile. Also, there is a visual cleanup where it grabs an image for the dragging, so that you don't get inconsistent behavior when you drag the image vs the text.

- Fixes #990 

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Code cleanup/refactor
- [ ] Documentation update
- [ ] Test coverage
- [ ] Dependency update
- [ ] Configuration change

## How Has This Been Tested?

I tested the following drag-drop behaviors:

* Dragging features between different adversaries in the top level sheet
* On the character sheet, taking items from the left side and moving them to the party sheet
* On the character sheet, taking items from the inventory area and rearranging them
* On the character sheet, taking items from the inventory area and moving them to the party sheet
* On the party sheet, moving the items back to a character
* On an environment, moving adversaries from one environment to another

- [X] Manual testing
- [ ] Other:

## Checklist

### DIFF
--- a/module/applications/sheets/actors/character.mjs
+++ b/module/applications/sheets/actors/character.mjs
@@ -46,7 +46,7 @@ export default class CharacterSheet extends DHBaseActorSheet {
         },
         dragDrop: [
             {
-                dragSelector: '[data-item-id]',
+                dragSelector: '[data-item-id][draggable="true"], [data-item-id] [draggable="true"]',
                 dropSelector: null
             }
         ],
@@ -866,6 +866,15 @@ export default class CharacterSheet extends DHBaseActorSheet {
         new game.system.api.applications.dialogs.Downtime(this.document, button.dataset.type === 'shortRest').render({
             force: true
         });
+    }
+
+    /** @inheritdoc */
+    async _onDragStart(event) {
+        const inventoryItem = event.currentTarget.closest('.inventory-item');
+        if (inventoryItem) {
+            event.dataTransfer.setDragImage(inventoryItem.querySelector('img'), 60, 0);
+        }
+        super._onDragStart(event);
     }
 
     async _onDropItem(event, item) {


module/applications/sheets/actors/character.mjs
```diff
@@ -45,10 +45,15 @@ export default class CharacterSheet extends DHBaseActorSheet {
...
         dragDrop: [
+            {
+                dragSelector: '[data-item-id] [draggable="true"]',
+                dropSelector: null
+            },
             {
                 dragSelector: '[data-item-id][draggable="true"]',
                 dropSelector: null
             }
+
```
CarlosFdez:
```suggestion
```

module/applications/sheets/api/application-mixin.mjs
```diff
@@ -363,8 +366,8 @@ export default function DHApplicationMixin(Base) {
                         fromInternal: this.document.uuid
                     };
                     event.dataTransfer.setData('text/plain', JSON.stringify(effectData));
-                    event.dataTransfer.setDragImage(inventoryItem.querySelector('img'), 60, 0);
                 }
+                event.dataTransfer.setDragImage(inventoryItem.querySelector('img'), 60, 0);
```
CarlosFdez: I would actually prefer if this isn't moved, since this may cause issues with the currency stuff.

module/applications/sheets/actors/character.mjs
CarlosFdez: Though TBH given it overrides, I wonder why this is needed? Wouldn't it inherit the drag/drop selectors?

templates/sheets/global/partials/inventory-item-V2.hbs
```diff
@@ -18,24 +18,24 @@ Parameters:
...
 <li class="inventory-item" data-item-id="{{item.id}}" {{#if (or (eq type 'action' ) (eq type 'attack' ))}}
-  data-action-id="{{item.id}}" {{/if}} data-item-uuid="{{item.uuid}}" data-type="{{type}}" data-no-compendium-edit="{{noCompendiumEdit}}" draggable="true">
+  data-action-id="{{item.id}}" {{/if}} data-item-uuid="{{item.uuid}}" data-type="{{type}}" data-no-compendium-edit="{{noCompendiumEdit}}">
   <div class="inventory-item-header {{#if hideContextMenu}}padded{{/if}}" {{#unless noExtensible}}data-action="toggleExtended" {{/unless}}>
     {{!-- Image --}}
     <div class="img-portait" data-action='{{ifThen (or (hasProperty item "use") (eq type "attack")) "useItem" (ifThen
       (hasProperty item "toChat" ) "toChat" "editDoc" ) }}' {{#unless hideTooltip}} {{#if (eq type 'attack' )}}
       data-tooltip="#attack#{{item.actor.uuid}}" {{else}} data-tooltip="#item#{{item.uuid}}" {{/if}} {{/unless}}>
-      <img src="{{item.img}}" class="item-img {{#if isActor}}actor-img{{/if}}" />
+      <img src="{{item.img}}" class="item-img {{#if isActor}}actor-img{{/if}}" draggable="true"/>
```
CarlosFdez: you might be able to simplify this if you put the draggable on the img-portrait.

module/applications/sheets/api/application-mixin.mjs
nsalyzyn: Moved back. It's a bit ugly in some cases (eg. it will move hope dice if you drag from them), but I think that's fine.

styles/less/global/inventory-item.less
```diff
@@ -129,7 +129,9 @@
 
             .item-label {
                 flex: 1;
-                align-self: center;
+                align-self: stretch;
+                align-content: center;
+
```
CarlosFdez, line 134:
```suggestion
```

CarlosFdez:
Doesn't seem like it works anymore (dragging on the quantity field drags the whole thing). As for the inherit thing, it seems that the super one only does effects anyways, so I was off base and misread it.

A drag selector can use a comma to separate the entries to keep it as one entry, if they're related enough. `[data-item-id][draggable="true"], [data-item-id] [draggable="true"]` might work fine.

nsalyzyn:
Oh, what a thing to skip when testing! The actual bug.

Now I know that adding the dragSelector adds draggable=true to the element! I'll only fix the character sheet, since it's currently the only one with a text box. I'll also add the image thing to just the character sheet, since it looks pretty awful without doing so.

CarlosFdez:
I'd like to wait till after https://github.com/Foundryborne/daggerheart/pull/1431 to then test this one after.

Ikraik:
@CarlosFdez #1431 is now merged. Go ahead for review.

nsalyzyn:
Rebased and retested this along with the currency transfer functionality.

PR #1446, opened 2025-12-19
[PR] Adding max attributes to the action list and missing localization for them

### LINKED ISSUE #1137: [Bug] ActiveEffect Autocomplete doesn't return "max" for bar resources
**Describe the bug**
In ActiveEffect settings dialog, when you select bar attribute key (like hit points, stress, etc) the returned string is the resource object instead of its max value

**To Reproduce**

**Expected behavior**

**Screenshots**

**Setup Information:**
 - OS: [e.g. iOS, Windows]
 - Foundry Version [e.g. v13 b342]
- System Version [e.g. v.1.0, v.1.0.1]

**Additional context**

nsalyzyn:
## Description

Added an option for max values for those in the `bar` category of attributes. eg. stress, hp and hope.

- Fixes #1137 

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Code cleanup/refactor
- [ ] Documentation update
- [ ] Test coverage
- [ ] Dependency update
- [ ] Configuration change

## How Has This Been Tested?

In the effect dialog, under changes and in the autocomplete popup:

* Looking for max hp, max stress in adversaries
* Looking for max hp, max stress, and max hope in characters
* Looking for max stress in companions

- [X] Manual testing
- [ ] Other:

Screenshot 2025-12-18 201136

## Checklist

### DIFF
--- a/templates/sheets-settings/adversary-settings/details.hbs
+++ b/templates/sheets-settings/adversary-settings/details.hbs
@@ -22,12 +22,12 @@
         <fieldset class="flex">
             <legend>{{localize "DAGGERHEART.GENERAL.HitPoints.plural"}}</legend>
             {{formGroup systemFields.resources.fields.hitPoints.fields.value value=document._source.system.resources.hitPoints.value label=(localize "DAGGERHEART.ACTORS.Adversary.FIELDS.resources.hitPoints.value.label")}}
-            {{formGroup systemFields.resources.fields.hitPoints.fields.max value=document._source.system.resources.hitPoints.max}}
+            {{formGroup systemFields.resources.fields.hitPoints.fields.max value=document._source.system.resources.hitPoints.max label=(localize "DAGGERHEART.ACTORS.Adversary.FIELDS.resources.hitPoints.max.label")}}
         </fieldset>
         <fieldset class="flex">
             <legend>{{localize "DAGGERHEART.GENERAL.stress"}}</legend>
             {{formGroup systemFields.resources.fields.stress.fields.value value=document._source.system.resources.stress.value label=(localize "DAGGERHEART.ACTORS.Adversary.FIELDS.resources.stress.value.label")}}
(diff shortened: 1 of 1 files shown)


Ikraik:
hmm not keen on this for me. It is good for search but not sure. I tagged Harry for review as he was assigned to it earlier and would have a better hold if this is good or not.

module/data/actor/adversary.mjs
```diff
@@ -59,15 +59,13 @@ export default class DhpAdversary extends BaseDataActor {
                     0,
                     0,
                     'DAGGERHEART.GENERAL.HitPoints.plural',
-                    true,
-                    game.i18n.localize('DAGGERHEART.GENERAL.max')
```
WBHarry, line 64: I think we should keep the custom label here. The reason is that we are constrained for space, and just having `current`/`max` should be fine, since the fields are in a fieldset context already, making it clear what they are referring to.
Removing it leads to a title that won't fit in any good way.
image
Compared to
image

module/data/actor/adversary.mjs
```diff
@@ -59,15 +59,13 @@ export default class DhpAdversary extends BaseDataActor {
...
+                    true
                 ),
                 stress: resourceField(
                     0,
                     0,
                     'DAGGERHEART.GENERAL.stress',
-                    true,
-                    game.i18n.localize('DAGGERHEART.GENERAL.max')
```
WBHarry, line 71: Same as above.

nsalyzyn:
I see. That would leave two entries in the autocomplete that read as "Max" in the adversary section... I'm not sure that will be easy to workaround, since the translations in the autocomplete seem pretty automated. I might have to find locations where max needs to be shortened elsewhere - like the example you've shown.

WBHarry:
Ah, right. Well, another way around it is to do the change you did, and then just overwrite the label in the templates

WBHarry (approved):
🤖 :+1:

PR #1452, opened 2025-12-19
[PR] Checking for empty data in renderChatMessageHTML calls

### LINKED ISSUE #1199: [Bug] Error when trying to create module's ChatMessage
### **Describe the bug**

Using the `renderChatMessageHTML` hook callback causes an error when messages are rendered via a subtype’s `renderHTML` method. This happens because the hook’s third argument only exists when the default rendering path is used, not when the subtype’s `renderHTML` method is called.

---

### **To Reproduce**

Steps to reproduce:

1. Create a message using the subtype’s `renderHTML` method.
2. Observe the error thrown when the `renderChatMessageHTML` hook executes.

---

### **Screenshots**

Image  

---

### **Additional Context**

Refer to the Foundry source code for details on the hook call order and the `renderHTML` implementation.

```js
if (typeof this.system.renderHTML === "function") {
  const html = await this.system.renderHTML({ canDelete, canClose, ...rest });
  Hooks.callAll("renderChatMessageHTML", this, html);
  return html;
}
```

nsalyzyn:
## Description

To support modules modification of chat messages (ie. via https://github.com/foundryvtt/foundryvtt/issues/11984), made some of the calls optional in renderChatMessageHTML callbacks.

- Fixes #1199 

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Code cleanup/refactor
- [ ] Documentation update
- [ ] Test coverage
- [ ] Dependency update
- [ ] Configuration change

## How Has This Been Tested?

1. Added a roll to the chat message bar
2. Put the code snipped below into `module/data/chat-message/actorRoll.mjs` - this might be something a module would do to fully customize the content of a chat message
3. Refreshed the application
4. Saw that the roll was replaced with dummy content.

```js
     renderHTML(_) {
         return foundry.utils.parseHTML('<div>dummy content</div>');
     }
```

- [X] Manual testing
- [ ] Other:

## Checklist

### DIFF
--- a/daggerheart.mjs
+++ b/daggerheart.mjs
@@ -193,11 +193,9 @@ Hooks.on('ready', async () => {
 
 Hooks.once('dicesoniceready', () => {});
 
-Hooks.on('renderChatMessageHTML', (app, element, _) => {
+Hooks.on('renderChatMessageHTML', (document, element) => {
     enricherRenderSetup(element);
-    // Despite documentation, the third parameter may be empty as per
-    // https://github.com/foundryvtt/foundryvtt/issues/11984
-    const cssClass = app.flags?.daggerheart?.cssClass;
+    const cssClass = document.flags?.daggerheart?.cssClass;
     if (cssClass) cssClass.split(' ').forEach(cls => element.classList.add(cls));
 });
 
--- a/module/applications/ui/chatLog.mjs
+++ b/module/applications/ui/chatLog.mjs
@@ -55,10 +55,8 @@ export default class DhpChatLog extends foundry.applications.sidebar.tabs.ChatLo
         ];
     }
 
-    addChatListeners = async (app, html, data) => {
-        // Despite documentation, the third parameter may be empty as per
-        // https://github.com/foundryvtt/foundryvtt/issues/11984
-        const message = data?.message ?? app.toObject(false);
+    addChatListeners = async (document, html, data) => {
+        const message = data?.message ?? document.toObject(false);
         html.querySelectorAll('.simple-roll-button').forEach(element =>
             element.addEventListener('click', event => this.onRollSimple(event, message))
         );


CarlosFdez:
The first argument is the message document itself, so I think you can swap the code to using that instead of the plain object copy in the message data.

nsalyzyn:
Unfortunately, the message data in the ChatMessage is only accessible through a private accessor (https://foundryvtt.com/api/classes/foundry.documents.ChatMessage.html#_source). I've made the change and here's what that looks like. But, I doubt you'll like it.

That said, it would let a module customize the chat messages _and_ still use all of existing chat hooks, like rerolling dice. So, maybe it's necessary.

CarlosFdez:
Rename first argument to `message` or `document`, since its the actual ChatMessage client document (not an application subclass). Then use `data?.message ?? document.toObject(false)` to save on cycles. You can avoid the the comment about despite the documentation.

You won't need this for `renderChatMessageHTML` though, I think you can access the document there directly. Its the other one that breaks?

daggerheart.mjs
```diff
@@ -193,9 +193,11 @@ Hooks.on('ready', async () => {
 
 Hooks.once('dicesoniceready', () => {});
 
-Hooks.on('renderChatMessageHTML', (_, element, message) => {
+Hooks.on('renderChatMessageHTML', (app, element, _) => {
```
CarlosFdez:
```suggestion
Hooks.on('renderChatMessageHTML', (document, element) => {
```

daggerheart.mjs
```diff
@@ -193,9 +193,11 @@ Hooks.on('ready', async () => {
...
     enricherRenderSetup(element);
-    const cssClass = message.message.flags?.daggerheart?.cssClass;
+    // Despite documentation, the third parameter may be empty as per
+    // https://github.com/foundryvtt/foundryvtt/issues/11984
+    const cssClass = app.flags?.daggerheart?.cssClass;
```
CarlosFdez:
```suggestion
    const cssClass = document.flags?.daggerheart?.cssClass;
```

module/applications/ui/chatLog.mjs
```diff
@@ -56,26 +56,29 @@ export default class DhpChatLog extends foundry.applications.sidebar.tabs.ChatLo
     }
 
     addChatListeners = async (app, html, data) => {
+        // Despite documentation, the third parameter may be empty as per
+        // https://github.com/foundryvtt/foundryvtt/issues/11984
+        const message = data?.message ?? app.toObject(false);
```
CarlosFdez:
```suggestion
    addChatListeners = async (document, html, data) => {
        const message = data?.message ?? document.toObject(false);
```

nsalyzyn:
Tested as many cases as I could.

* css still shows up
* Dice rerolls still work
* The original bug is actually addressed

PR #1454, opened 2025-12-21
[Fix] 1453 - Async Resource Generation Errors

### LINKED ISSUE #1453: [Bug] Attacking with Hope using an Experience does not work Hope-balance wise
**Describe the bug**
If rolling with Hope for an Attack using an Experience, the balance of Hope should be +/- 1 = 0, but it does reduce it by one while the scrolling text correctly shows +/- 1. Happens starting with 3 Hope, so no problem at corner cases of hitting max or zero in the process.

**To Reproduce**
1. Go to Attack with any weapon
2. Click on Attack, use any Experience
3. See error if rolling with hope (see above)

Might not be always reproducable. While I am getting a 20+ bug streak now, last night it did seem to work sometimes, but most times not. Did reproduce now with only GM, so no multiplayer race conditions. If it is a race condition it is local.

**Expected behavior**
no Hope change

**Screenshots**
- https://www.youtube.com/watch?v=O19u-oH6tFY

**Setup Information:**
 - OS: Windows
 - Browser chrome
 - Foundry Version 13.351
- System Version 1.3.2
- all modules deactivated

WBHarry:
Fixes #1453 
Fixed so that we do not run separate actor.modifyResource calls during actions and dice rolls.

**Previously**
Either indirectly through hook calls, or directly we were calling `actor.modifyResource` in multiple places during action/dice execution. This easily ends up with race conditions where one update gets lost.
**Now**
We instead add all resource updates to a map in the config object: `resourceUpdates`. I made an extended map class for this that has a conveniance method for adding resources so updates to the same resource are smashed together into one, aswell as a method to run the `actor.modifyResource`method.

---
The case that was raised where we previously got an issue was when using hope/fear automation, aswell as using an experience. Hope result + using experience (costs hope) ended up with only one of the hope changes winning out. Now it works.

### DIFF
--- a/module/data/action/baseAction.mjs
+++ b/module/data/action/baseAction.mjs
@@ -241,7 +241,7 @@ export default class DHBaseAction extends ActionMixin(foundry.abstract.DataModel
             selectedRollMode: game.settings.get('core', 'rollMode'),
             data: this.getRollData(),
             evaluate: this.hasRoll,
-            resourceUpdates: new ResourceUpdateMap(this.actor.uuid)
+            resourceUpdates: new ResourceUpdateMap(this.actor)
         };
 
         DHBaseAction.applyKeybindings(config);
@@ -332,11 +332,12 @@ export default class DHBaseAction extends ActionMixin(foundry.abstract.DataModel
 }
 
 export class ResourceUpdateMap extends Map {
-    #actorUuid;
+    #actor;
 
-    constructor(actorUuid) {
+    constructor(actor) {
         super();
-        this.#actorUuid = actorUuid;
+
+        this.#actor = actor;
     }
 
     addResources(resources) {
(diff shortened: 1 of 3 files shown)


CarlosFdez:
Any chance of allowing ResourceUpdateMap to take an actor, and use fromUuid in the one place that only has the uuid? Or is that going to be tricky? It looks like in most places you're calling it you already have the actor.

WBHarry:
That'd be fine, certainly. I figured it might be neat this way, but I suppose it is ultimately potentially less stable to keep a uuid value rather than sending on in at the correct time.

CarlosFdez:
My concern was more of simplicity rather than stability, I assume its stable either way.

WBHarry:
Changed it to take in an actor rather than an actorUuid

PR #1455, opened 2025-12-22
[PR][Fix] 1448 - Player Reaction Rolls

### LINKED ISSUE #1448: Action that triggered reaction roll does not update results in the chat prompt.
Hi, I wanted to ask how to enable automated recognition of reaction roll results. Currently, when an adversary triggers a reaction roll, they roll damage for that roll and the players are prompted with a reaction roll. However, when they finish, it does not update the attack in the chat – the white shield does not change to either failure or success. As a result, when I press the 'deal damage' button, it says that the targets have not rolled their reaction rolls yet.

Can someone please tell me how to fix this? Perhaps I have the wrong automation settings.

Image
Image
Image
Image

WBHarry:
Fixes #1448 
Fixed so that players rolling reactions will update the message


Ikraik:
Hmm Branch has conflicts from a DHRoll PR?

PR #1458, opened 2025-12-22
[PR][Feature] Adversary Compendium Name Lookup

WBHarry:
- Exchanged all name references for @lookup. 
- Removed duplicated descriptions on feature actions
- Corrected Edit tooltip localization.
- Corrected so that the description fallback for actions to the feature description is used generally.


CarlosFdez (approved):
Code looks fine, I'm not reviewing all the adversaries.

PR #1459, opened 2025-12-22
[PR][Fix] Cleaned Up Countdown Ownership View

WBHarry:
- Removed the disabled default ownership select in the ownership view. 
- Added a fallback text incase there are no players added yet


CarlosFdez:
I got the no players to assign ownership to message when there are players in the world.

PR #1461, opened 2025-12-22
[Fix] Levelup New Experience Increase

WBHarry:
- Made experiences gained in the levelup screen available to be increased when you select the +1 to two experiences option.
- Fixed a bug introduced during this dev cycle where experiences in the roll dialog didn't get any dataPrep bonuses (such as from levelup) applied to them.

### DIFF
--- a/module/applications/levelup/characterLevelup.mjs
+++ b/module/applications/levelup/characterLevelup.mjs
@@ -283,9 +283,10 @@ export default class DhCharacterLevelUp extends LevelUpBase {
                                     const allExperiences = {
                                         ...this.actor.system.experiences,
                                         ...Object.values(this.levelup.levels).reduce((acc, level) => {
-                                            Object.keys(level.achievements.experiences).forEach(
-                                                x => (acc[x] = level.achievements.experiences[x])
-                                            );
+                                            for (const key of Object.keys(level.achievements.experiences)) {
+                                                acc[key] = level.achievements.experiences[key];
+                                            }
+
                                             return acc;
                                         }, {})
                                     };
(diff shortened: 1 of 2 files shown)


module/applications/levelup/characterLevelup.mjs
```diff
@@ -280,11 +280,18 @@ export default class DhCharacterLevelUp extends LevelUpBase {
                                     break;
                                 case 'experience':
                                     if (!advancement[choiceKey]) advancement[choiceKey] = [];
+                                    const allExperiences = {
+                                        ...this.actor.system.experiences,
+                                        ...Object.values(this.levelup.levels).reduce((acc, level) => {
+                                            Object.keys(level.achievements.experiences).forEach(
```
CarlosFdez: can we use a for loop here p.please? When I skimmed this, it looked like you were doing map on a quick skim and I had to double take.

module/applications/levelup/levelup.mjs
```diff
@@ -357,11 +357,22 @@ export default class DhlevelUp extends HandlebarsApplicationMixin(ApplicationV2)
 
         const experienceIncreaseTagify = htmlElement.querySelector('.levelup-experience-increases');
         if (experienceIncreaseTagify) {
+            const allExperiences = {
+                ...this.actor.system.experiences,
+                ...Object.values(this.levelup.levels).reduce((acc, level) => {
+                    Object.keys(level.achievements.experiences).forEach(
```
CarlosFdez: Same here too.

CarlosFdez (approved):

PR #1469, opened 2025-12-24
[PR][Feature] Support drag/drop resorting of adversary features

CarlosFdez:
All adversary features are sorted by feature type, and then by the sort property responsive to drag/drop reordering.


Ikraik:
should we have an automated alphabetical sort option?

CarlosFdez:
Not sure where I'd put the toggle to place that. We could also force alphabetical and disable drag drop resort entirely.

CarlosFdez:
After discussion elsewhere it looks like everyone is leaning on not alphabetical, with limited desire for alphabetical sorting.

WBHarry (approved):
🤖 :+1:

PR #1470, opened 2025-12-24
[Feature] TokenConfig Actor Size Edit

WBHarry:
Added Actor TokenSize (Tiny, Small, Medium etc) to the token config to make it smoother when you do want to set it to custom to change dimensions. Currently you have to go to the actor edit first to set it to custom which isn't great.

https://github.com/user-attachments/assets/b357e0bf-104e-4ca7-aef5-960b9afa19ff

### DIFF
--- a/templates/sheets-settings/token-config/appearance.hbs
+++ b/templates/sheets-settings/token-config/appearance.hbs
@@ -10,7 +10,6 @@
         </select>
     </div>
     {{/if}}
-    Test
     <fieldset>
         <legend>{{localize "Token Size"}}</legend>
         {{#if usesActorSize}}


module/applications/sheets-configs/prototype-token-config.mjs
```diff
@@ -34,8 +40,89 @@ export default class DhPrototypeTokenConfig extends foundry.applications.sheets.
...
+     */
+    static async #onSubmit(event, form, formData) {
+        const submitData = this._processFormData(event, form, formData);
+        submitData.detectionModes ??= []; // Clear detection modes array
+        this._processChanges(submitData);
+        const changes = { prototypeToken: submitData };
+
+        const tokenSizeSelect = this.element.querySelector('#dhTokenSize');
+        if (tokenSizeSelect && this.actor) {
+            if (tokenSizeSelect.value !== this.actor.system.size) {
```
CarlosFdez: I think for this it would be neater to be one if. If it formats to two lines you could store isSizeDifferent in a variable.

CarlosFdez:
Thoughts on doing what foundry does and making a mixin to reduce the duplicate code? Would that work?

WBHarry:
> Thoughts on doing what foundry does and making a mixin to reduce the duplicate code? Would that work?

I was thinking of that. The issue I find is that foundry's prototypeConfig and tokenConfig extend different mixins.
Hm. When I was messing with it I was thinking this would make it not working..but thinking of it, that's not true. I can just make a wrapping mixin, yeah.

I'll give it a shot.

module/applications/sheets-configs/prototype-token-config.mjs
WBHarry: Shortened it to one if :+1:

WBHarry:
Extracted all the common logic to token-config-mixin.mjs. The actual classes are now very tiny 😎

templates/sheets-settings/token-config/appearance.hbs
```diff
@@ -10,22 +10,32 @@
         </select>
     </div>
     {{/if}}
+    Test
```
CarlosFdez:
```suggestion
```

WBHarry: Everyone's a hater 😤 

Yah, removed the forgotten testing thingy :grin:

CarlosFdez (approved):

PR #1471, opened 2025-12-25
[Fix] LookupEnricher Without Actor

WBHarry:
- Fixed so that lookupEnricher's fallback state is securely returning the lookup string when the lookup fails.

The specific issue that was raised was that when pulling out a feature from an adversary that was making use of `@Lookup[@name]` to the item sidebar to be standalone. It would throw an error on viewing the sheet as the lookupEnricher would fail.

### DIFF
--- a/module/enrichers/LookupEnricher.mjs
+++ b/module/enrichers/LookupEnricher.mjs
@@ -1,8 +1,14 @@
 import { parseInlineParams } from './parser.mjs';
 
 export default function DhLookupEnricher(match, { rollData }) {
-    const results = parseInlineParams(match[1], { first: 'formula' });
+    const lookupParam = match[1];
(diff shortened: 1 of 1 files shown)


CarlosFdez:
Let me look into this, I think we can get it to work if the formula doesn't resolve for a particular actor (may be handy once we expand to origins).

WBHarry:
Sure!

Just to be clear. The issue that was happening before was specifically when a feature was -not- on an actor. It made `rollData` come in as null, which ended up with an error.

WBHarry:
Yah, this is a deeper issue that's rooted in Actions being set up to return the rollData for an actor they are on in their `getRollData` method.
Needs to be looked at. The initial workound I did here isn't the good way to solve it.

PR #1475, opened 2025-12-26
[PR] Adding the ability to target downtime actions.

### LINKED ISSUE #1161: [Bug] Downtime moves Tend to Wounds and Repair Armor - targetting issues
**Describe the bug**
For the Downtime moves Tend to Wounds and Repair Armor allow the target to be Self or an Ally. The target shown (development branch tested) is self only and does not allow change of target.

**To Reproduce**
1. Create a character
2. Choose a downtime
3. Select Tend to Wounds and Repair Armor
4. Roll to get the values
5. Note the target is self and can't be changed.

**Expected behavior**
The chosen target is healed appropriately.  And self targeted if no target chosen.

Note: you can select each one twice (e.g. 2 x Tend to Wounds), so it's important to be able to target.

nsalyzyn:
## Description

Changed the appropriate default downtime actions to target 'friendly' and made the long rest actions heal "100" HP or Armor - since the current player's max HP and max Armor may be lower than the target's. Then, added a selection box underneath the appropriate actions that allows the selection of a target - using all visible characters as options. Defaults to self. Unfortunately, the selection on this `<select>` box is lost of reloading of the application. Next, this PR plumbs through the target uuid to be used when the action is invoked.

I'm marking this as a feature, since it seems like it's more than just a bug fix to add this functionality.

- Fixes #1161

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Code cleanup/refactor
- [ ] Documentation update
- [ ] Test coverage
- [ ] Dependency update
- [ ] Configuration change

## How Has This Been Tested?

* Tested as the GM with various mixes of moves.
* Tested as a player trying to select another player - of course the GM will need to hit the final apply action for permission reasons.

- [X] Manual testing
- [ ] Other:

alternative

## Checklist

## Additional Comments

There is a pre-existing console error that doesn't seem to do anything. If the player does all the steps, but then another player of the GM clicks 'apply healing', the following error is in the logs:

```
costField.mjs:37 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'system')
    at DHHealingAction.execute (costField.mjs:37:34)
    at DHHealingAction.consume (baseAction.mjs:285:42)
    at DhpChatMessage.consumeOnSuccess (chatMessage.mjs:248:88)
    at DhpChatMessage.onApplyDamage (chatMessage.mjs:191:14)
```

Given the steps needed for a player to heal another player, this error is likely to show up. It was an unlikely corner case beforehand.

### DIFF
--- a/module/applications/ui/chatLog.mjs
+++ b/module/applications/ui/chatLog.mjs
@@ -135,7 +135,7 @@ export default class DhpChatLog extends foundry.applications.sidebar.tabs.ChatLo
     async actionUseButton(event, message) {
         const { moveIndex, actionIndex, movePath } = event.currentTarget.dataset;
         const targetUuid = event.currentTarget.closest('.action-use-button-parent').querySelector('select')?.value;
-        const parent = await foundry.utils.fromUuid(message.system.actor);
+        const parent = await foundry.utils.fromUuid(targetUuid || message.system.actor)
 
         const actionType = message.system.moves[moveIndex].actions[actionIndex];
         const cls = game.system.api.models.actions.actionsTypes[actionType.type];
--- a/module/config/generalConfig.mjs
+++ b/module/config/generalConfig.mjs
@@ -350,7 +350,7 @@ export const defaultRestOptions = {
                                 value: {
                                     custom: {
                                         enabled: true,
-                                        formula: '100'
+                                        formula: '@system.resources.hitPoints.max'
                                     }
                                 }
                             }
@@ -416,7 +416,7 @@ export const defaultRestOptions = {
                                 value: {
                                     custom: {
                                         enabled: true,
-                                        formula: '100'
+                                        formula: '@system.armorScore'
                                     }
                                 }
                             }


nsalyzyn:
I have another version of this where the target selection is in the downtime dialog itself. If that's preferred, I can send that PR out instead.

nsalyzyn:
Here's what the alternative looks like. Keep in mind, it might make the dialog too tall as per #1480.

mixed-rest
simple-long-rest

WBHarry:
Hm..considering the matter with dialog height, I think we can keep it in the chat message.

An issue I've come across is that when using Treat Wounds or Repair Armor in this branch, you end up with 100 healing.
This might be because the branch is not up to date with main. Would be good to merge in main so we can see that it works 🙃

nsalyzyn:
I rebased this onto main, but I think you are seeing an intentional change I made in this PR. Long rest treat wounds and repair armor do 100 healing. Is there a way to make a formula that uses the target's resource counts? Giving a large number was the only way I could think of.

See the changes made in `generalConfig.mjs`.

nsalyzyn:
I've fixed this to address your concerns. It now uses the target as the action's parent, which uses it as the roll target. I ran through the manual tests again.

PR #1486, opened 2025-12-28
[Fix] Evolved Beastform Wildcard

WBHarry:
Fixed so picking an evolved beastform will properly handle wildcard images

### DIFF
--- a/module/applications/dialogs/beastformDialog.mjs
+++ b/module/applications/dialogs/beastformDialog.mjs
@@ -286,13 +286,9 @@ export default class BeastformDialog extends HandlebarsApplicationMixin(Applicat
                     if (data) {
                         if (!data.selectedImage) selected = null;
                         else {
-                            if (evolved) {
-                                if (evolved.usesDynamicToken) evolved.system.tokenRingImg = data.selectedImage;
-                                else evolved.system.tokenImg = data.selectedImage;
-                            } else {
-                                if (data.usesDynamicToken) selected.system.tokenRingImg = data.selectedImage;
-                                else selected.system.tokenImg = data.selectedImage;
-                            }
+                            const imageSource = evolved ?? selected;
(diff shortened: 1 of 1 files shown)


module/applications/dialogs/beastformDialog.mjs
```diff
@@ -278,19 +278,30 @@ export default class BeastformDialog extends HandlebarsApplicationMixin(Applicat
...
                         else {
-                            if (data.usesDynamicToken) selected.system.tokenRingImg = data.selectedImage;
-                            else selected.system.tokenImg = data.selectedImage;
+                            if (evolved) {
+                                if (evolved.usesDynamicToken) evolved.system.tokenRingImg = data.selectedImage;
+                                else evolved.system.tokenImg = data.selectedImage;
+                            } else {
+                                if (data.usesDynamicToken) selected.system.tokenRingImg = data.selectedImage;
+                                else selected.system.tokenImg = data.selectedImage;
+                            }
```
CarlosFdez: these two branches look about the same, other than the first branch checks evolved and assigns evolved, and the second checks data and assigns selected. There a way to consolidate?

WBHarry: There is. I just thought it looked uglier to do the tertiary on the assignment because of the length 🤔

WBHarry: Changed it around to be like this:
```
const imageSource = evolved ?? selected;
if (imageSource.usesDynamicToken) imageSource.system.tokenRingImg = data.selectedImage;
else imageSource.system.tokenImg = data.selectedImage; 
```

CarlosFdez (approved):
Didn't test it, but looks fine conceptually. If you want me to test it I can tho.

WBHarry:
I think it's fine. It's a very distinct case that I went over back and forth a good few times 🙃

Issue #1492, opened 2025-12-29
[Bug] "Paired" effect considers only top left square of a large creature when applied automatically


mudo007:
**Description**
All but the top left square of a large adversary (2x2) triggers the automatic application and removal of the Paired condition 

**To Reproduce**
1. Create an actor and equip it with a dagger and a small dagger, place it on the canvas
2. Place a large adversary south of the actor on a close, but not very close range
3. Select the actor, and then select the adversary as its target
4. Move the actor into melee range (south edge) of the adversary, 
5. mouseover the adversary and confim it shows "melee" range
6. The "Paired"condition is not applied
7. Move the actor one more square north, overlapping its token with the adversary's
8. The "paired" condition is applied

**Expected behavior**
The Paired condition should have been applied as soon as the actor reached melee range of the adversary when approaching from its south or eastern edges.

**Setup Information:**
 - OS: Mac Os X 15.7.1
 - Browser Chrome
 - Foundry Version  v13 b351
- System Version v 1.4.2

**Additional context**
It is also reproducible when the actor moves into melee range from east to west, but not reproducible when approaching from west to east or from north to south

nsalyzyn:
As per https://foundryvtt.com/api/interfaces/foundry.documents.types.TokenPosition.html, the token's position's x and y coordinate are in the top left. We'll also need a more comprehensive distance calculation formula, since just centering it will leave targets 'very close' rather than melee.

PR #1497, opened 2026-01-01
[PR][Fix] Detection of range dependencies

CarlosFdez:
This adds a new range detection function that attempts to handle the use case of large or larger tokens. It also adjusts the hooks to remove issues that may occur when moving into the range of something for the first time using wasd. I recommend not releasing this without some testing in main for a bit first in case of errors.

Closes https://github.com/Foundryborne/daggerheart/issues/1492

### DIFF
--- a/daggerheart.mjs
+++ b/daggerheart.mjs
@@ -330,14 +330,14 @@ const updateActorsRangeDependentEffects = async token => {
                 break;
             }
 
-            const distanceBetween = canvas.grid.measurePath([
-                userTarget.document.movement.destination,
-                token.movement.destination
-            ]).distance;
-            const distance = rangeMeasurement[range];
-
+            // Get required distance and special case 5 feet to test adjacency
+            const required = rangeMeasurement[range];
             const reverse = type === CONFIG.DH.GENERAL.rangeInclusion.outsideRange.id;
-            if (reverse ? distanceBetween <= distance : distanceBetween > distance) {
+            const inRange =
+                required === 5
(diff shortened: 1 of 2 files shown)


WBHarry:
Very complicated to try and grasp, but something is definately wrong. I can't seem to ever get the paired condition to be applied. The distance keeps being calculated as 20, even when the tokens are inside of each other.
image

WBHarry:
image

CarlosFdez:
Might there be multiple of the same linked actor on the scene?

WBHarry (approved):
Yup, that must have been the issue. 
The change works well - the activation/deactivation is a lot more stable and reliable with the different hook used. It even updates if you target something that is -already- in range.
👍

Ikraik:
@WBHarry @CarlosFdez okay to merge?

WBHarry:
Depends. If we want to bring out a little patch before all the features of 1.5, then we probably want to wait until after that.

CarlosFdez:
Its tricky. Its a bugfix, not a new feature, but if something is wrong with it you'd need a second bugfix,

WBHarry:
Hm..really, I think it seems to work well - and the feature itself is rather fringe. It's an optional autmation behavior, and the only usecase I think we have atm is on `Paired`.
I think I'm fine with bringing it in for the patch.

CarlosFdez:
I didn't consider that it was only for paired right now. Yeah, good point.

PR #1499, opened 2026-01-02
[Feature] Scene Environments

WBHarry:
Added the ability to link environments to scenes for easy access.

Foundry scenes do not add a system property for subtypes, and our data is therefore stored in a flag.
This means that in difference to the `Party` actors implementation of `partyMembers`, we cannot make use of the `ForeignDocumentUUIDArrayField` fields `prune` property to neatly handle the data refresh when linked Environments are deleted.
Because of this, I had to implement a data updata in `environment.onUpdate` run by the active GM to make it happen. If Scenes are ever changed to have system data, then we can remove all that.

### DIFF
--- a/module/applications/scene/sceneConfigSettings.mjs
+++ b/module/applications/scene/sceneConfigSettings.mjs
@@ -40,7 +40,7 @@ export default class DhSceneConfigSettings extends foundry.applications.sheets.S
 
     static TABS = DhSceneConfigSettings.buildTabs();
 
-    async _preFirstRender(context, options) {
+    async _preRender(context, options) {
         await super._preFirstRender(context, options);
         this.daggerheartFlag = new game.system.api.data.scenes.DHScene(this.document.flags.daggerheart);
     }


module/applications/scene/sceneConfigSettings.mjs
```diff
@@ -28,36 +40,72 @@ export default class DhSceneConfigSettings extends foundry.applications.sheets.S
...
+            )
+        });
+        this.render();
+    }
+
+    /** @override */
+    async _processSubmitData(event, form, submitData, options) {
+        submitData.flags.daggerheart = this.daggerheartFlag.toObject();
+        if (this.document._source.flags.daggerheart) {
+            for (const key of Object.keys(this.document._source.flags.daggerheart.sceneEnvironments)) {
```
CarlosFdez: I had an error for an existing scene if the daggerheart flag already exists and you try to save changes. What if instead of the if and the for on lines 101 and 102, you do a single for loop with this:

`for (const key of Object.keys(this.document._source.flags.daggerheart?.sceneEnvironments ?? {})) {`

WBHarry: Ah, yeah true. Was intending to safeguard existing daggerheart flags, but the if stopped a bit short.
I swapped over to just the for loop :+1:

module/documents/scene.mjs
```diff
@@ -37,4 +37,30 @@ export default class DhScene extends Scene {
...
+
+    _onDelete(options, userId) {
+        super._onDelete(options, userId);
+
+        if (this instanceof game.system.api.documents.DhScene) {
+            const system = new game.system.api.data.scenes.DHScene(this.flags.daggerheart);
+
+            // Clear this scene from all environments that aren't deleted
+            for (const environement of system.sceneEnvironments) {
+                environement?.system?.scenes?.delete(this);
```
CarlosFdez:
```suggestion
            for (const environment of system.sceneEnvironments) {
                environment?.system?.scenes?.delete(this);
```

module/applications/scene/sceneConfigSettings.mjs
```diff
@@ -28,36 +40,70 @@ export default class DhSceneConfigSettings extends foundry.applications.sheets.S
 
     static TABS = DhSceneConfigSettings.buildTabs();
 
+    async _preFirstRender(context, options) {
+        await super._preFirstRender(context, options);
+        this.daggerheartFlag = new game.system.api.data.scenes.DHScene(this.document.flags.daggerheart);
```
CarlosFdez, line 45: Is there anything that keeps this up to date if the flag is independently updated in the scene? Should this be set every re-render maybe?

WBHarry: Hm. As it stands, we can't have it being reset.
We are keeping a local version in the config, and gradually updating it. The SceneConfig is not UpdateOnChange, so we build a local version, and then it's saved on hitting `Save Changes`

CarlosFdez: what happens if the scene is updated by a module or something while the dialog is open? Is there something to handle that case for the other properties, or no handling whatsoever? If no handling then I guess is fine, if yes, is there a way to copy what they do?

WBHarry: This is just a general limitation of forms that are not `submitOnChange: true`, right?
If we ever reset the current local buildup of changes, they'll just be lost. 🤔

CarlosFdez: Might be, though if the core scene stuff solved it we should copy their homework.

CarlosFdez: At a quick glance it looks like they do it by checking the form values....which is unfortunately quite messy. Not much we can do I think in this aspect.

WBHarry: Huh. Interesting. So the way it appearantly works:
- Forms like this accumlate their changes that are not saved
- If an external update happens, it does just throw away all those changes and loads the new version.

WBHarry: So I suppose we should do the same. 🙃

CarlosFdez: Wait, you found where the external update is done? Well, if there's a clean way to do it, that'll be handy, so long as there are no diffs in those fields.

WBHarry: No no. I just tested to see how the core behavior works.
1) Opened scene config
2) Edited a few fields - did not click save
3) I ran a console update on the scene's name
4) All my edited fields were reseted as the newly updated version was loaded in

CarlosFdez: My immediate guess is that it occured due to a re-render. Checking the source code, the preview is handled by pulling from form data. Anyways this looks messy so forget I said anything.

WBHarry: It's not a big change, but I simply changed from `_preFirstRender` to `_preRender`. Makes the behavior consistent with the core form behavior. :+1:

CarlosFdez: oh, well, if it worked we're good!

CarlosFdez (approved):

PR #1500, opened 2026-01-03
[Feature] Trigger System

WBHarry:
Added a trigger system that currently handles triggers registered from Actions. Essentially automated macros.
Added one to the Wizard feature StrangePatterns.


chrisryan10:
Just doing some testing with the Wizard Strange Patterns. The Trigger worked, but if you roll 2 1s on the Duality Roll (a double 1 crit), you only get the dialog once, when you should get the option twice.

WBHarry:
> Just doing some testing with the Wizard Strange Patterns. The Trigger worked, but if you roll 2 1s on the Duality Roll (a double 1 crit), you only get the dialog once, when you should get the option twice.

Redone the trigger to let you pick increases up to the max matches rolled :+1:

chrisryan10:
This works, although in my test env, I had to create a new Wizard character. The old one came up with some issues. Will this require a migration for existing characters?

I'll get it ready for merging if the migration is not an issue.

PR #1503, opened 2026-01-04
[PR][Feature] Add support for changing the tier of an adversary

CarlosFdez:
Still in the draft forms, the dialog should be different, needs localization, we need to handle inlines, as well as lesser smaller damages and spikes might need special handling.

### DIFF
--- a/module/data/actor/adversary.mjs
+++ b/module/data/actor/adversary.mjs
@@ -247,7 +247,7 @@ export default class DhpAdversary extends BaseDataActor {
                             adjusted.bonus
                         ]
                             .filter(p => !!p)
-                            .join(' + ');
+                            .join('+');
                         return match.replace(formula, newFormula);
                     } catch {
                         return match;
@@ -381,7 +381,7 @@ export default class DhpAdversary extends BaseDataActor {
             const value = this.#calculateAdjustedDamage(previousFormula, damageMeta);
             const formula = [value.diceQuantity ? `${value.diceQuantity}d${value.faces}` : null, value.bonus]
                 .filter(p => !!p)
-                .join(' + ');
+                .join('+');
             if (value.diceQuantity) {
                 data.custom.enabled = false;
                 data.bonus = value.bonus;


module/data/actor/adversary.mjs
```diff
@@ -188,4 +189,407 @@ export default class DhpAdversary extends BaseDataActor {
...
+        const convertDamage = (damage, newMean) => {
+            const hitPointParts = damage.parts.filter(d => d.applyTo === 'hitPoints');
+            if (hitPointParts.length === 1 && !hitPointParts[0].value.custom.enabled) {
+                const value = hitPointParts[0].value;
+                value.flatMultiplier = Math.max(0, value.flatMultiplier + tier - source.system.tier);
+                const baseAverage = (value.flatMultiplier * (Number(value.dice.replace('d', '')) + 1)) / 2;
+                value.bonus = Math.round(newMean - baseAverage);
+            } else {
+                ui.notifications.error(
+                    `Failed to convert item ${item.name}: Other kinds of damage is currently unsupported`
```
WBHarry: `item` from the loop is not available here in the helper function

module/data/actor/adversary.mjs
```diff
@@ -188,4 +189,407 @@ export default class DhpAdversary extends BaseDataActor {
...
+ * @property {number} hp
+ * @property {number} stress
+ * @property {number} attack
+ * @property {number[]} damage
+ */
+
+/** @typedef {Record<2 | 3 | 4, TierData> & Record<1, { damage: number[] }} AdversaryScalingData */
+
+/** @type {Record<string, AdversaryScalingData>} */
+export const ADVERSARY_DATA = {
```
WBHarry: I'd say this should go in the configs. It's where generally static stuff should live, and it's always possible some module would want to reference the standard deviation values the system makes use of.

We could make some specific config for it, but I think it's also fine to place it in `actorConfig`.

CarlosFdez: actorConfig works! I'll do that after the session.

WBHarry:
Hm. I think changing the `Duplicate To New Tier` dialog to have a select of the 4 tiers would be better than leaving it as an input.
Funnily, it currently returns `The actor is already at this tier` for any invalid input, such as typing in 555.

WBHarry:
I'm not sure if this is really something actionable, but I tested transforming a lot of things, and if there's one thing that -seems- off, it is upscaling adversary Features that are attacks with damage.
They seem to end up with pretty outrageous damage numbers, especially the bear, although I guess the bear might just be a heavy outlier on damage.
Tier 4 Bear and Tier 4 Archer Guard.
image
This is compared to a Kraken's basic attack of `4d10+10`

Edit: To be fair, it should also be expected that the upscaling/downscaling becomes less 'accurate' the further you do it. Normal useage should be raising/lowering something 1 tier rather than going 1-->4 or 4-->1.

CarlosFdez:
The caveats are fair, but I think its worth analyzing anyways. So I am testing both on Fresh Cut Grass, which I think uses a table instead of something mathematical.

So at tier 1, the creatures do:
**Bear**: 3d4 + 10 (17.5)
**Archer Guard**: 1d12 + 3 (9.5)

So the results of this algorithm were
**Bear**: 6d10 + 27 (60)
**Archer Guard**: 5d12 + 3 (35.5)

Fresh Cut Grass gives these results instead
**Bear**: 4d10 + 33 (55)
**Archer Guard**: 2d20 + 10 (31)

We should compare against players at these ranks.
At tier 1, a player in leather armor has thresholds 9/14. Archer hits 2HP on average but can swing below (not likely above), and Bear hits well above 3HP unless it wins to turd lottery.
At level 10, a player in legendary leather has thresholds 23/46. Archer hits 2HP on average but is perhaps too close to 3HP on ours. Bear seems within bounds either way.

The numbers are quite similar, though it appears that this mechanism skews upwards. It might be due to a greater representation of boss-like enemies at the last tier, and no standards. I've been considering adding a manual correction to tier 4 specifically because our data trends too high. Thoughts?

WBHarry:
Hm..yeah, it could be a good idea, if we are conciously aware that the current available Adversaries skews the results there.

Could perhaps add something like
```
/* Manual adjustments of damage due to perceived lack of adversary data */
const damageModifications = {
    tier4: {
        bruiser: -5,
        skulker: -5,
        ....
        solo: 0
    }
}
```
Or some such, since in this case solos are what we imagine are overrepresented.

WBHarry:
We could also say that it's fine, though. 
The difference wasn't actually as large as I though tit would be (60 vs 55)

CarlosFdez:
Given that the results in other ranks include all creatures, we would simply change the rank 4 numbers and leave a note. My thought is to maybe reduce the average by 6 and crank up the deviation a tad.

WBHarry:
Works :+1:

Ikraik:
@WBHarry Is this approved? Still says pending

WBHarry:
I think Supes had some ideas to tune the formula a bit

CarlosFdez:
Its as tuned as I can make it unless people find problems with it by testing it.

WBHarry:
I think it's all good from my testing.
At this point, the best potential refinements that can be made would be found out by letting users mess with it to increase how much its utilized. :+1:

WBHarry (approved):

CarlosFdez:
Sure, then lets wait two days, make sure there's no bugs with existing stuff, then get this in after.

WBHarry:
Minions are typically set up with a custom damage formula, since they always have a static damage like 1 or 5.
Currently, this means it's not being adjusted by `#convertDamage`
```js
// others are not supported yet. Later on we should convert to terms, then convert from terms back to real data
if (!(hitPointParts.length === 1 && !hitPointParts[0].value.custom.enabled)) return; 
```

CarlosFdez:
While the above has been fixed, descriptions of abilities in minions and hordes still need some handling. 

The Horde Passive can be handled by checking the item's effects, and if present, perform a replacement of the original alt formula. I can do that here soon.

Group attack will need to be adjusted on existing minions to use `@Lookup`. We might need an action's formula to be available as a getter. That might need to be handled in a separate PR.

PR #1514, opened 2026-01-08
[Feature] Summon Action Implementation

WBHarry:
Implementation of the summon action.

### DIFF
--- a/module/data/fields/action/summonField.mjs
+++ b/module/data/fields/action/summonField.mjs
@@ -79,7 +79,7 @@ export default class DHSummonField extends fields.ArrayField {
         summon.actor = result.actor;
 
         summon.count--;
-        if (summon.count === 0) {
+        if (summon.count <= 0) {
             summonIndex++;
             if (summonIndex === summonData.length) return actionActor.sheet?.maximize();
         }


module/data/fields/action/summonField.mjs
```diff
@@ -0,0 +1,89 @@
...
+        const summon = summonData[summonIndex];
+        const result = await CONFIG.ux.TokenManager.createPreviewAsync(summon.actor, {
+            name: `${summon.actor.prototypeToken.name}${summon.count > 1 ? ` (${summon.count}x)` : ''}`
+        });
+
+        if (!result) return actionActor.sheet?.maximize();
+        summon.actor = result.actor;
+
+        summon.count--;
+        if (summon.count === 0) {
```
CarlosFdez:
```suggestion
        if (summon.count <= 0) {
```

Please forgive my paranoia, I'm worried about the case where somehow summon.count === 0 gets into this function. It'd become summonCount = -1 and then recurse infinitely.

CarlosFdez:
Looks fine after a rudimentary look, and I assume you already got a much more thorough one!

module/data/fields/action/summonField.mjs
WBHarry: Doesn't hurt with some extra safety. Added 👍

PR #1515, opened 2026-01-08
Safety check for experiences

### LINKED ISSUE #1506: [Bug] /dr requires selected player character token
**Describe the bug**
Creating a generic `/dr` roll in chat requires a player character token to be selected because it tries to read experiences, which will not exist if a token is not selected.

On a player login that's setup with a character for the player, it seems to default to the player's character

**To Reproduce**
1. Log In as a GM
2. Don't select any tokens (or select an adversary token)
3. Roll in chat with `/dr`
4. No dialog is created. Console error shown below mentions it trying to read the experiences and failing

**Expected behavior**
When no character is selected, or no fallback is available for the character, it should do a generic roll without needing a character's experiences.

**Screenshots**

Image

**Setup Information:**
 - OS:  Windows
 - Browser chrome

Image

chrisryan10:
## Description

Safety check added when trying access experiences

- Fixes #1506 
- Closes #1506

## Type of Change

- [x] Bug fix
## How Has This Been Tested?

- [x] Manual testing

### DIFF
--- a/module/applications/dialogs/d20RollDialog.mjs
+++ b/module/applications/dialogs/d20RollDialog.mjs
@@ -104,7 +104,7 @@ export default class D20RollDialog extends HandlebarsApplicationMixin(Applicatio
             context.roll = this.roll;
             context.rollType = this.roll?.constructor.name;
             context.rallyDie = this.roll.rallyChoices;
-            const experiences = this.config.data?.system.experiences || {};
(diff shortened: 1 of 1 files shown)


Ikraik (approved):
LGTM as the command worked without player token.

👍 

As for implementation, I requested Harry to look in as well.

Issue #1438, opened 2025-12-18
[TYPO] - Display Feature Descriptions for Non-GM Users


brunocalado:
When a non-GM user opens the compendium to choose items, they are unable to see the feature description by hovering the mouse over it. If you are a GM or import the item into the world, the player can see it. However, players normally look directly in the compendium when they are choosing something.

I would like to suggest adding the feature description to the 'description' field in a way that allows players to view it directly within the compendium.

<img width="603" height="475" alt="Image" src="https://github.com/user-attachments/assets/602a56ac-ddbc-411f-b4c0-d73e93e00209" />

Ikraik:
Weapons don't have descriptions/lore. They have labels however.

Ikraik:
As for the feature is bound with, it is something to be checked anyways if feature description gets passed through on main description one way or another. Will see if our current implementation is allowing it to transfer. If not, the team will look into implementing it post discussion internally.

WBHarry:
#1521 has been solved and released. You can now see the descriptions of Item Features (Quick, Reliable etc) in an item's normal description. This makes it accessable to see in the Character Sheet inventory list without opening the item, aswell as in the Compendium Browser application.

It's not possible, or desireable to add information directly in the compendiums, as that would require always loading their data, which completly breaks the premise of compendiums, which is to not always load everything.

---

Considering this solved.

PR #1521, opened 2026-01-10
[Feature] Armor/Weapon Features In Description

WBHarry:
ItemFeatures are now prepended to the description. While editing, the text is not present.

image

### DIFF
--- a/styles/less/ux/tooltip/tooltip.less
+++ b/styles/less/ux/tooltip/tooltip.less
@@ -282,18 +282,6 @@ aside[role='tooltip'].locked-tooltip:has(div.daggerheart.dh-style.tooltip.card-s
         .tooltip-description {
             font-style: italic;
             text-align: start;
-
-            .feature-descriptions {
-                .features-header {
-                    font-size: 14px;
-                    font-weight: bold;
-                    text-decoration: underline;
-
-                    &.large {
-                        font-size: 18px;
-                    }
-                }
-            }
         }
 
         .simple-info {


CarlosFdez:
Thoughts on omitting the Item Features header?

WBHarry:
Works for me. I was going back and forth on it constantly while doing it 😆

WBHarry:
- Removed the header.
- Made a better separation of logic. ´getDescripionData´ now returns `{ prefix, value, suffix }`. `getEnrichedDescription` handles stitching these together with separators in accordance to which are not null.

module/data/item/base.mjs
```diff
@@ -124,6 +124,41 @@ export default class BaseDataItem extends foundry.abstract.TypeDataModel {
...
+        const appendWithSeparator = (text, add) => {
+            if (!add) return text;
+            if (text) return `${text}\n<hr>\n${add}`;
+            else return add;
+        };
+
+        const { prefix, value, suffix } = await this.getDescriptionData();
+        let fullDescription = prefix ?? '';
+        fullDescription = appendWithSeparator(fullDescription, value);
+        fullDescription = appendWithSeparator(fullDescription, suffix);
```
CarlosFdez:
```suggestion
        const { prefix, value, suffix } = await this.getDescriptionData();
        const fullDescription = [prefix, value, suffix].filter(p => !!p).join('\n<hr>\n');
```

didn't test it but should work hopefully.

styles/less/global/inventory-item.less
```diff
@@ -211,6 +211,18 @@
                     ul {
                         list-style: disc;
                     }
+
+                    .feature-descriptions {
```
CarlosFdez: this and the one in tooltip can maybe be killed hopefully.

WBHarry: Forgot the style was replicated in another place. Removed :+1:

module/data/item/base.mjs
WBHarry: Tested it, and it works. `join` makes short work of the logic cases. Changed it. 🥳

styles/less/ux/tooltip/tooltip.less
```diff
@@ -283,6 +282,18 @@ aside[role='tooltip'].locked-tooltip:has(div.daggerheart.dh-style.tooltip.card-s
         .tooltip-description {
             font-style: italic;
             text-align: start;
+
+            .feature-descriptions {
```
CarlosFdez: dis too

CarlosFdez: after this it looks fine to me! neato.

WBHarry: 🪄 Gone

Ikraik:
This PR would handle #1438 I feel

WBHarry:
Yup. Solves it. I'll make a note of that in the Issue and close that one 😎

WBHarry:
Or hm. I guess they are asking to be able to see the description straight in the compendium?

WBHarry:
That's not something we want to or can do however. That'd require always loading in compendium data, which breaks the whole premise of avoiding that by having stuff in compendiums.

PR #1522, opened 2026-01-10
[Feature] ActiveEffect Default  Hope/Fear dice

WBHarry:
Added ActiveEffect rules for default hope and fear dice so the roll dialog can be made to default to a specific dice size for hope and fear respectively.

### DIFF
--- a/system.json
+++ b/system.json
@@ -2,7 +2,7 @@
     "id": "daggerheart",
     "title": "Daggerheart",
     "description": "An unofficial implementation of the Daggerheart system",
-    "version": "1.4.6",
+    "version": "1.5.0",
     "compatibility": {
         "minimum": "13.346",
         "verified": "13.351",


module/data/actor/character.mjs
```diff
@@ -280,6 +280,22 @@ export default class DhCharacter extends BaseDataActor {
...
+                        choices: CONFIG.DH.GENERAL.diceTypes,
+                        initial: CONFIG.DH.GENERAL.diceTypes.d12,
+                        label: 'DAGGERHEART.ACTORS.Character.defaultHopeDice'
+                    }),
+                    defaultFearDice: new fields.StringField({
+                        nullable: false,
+                        required: true,
+                        choices: CONFIG.DH.GENERAL.diceTypes,
+                        initial: CONFIG.DH.GENERAL.diceTypes.d12,
+                        label: 'DAGGERHEART.ACTORS.Character.defaultFearDice'
```
CarlosFdez, line 298: Is there any reason why these can't be a number where the valid choices are 4, 6, 8, 10, 12, 20? Is there convention and consistency to consider?

CarlosFdez:
Other than the comment (which may be better to keep it as a string), this looks fine to me. 

Unrelated to the code, and more about game mechanics, but it'd be nice if the game made them step ups or downs. That cultist effect would clash with something that increases your hope die, when ideally they would neutralize.

WBHarry:
I can see a situation where that might be used, but there isn't really one right now.
I don't think you can really make a merge of two effects that each say: "Your fear dice is now d20" and "your fear dice is now d6". They're absolute effects. I think it'd be reasonable in such a situation for the GM to just decide which one will be active.

module/data/actor/character.mjs
WBHarry: Considering the only use the property will ever have is to set the `_faces` property on rolls, there isn't really any reason as such. I guess I felt a bit silly about the idea of adding a new config property for `diceFaces` when `diceTypes` already existed.

WBHarry: Changed it to be a number :+1:

PR #1523, opened 2026-01-11
[Feature] Advanced Effect Settings

WBHarry:
- Added DomainTouched. Domain cards can now declare a requirement of X cards in the loadout from the domain. If this isn't satisfied its effects are inactive and its actions cannot be used. (All <Domain>-Touched Domain Cards)
- Added VaultActive. Domain cards can now declare that their effects should be active even if they are in the vault. (Vitality Domain Card)
- Added LoadoutIgnore. Domain cards can now declare that they should not count towards the loadout limit. (Notorious Domain Card)
- Updated the compendium items with these properties.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -2295,9 +2295,8 @@
                 "type": "Type",
                 "recallCost": "Recall Cost",
                 "vaultActive": "Active In Vault",
-                "loadoutIgnore": "Doesn't Count For Loadout",
+                "loadoutIgnore": "Ignores Loadout Limits",
                 "domainTouched": "Domain Touched",
-                "domainTouchedPlaceholder": "loadout {domain} requirement",
                 "foundationTitle": "Foundation",
                 "specializationTitle": "Specialization",
                 "masteryTitle": "Mastery"
--- a/module/applications/sheets/items/domainCard.mjs
+++ b/module/applications/sheets/items/domainCard.mjs
@@ -39,9 +39,6 @@ export default class DomainCardSheet extends DHBaseItemSheet {
         const context = await super._prepareContext(options);
         context.domain = CONFIG.DH.DOMAIN.allDomains()[this.document.system.domain];
         context.domainChoices = CONFIG.DH.DOMAIN.orderedDomains();
-        context.domainTouchedPlaceholder = game.i18n.format('DAGGERHEART.ITEMS.DomainCard.domainTouchedPlaceholder', {
-            domain: game.i18n.localize(context.domain.label)
-        });
 
         return context;
     }
--- a/templates/sheets/items/domainCard/settings.hbs
+++ b/templates/sheets/items/domainCard/settings.hbs
@@ -19,7 +19,7 @@
         <span>{{localize "DAGGERHEART.ITEMS.DomainCard.loadoutIgnore"}}</span>
         {{formField systemFields.loadoutIgnore value=source.system.loadoutIgnore}}
         <span>{{localize "DAGGERHEART.ITEMS.DomainCard.domainTouched"}}</span>
-        {{formField systemFields.domainTouched value=source.system.domainTouched placeholder=domainTouchedPlaceholder }}
+        {{formField systemFields.domainTouched value=source.system.domainTouched placeholder=0 }}
     </fieldset>
 
     {{> "systems/daggerheart/templates/sheets/global/partials/resource-section/resource-section.hbs" }}


src/packs/domains/folders_Level_8_me7ywrVh38j6T8Sm.json
```diff
@@ -6,7 +6,7 @@
   "sorting": "a",
   "_id": "me7ywrVh38j6T8Sm",
   "description": "",
-  "sort": 800000,
```
Ikraik, line 9: Hmm why is the sorting affected here? was there some sort of reorganizing effort or something?

WBHarry: No, while changing a feature I just happened to manage to drag the folder so it sorted itself differently.
I redid the sorting to fix it, and the numbers just ended up different - but the ordering the same way it should be :grin:

module/applications/sheets/actors/character.mjs
```diff
@@ -822,7 +827,7 @@ export default class CharacterSheet extends DHBaseActorSheet {
     static async #toggleVault(_event, button) {
         const doc = await getDocFromElement(button);
         const { available } = this.document.system.loadoutSlot;
-        if (doc.system.inVault && !available) {
+        if (doc.system.inVault && !available && !doc.system.loadoutIgnore) {
```
CarlosFdez, line 830: just starting to review this, but this looks like what this does is allow the loadout ignore card to go into the loadout rather than it staying active despite it being vaulted?

WBHarry: Yeah. LoadoutIgnore = you're allowed to have max loadout + this card in the loadout.

This check usually throws an error and aborts when you try to bring a vaulted card into the loadout if there is no more loadout slots. I added a check so that it only aborts if `loadoutIgnore` is not set to true.

WBHarry: Only one card that works like this as is. And I didn't care to implement a block against being in the Vault. That seems less neccessary.
image

CarlosFdez: So for those cards, you can put it either in the active cards or in the vault, and its up to the player to decide where. That's fine! I wasn't aware of Notorious, clearly not notorious enough for me.

module/data/item/domainCard.mjs
```diff
@@ -38,6 +52,30 @@ export default class DHDomainCard extends BaseDataItem {
...
+
+        const matchingDomainCards = this.parent.parent.items.filter(
+            item => !item.system.inVault && item.system.domain === this.parent.system.domain
+        ).length;
+        return matchingDomainCards < this.parent.system.domainTouched;
+    }
+
+    get cannotUse() {
+        if (this.domainTouchedSuppressed) {
+            return ui.notifications.warn(
```
CarlosFdez: Given that this is a exposed getter, that could theoretically be used by modules or later features to filter or render stuff in certain ways, I think it'd be better if instead of warning it returned a boolean, and the warning was placed where attempting to use actually occurred (in `DHItem.use()`)

CarlosFdez: Though I guess "why" it cannot be used is also important, if this is expanded. Not sure. Maybe should be a method instead?

WBHarry: Hm...does changing it to a method really change much? It's still an exposed, usable thing 🤔

module/data/item/domainCard.mjs
```diff
@@ -38,6 +52,30 @@ export default class DHDomainCard extends BaseDataItem {
         return game.i18n.localize(allDomainData[this.domain].label);
     }
 
+    get vaultSupressed() {
+        return this.inVault && !this.vaultActive;
+    }
+
+    get domainTouchedSuppressed() {
```
CarlosFdez: I am uncertain on this, so if you disagree no worries, but I'm wondering if `isDomainTouchedSuppressed` and `isVaultSuppressed` would make it more obvious that these are getters and not persisted values that define behavior.

module/data/item/domainCard.mjs
CarlosFdez: Actually, since there is only a single use, and a single way you cannot use something, maybe its best to remove the helper, and check for domainTouchedSuppressed directly in the use function. It'll be easier to figure out the correct abstraction once more examples exist.

module/data/item/domainCard.mjs
WBHarry: No, `is` is a better descriptor that it's a bool :+1:
Changed.

module/data/item/domainCard.mjs
WBHarry: Yah, lets not overthink it. 
I removed it and added the check and warning in-place instead. 👍

CarlosFdez: Github didn't show me your message till I hard refreshed, sorry!

My concern is that it looks like a getter to check if it can be used. Theoretically it could be used in handlebars to hide a button or do some css or something. However, it has a side effect, so it cannot be used that way even though it looks like it could. Either moving the side effect or some sort of "checkUsable()" function would avoid that misunderstanding. But I don't think we actually need this getter, unless there was something you had in mind in the future involving it of course.

module/data/item/domainCard.mjs
CarlosFdez: Thank you. Yeah I was confused when I saw where it was invoked till I looked up the definition.

CarlosFdez:
image

CarlosFdez:
Thoughts on `Doesn't Count for Loadout` to `Ignores Loadout Limits`? Perhaps I'm being assumptive, but doesn't count for loadout sounds like some people might interpret it as "does nothing in the loadout".

WBHarry:
Hm, yeah `Ignores Loadout Limits` does sounds better 👍

CarlosFdez:
im now realizing that was a number field, rather than a bug. Since placeholders are supposed to represent a valid form (the recent material ui usage of them as labels is abberant), what about 0 as the placeholder? If that's confusing, maybe the label could be something else? `Grace Cards Required?` `Domain Touched: 0` might be fine tho

WBHarry:
Yah, I was feeling unsure about people understanding what it was for, but it's probably fine :+1:

PR #1532, opened 2026-01-14
[Feature] Beastform Token Scale

WBHarry:
Added so that beastforms can set a scale to their transform image, so it  can handle artwork with non-standard dimensions.
image

### DIFF
--- a/module/data/item/beastform.mjs
+++ b/module/data/item/beastform.mjs
@@ -49,6 +49,7 @@ export default class DHBeastform extends BaseDataItem {
                     choices: CONFIG.DH.ACTOR.tokenSize,
                     initial: CONFIG.DH.ACTOR.tokenSize.custom.id
                 }),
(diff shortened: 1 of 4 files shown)


module/data/item/beastform.mjs
```diff
@@ -184,6 +185,7 @@ export default class DHBeastform extends BaseDataItem {
                     tokenImg: this.parent.parent.prototypeToken.texture.src,
                     tokenRingImg: this.parent.parent.prototypeToken.ring.subject.texture,
                     tokenSize: {
+                        scale: this.parent.parent.prototypeToken.texture.scaleX,
```
CarlosFdez, line 188: Why is there a single scale when the source data has X and Y?

For bestiary its fine, but is this guaranteed to restore a character to their original state? I don't remember token data off the top of my head.

PR #1534, opened 2026-01-14
[Fix] Daggerheart Menu LightMode

WBHarry:
Fixed so the menu and Menu Icon coloration correctly changes with mode change.

### DIFF
--- a/styles/less/global/global.less
+++ b/styles/less/global/global.less
@@ -52,6 +52,7 @@
     }
 }
 
+/* TODO: Remove me when this issue is resolved https://github.com/foundryvtt/foundryvtt/issues/13734 */
 body.theme-dark,
 .themed.theme-dark {
     color-scheme: dark;


CarlosFdez:
The tabs.less change makes sense to me, but why are we moving from light-dark to specific theme specific overrides? Shouldn't light dark be a much more succinct way to handle that?

WBHarry:
It's because `light-dark` doesn't work for these foundry templates. The way they inject the theme classes is different from how we do it.
I haven't delved into the exactities in a long while, so I don't actually remember the technicality of it.

CarlosFdez:
`light-dark` should be based on the `color-scheme` property, with the first var picked if its light and the second var picked if its dark, if MDN is correct. Maybe that's not getting set properly?

WBHarry:
Or actually, no, so light-dark will react to the application theme flag. The issue is that this section of the UI should be tied to the interface flag, which this solves.
image

CarlosFdez:
This is a foundry bug. Looks like for light mode interfaces, the color-scheme is still dark even if its a light theme. It needs to be override for `.theme.theme-blah`

WBHarry:
Yup, the global style addition to set the color-scheme is great. We'll have to do some cleanup of ugly style solutions like this at a later time 👍 

I changed it.

CarlosFdez (approved):

PR #1538, opened 2026-01-15
[Fix] 1535 - Toggle Bonuses

### LINKED ISSUE #1535: [Bug] Powerful feature adds +1 to damage.
**Describe the bug**
The "Powerful" effect that comes from weapons, such as greatstaff tier1, has been changed to show the powerful modifier. However, there is also a powerful effect that adds +1 to damage, which should not be there.

**To Reproduce**
1. Equip a character with a greatstaff
2. Roll a weapon attack and roll for damage
3. See you have the "Powerful" modifier and the "Powerful" effect

**Expected behavior**
The "Powerful" weapon feature should only grant the powerful modifier and not the +1 to damage.

**Screenshots**

Image

**Setup Information:**
 - OS: Windows
 - Browser Firefox
 - Foundry Version v13 b351
- System Version 1.5.0

WBHarry:
Fixes #1535 
- Fixed all SRD item instances of Powerful and Massive. They had old effects on them that included incorrect damage data.

**Toggle Bonuses**
- Supressed effects were being presented for use. Suppressed is our own internal logic which is different from if the effect is inactive - IE, a weapon that's not equipped is suppressed. They're no longer presented as available.

- Weapon Effects should only ever be applied for actions done with the weapon itself. Fixed this. 
Previously you could have a broadsword (reliable) in your inventory unequipped, and when making an attack with another weapon you would see reliable as an option to use.

### DIFF
--- a/module/documents/chatMessage.mjs
+++ b/module/documents/chatMessage.mjs
@@ -160,7 +160,7 @@ export default class DhpChatMessage extends foundry.documents.ChatMessage {
         if (this.system.action) {
             const actor = await foundry.utils.fromUuid(config.source.actor);
             const item = actor?.items.get(config.source.item) ?? null;
-            config.effects = await await game.system.api.data.actions.actionsTypes.base.getEffects(actor, item);
+            config.effects = await game.system.api.data.actions.actionsTypes.base.getEffects(actor, item);
             await this.system.action.workflow.get('damage')?.execute(config, this._id, true);
         }
 


module/data/action/baseAction.mjs
```diff
@@ -198,7 +198,7 @@ export default class DHBaseAction extends ActionMixin(foundry.abstract.DataModel
         let config = this.prepareConfig(event);
         if (!config) return;
 
-        await this.addEffects(config);
+        config.effects = await DHBaseAction.getEffects(this.actor, this.item);
```
CarlosFdez: Would you not use the very long `game.system.api.data.actions.actionsTypes.base` here as well?

CarlosFdez:
> Weapon Effects should only ever be applied for actions done with the weapon itself. Fixed this.
Previously you could have a broadsword (reliable) in your inventory unequipped, and when making an attack with another weapon you would see reliable as an option to use.

How does this work for secondary weapons whose effects apply to primary attacks, such as paired and double duty?

WBHarry:
..it does not!

Going to have to add a check if the weapon is `secondary`, and still add the effect in that case. And then check the item currently being rolled for if it's a primary weapon and in that case use it 🙃

WBHarry:
Daggerheart is a simple game my ass. 😑

module/data/action/baseAction.mjs
WBHarry: Exchanged it :+1:

WBHarry:
Added a case for secondary weapons so they work :+1:

module/applications/sheets/actors/character.mjs
```diff
@@ -712,7 +712,7 @@ export default class CharacterSheet extends DHBaseActorSheet {
             headerTitle: game.i18n.format('DAGGERHEART.UI.Chat.dualityRoll.abilityCheckTitle', {
                 ability: abilityLabel
             }),
-            effects: Array.from(await this.document.allApplicableEffects()),
+            effects: await await game.system.api.data.actions.actionsTypes.base.getEffects(this.document),
```
CarlosFdez, line 715:
```suggestion
            effects: await game.system.api.data.actions.actionsTypes.base.getEffects(this.document),
```
🌞

module/data/action/baseAction.mjs
```diff
@@ -266,14 +266,27 @@ export default class DHBaseAction extends ActionMixin(foundry.abstract.DataModel
...
+                    /* Unless they're secondary - then they apply only to other primary weapons */
+                    if (effect.parent.system.secondary) {
+                        if (effectParent.type !== 'weapon' || effectParent.system.secondary) return false;
+                    } else if (effectParent?.id !== effect.parent.id) return false;
+                }
+
+                return !effect.isSuppressed;
+            });
+
+        return [];
```
CarlosFdez:
```suggestion
        if (!actor) return [];
        
        return Array.from(await actor.allApplicableEffects()).filter(effect => {
            /* Effects on weapons only ever apply for the weapon itself */
            if (effect.parent.type === 'weapon') {
                /* Unless they're secondary - then they apply only to other primary weapons */
                if (effect.parent.system.secondary) {
                    if (effectParent.type !== 'weapon' || effectParent.system.secondary) return false;
                } else if (effectParent?.id !== effect.parent.id) return false;
            }

            return !effect.isSuppressed;
        });
```

CarlosFdez: While the contents of the filter block can maybe be simplified, probably not worth doing so until the logic expands.

module/documents/chatMessage.mjs
```diff
@@ -158,7 +158,9 @@ export default class DhpChatMessage extends foundry.documents.ChatMessage {
         const config = foundry.utils.deepClone(this.system);
         config.event = event;
         if (this.system.action) {
-            await this.system.action.addEffects(config);
+            const actor = await foundry.utils.fromUuid(config.source.actor);
+            const item = actor?.items.get(config.source.item) ?? null;
+            config.effects = await await game.system.api.data.actions.actionsTypes.base.getEffects(actor, item);
```
```suggestion
            config.effects = await game.system.api.data.actions.actionsTypes.base.getEffects(actor, item);
```

WBHarry:
Good catches. Some silly business that had snuck in.

Accepted suggested code changes :+1:

PR #1542, opened 2026-01-15
[Feature] 1541 - Trigger Improvements

WBHarry:
**Improved Registration and Unregistraton**
---
**Characters** 

Always have their triggers registered as they are assumed to always be important.

**Adversaries/Environments**
- Their triggers are only registered and run if they have a token on the active scene.
- Swapping scene unregisters these triggers from that scene and registers the new triggers from that scene.

**Logging**
---
Added console output for triggers as they are run. It's accessed in the same way as foundry hooks by setting a flag in the console.
`CONFIG.debug.triggers = true`

### DIFF
--- a/module/data/registeredTriggers.mjs
+++ b/module/data/registeredTriggers.mjs
@@ -121,7 +121,7 @@ export default class RegisteredTriggers extends Map {
                     else if (triggeringActorType === 'other' && currentActor?.uuid === actorUuid) continue;
                 }
 
-                for (let command of commands) {
+                for (const command of commands) {
                     try {
                         if (CONFIG.debug.triggers) {
                             const item = await foundry.utils.fromUuid(itemUuid);


CarlosFdez:
Wouldn't environments linked in the scene config not have a token?

WBHarry:
True. Mm..I guess that could easily be handled in the two hooks I set up.

WBHarry:
I just added a little outcommented stub for Scene Environments for now. I know what to do, but it's a fair bit of stuff, and it doesn't feel immediately neccessary as we don't even include any environment triggers yet.
Made an issue to solve.
https://github.com/Foundryborne/daggerheart/issues/1543

module/data/registeredTriggers.mjs
```diff
@@ -0,0 +1,154 @@
...
+                if (!actor || !triggerActors.includes(actor.type)) continue;
+                if (tokenBoundActors.includes(actor.type) && !actor.getActiveTokens().length) continue;
+
+                const triggerData = CONFIG.DH.TRIGGER.triggers[trigger];
+                if (triggerData.usesActor && triggeringActorType !== 'any') {
+                    if (triggeringActorType === 'self' && currentActor?.uuid !== actorUuid) continue;
+                    else if (triggeringActorType === 'other' && currentActor?.uuid === actorUuid) continue;
+                }
+
+                for (let command of commands) {
```
CarlosFdez:
```suggestion
                for (const command of commands) {
```
🌞 

It'd be nice if you could find a way to lower the indentation level to remove the march of } but its not that important. The giant if block is what comes to mind as the easiest one.

CarlosFdez:
I didn't hard focus the implementation but design wise what I did notice looked to be an improvement. lgtm.

PR #1566, opened 2026-01-20
[Fix] Damage Rerolls

WBHarry:
Fixed so that damage rerolls work again

### DIFF
--- a/module/dice/damageRoll.mjs
+++ b/module/dice/damageRoll.mjs
@@ -292,9 +292,6 @@ export default class DamageRoll extends DHRoll {
                 if (isDie) diceIndex++;
                 return fixedTerm;
             }),
-            options: {
-                roll: []
-            },
             class: 'DamageRoll',
             evaluated: false
         });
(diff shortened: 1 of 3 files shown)


module/dice/damageRoll.mjs
```diff
@@ -292,6 +292,9 @@ export default class DamageRoll extends DHRoll {
                 if (isDie) diceIndex++;
                 return fixedTerm;
             }),
+            options: {
+                roll: []
+            },
```
CarlosFdez: Is there a reason we can't try to make these the default options of a DamageRoll or DhRoll? That might be better for external use unless an empty roll array means something specific.

WBHarry: The main issue is that the Config for DamageRoll and DhRoll is not defined in them. It's passed in from Actions that use them. `BaseAction` has the `prepareBaseConfig` where most things are set.
That's why cases where we use the rolls outside of actions can end up with missing things in this way 🤔

CarlosFdez: No luck even in the constructor? Using mergeObject there?

PR #1570, opened 2026-01-22
Set default data for a roll instead and fix title

CarlosFdez:
Alternative implementation, and fixes the title when you right click and reroll from there.

### DIFF
--- a/module/dice/dhRoll.mjs
+++ b/module/dice/dhRoll.mjs
@@ -3,7 +3,7 @@ import D20RollDialog from '../applications/dialogs/d20RollDialog.mjs';
 export default class DHRoll extends Roll {
     baseTerms = [];
     constructor(formula, data = {}, options = {}) {
-        super(formula, data, foundry.utils.mergeObject({ roll: [] }, options));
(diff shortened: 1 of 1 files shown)


WBHarry:
Hm..I gave this a try. I had a to change the following line to `this.options.bonusEffects = this.bonusEffectBuilder();` instea dof just `options.bonusEffects..` since the object is changed  in the super now.

However, it creates pretty big knock-on effects. Try rolling an attack, and then using the roll damage button in chat. 
It's pretty tough to debug, but I think the issue there becomes that the original config object is sent in at two different places, and it ends up missmatching..or some such, unsure. It's explosive 🤯

WBHarry:
Hm, yeah it won't work like this.
When an `Action` sends in a config object, the Action and the Roll will end up with different config objects because of the `mergeObject` creating a separate object.

CarlosFdez:
Ah, so instead it should be the other way around, with a flag set to not override. Tho that does introduce side effects, so a bit strange, but oh well.

CarlosFdez:
Try again. I didn't test it in dheart, but I tested the mergeObject() itself and the options are mutated and kept.

WBHarry (approved):
Works. Tested 🤖 :+1:

PR #1572, opened 2026-01-22
[Feature] DR Command Resources

### LINKED ISSUE #1571: [Bug] Issues with Dice Roll Automation
I used the `/dr` command as a player (non-GM). While the roll functioned, the automation failed to trigger; it did not generate Fear or Hope.

WBHarry:
Fixes #1571 
`/dr` chatcommand and buttons now grant resources via automation by default. Optionally turned off via parameter noResources=true

### DIFF
--- a/module/enrichers/DualityRollEnricher.mjs
+++ b/module/enrichers/DualityRollEnricher.mjs
@@ -34,8 +34,6 @@ function getDualityMessage(roll, flavor) {
               ? 'Disadvantage'
               : undefined;
 
-    const noResources = Boolean(roll?.noResources);
-
     const dualityElement = document.createElement('span');
     dualityElement.innerHTML = `
         <button type="button" class="duality-roll-button${roll?.inline ? ' inline' : ''}" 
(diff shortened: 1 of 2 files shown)


module/dice/dualityRoll.mjs
```diff
@@ -261,7 +261,7 @@ export default class DualityRoll extends D20Roll {
     }
 
     static async handleTriggers(roll, config) {
-        if (!config.source?.actor) return;
+        if (!config.source?.actor || config.skips?.triggers) return;
```
CarlosFdez, line 264: Seems unrelated, but why would having triggers cause you to return?

module/enrichers/DualityRollEnricher.mjs
```diff
@@ -47,6 +49,7 @@ function getDualityMessage(roll, flavor) {
             ${roll?.trait && abilities[roll.trait] ? `data-trait="${roll.trait}"` : ''}
             ${roll?.advantage ? 'data-advantage="true"' : ''}
             ${roll?.disadvantage ? 'data-disadvantage="true"' : ''}
+            data-no-resources="${noResources}"
```
CarlosFdez:
```suggestion
            ${noResources ? 'data-no-resources="true"' : ''}
```

I think it'd be better to avoid having the dataset in there at all if false, since some things often check for the presence of the data. My preference would be to convert all of this to a document.createElement("button") and then update that if possible but this is fine.

module/dice/dualityRoll.mjs
CarlosFdez: Nevermind disregard.

module/enrichers/DualityRollEnricher.mjs
WBHarry: Changed it around so it's only added as a data property if it's there 👍

WBHarry:
Eeeh..tomato tomato.
It -is- a new feature as such, as we have never had resources generated from `/dr` rolls. It might not be viewed in that light from those that desire it to be there though. Not that it matters 👅

PR #1578, opened 2026-01-24
[Fix] Multiclass Spellcast Modifier

WBHarry:
Spellcast modifiers were not being sorted correctly for use on multiclass, making it not available.


WBHarry:
It's not intended to give a choice. It picks the highest available spellcast modifier you have.

The bug I fixed was that it failed to do that if you had a multiclass of <Non-Spellcaster> and <Spellcaster> , such as Guardian into Sorcerer.

Ikraik:
ah... that's what I did I think. I went warrior into bard

Ikraik:
it works as intended then

PR #1592, opened 2026-01-27
[Fix] 1582 - Automation Additions

### LINKED ISSUE #1582: [Feature] Toggle for Automatic SCAR Application
**Description**
Hi 

I’m writing to suggest a small improvement. Would it be possible to add an option in the Settings menu to disable the automation that automatically adds a SCAR?

The reason I'm asking is that I use a specific homebrew setup that doesn't always interact well with this feature (it might not affect 'Hope' as intended). Having a simple toggle to turn this automation on or off would give users like me much more flexibility.

Thank you very much for your time and for the great work!

Best regards,

**Screenshots**
Image

WBHarry:
Fixes #1582 
- `/dr` commands now follow fear/hope automation unless `grantResources=false` is specifically added.
- Added automation settings to turn off any automatic effects that happen because of death moves.

### DIFF
--- a/module/data/settings/Automation.mjs
+++ b/module/data/settings/Automation.mjs
@@ -55,23 +55,21 @@ export default class DhAutomation extends foundry.abstract.DataModel {
                 initial: true,
                 label: 'DAGGERHEART.SETTINGS.Automation.FIELDS.resourceScrollTexts.label'
             }),
-            deathMoves: new fields.SchemaField({
-                automation: new fields.SchemaField({
-                    avoidDeath: new fields.BooleanField({
-                        required: true,
-                        initial: true,
-                        label: 'DAGGERHEART.CONFIG.DeathMoves.avoidDeath.name'
-                    }),
-                    riskItAll: new fields.BooleanField({
-                        required: true,
-                        initial: true,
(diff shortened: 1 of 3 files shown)


module/data/settings/Automation.mjs
```diff
@@ -55,6 +55,25 @@ export default class DhAutomation extends foundry.abstract.DataModel {
                 initial: true,
                 label: 'DAGGERHEART.SETTINGS.Automation.FIELDS.resourceScrollTexts.label'
             }),
+            deathMoves: new fields.SchemaField({
+                automation: new fields.SchemaField({
```
CarlosFdez: Why the extra nested level if the entire data model is for automation?

WBHarry: We could flatten it out. 
I was thinking it doesn't seem unreasonable that we could end up with more settings related to DeathMoves at some point, and to save on a migration 🙃

WBHarry: I went ahead and flattened it

CarlosFdez: Alrighty. I guess if that was the concern, something more descriptive like types would have been fine.

CarlosFdez (approved):

PR #1609, opened 2026-01-30
[Fix] 1601 - Template Enrichment

### LINKED ISSUE #1601: [Bug] Template enrichment bugged?
I was testing the <https://github.com/Foundryborne/daggerheart/wiki/Dice-Rolls-&-Chat-Commands#template> and found some problems.

### Case-sensitive
If you use:
@Template[type:circle|range:Melee]
It will no work. You need to use: @Template[type:circle|range:melee]

### VeryFar
VeryFar don't work. I think this one is just remove from docs.
This: @Template[type:circle|range:veryfar] or veryFar or VeryFar don't work.

### rect is weird
Image

WBHarry:
Fixes #1601 
Fixed so param.range is compared as lowerCase

### DIFF
--- a/module/enrichers/TemplateEnricher.mjs
+++ b/module/enrichers/TemplateEnricher.mjs
@@ -4,10 +4,11 @@ export default function DhTemplateEnricher(match, _options) {
     const params = parseInlineParams(match[1]);
     const { type, angle = CONFIG.MeasuredTemplate.defaults.angle, inline = false } = params;
     const direction = Number(params.direction) || 0;
+    params.range = params.range?.toLowerCase();
     const range =
         params.range && Number.isNaN(Number(params.range))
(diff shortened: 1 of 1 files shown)


CarlosFdez:
Implementation wise It looks good to me but this looks like an issue with documentation. The range is a value in the object. It doesn't support other languages nor does it support `Very Close` with a space in the middle. Case insensitivity seems to be part of a larger issue with its use.

WBHarry:
Mm..we could do something like erasing any whitespace in input params, although I don't know if it's needed.
At some point, you just have to write it in the expected format. 🤔

CarlosFdez:
Maybe `params.range = params.range?.toLowerCase();` might be cleaner I guess.

WBHarry:
Added the lowercase coercion of `params.range` :+1:

PR #1622, opened 2026-02-01
[Fix] ActionField UUID for Actor Actions

WBHarry:
Fixed so that actionFields correctly grab a uuid for actions that are directly on an actor (Standard Attacks)

### DIFF
--- a/module/data/fields/actionField.mjs
+++ b/module/data/fields/actionField.mjs
@@ -152,14 +152,9 @@ export function ActionMixin(Base) {
         }
 
         get uuid() {
-            if (
-                !(
-                    this.item instanceof game.system.api.documents.DHItem ||
-                    this.item instanceof game.system.api.documents.DhpActor
-                )
-            )
-                return null;
-            return `${this.item.uuid}.${this.documentName}.${this.id}`;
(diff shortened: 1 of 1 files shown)


CarlosFdez (commented):
Not logical or?

WBHarry:
Huh, I dropped an `|`. Interesting how it still worked 🤔

module/data/fields/actionField.mjs
```diff
@@ -152,7 +152,13 @@ export function ActionMixin(Base) {
...
         get uuid() {
-            if (!(this.item instanceof game.system.api.documents.DHItem)) return null;
+            if (
+                !(
+                    this.item instanceof game.system.api.documents.DHItem ||
+                    this.item instanceof game.system.api.documents.DhpActor
+                )
+            )
+                return null;
             return `${this.item.uuid}.${this.documentName}.${this.id}`;
```
CarlosFdez:
```suggestion
            const isItem = this.item instanceof game.system.api.documents.DHItem;
            const isActor = this.item instanceof game.system.api.documents.DhpActor;
            return isItem || isActor ? `${this.item.uuid}.${this.documentName}.${this.id}` : null;
```

PR #1626, opened 2026-02-01
[Fix] RollMessage Order

WBHarry:
Fixed so that the description message always comes first with the action workflow


CarlosFdez:
If possible I'd like @moliloo to test this. I think for domain card abilities like enrapture this may cause difficulty.

moliloo:
> If possible I'd like @moliloo to test this. I think for domain card abilities like enrapture this may cause difficulty.

Sry For the late reply, i will teste it tonight!

moliloo:
@WBHarry @CarlosFdez Its good to me

PR #1636, opened 2026-02-04
[Fix] 1633 - ActiveEffect Autocomplete

### LINKED ISSUE #1633: [Bug] Features with HP Reduce result in Sheet being locked
**Describe the bug**
When adding a Feature that targets 'Character Hit Points', the PC sheet is caught in an error, preventing it from being opened or otherwise interacted with.

**To Reproduce**
1. Create a Character actor.
2. Create a Feature affecting the Character's HP, specifically `system.resources.hitPoints`.
3. Add Feature to Character

**Expected behavior**
After a brief chat in the bugs channel on Discord, Dev Harry mentioned that we should not offer this attribute option in this context, since we are likely creating a loop. Additionally, the Max HP attribute is only being shown to Adversaries, which is misleading, since the path is the same (which lead me down this path to begin with). Thanks Harry for the quick support!

**Screenshots**

Image

Link to the Video in the Discord Channel (unable to upload here)
https://discord.com/channels/1374793429191491725/1403004936127512668/1468547056292335637

**Setup Information:**
 - OS: Windows
 - Browser: Chrome
 - Foundry Version: v13 b351
 - System Version: 1.6.4

WBHarry:
Fixes #1633 
- Improved the autocomplete typing experience.
- Made the ActiveEffect autocomplete display max values for resources
- Made the resources bars in tokenConfig display bars resources correctly.

### DIFF
--- a/daggerheart.mjs
+++ b/daggerheart.mjs
@@ -242,6 +242,38 @@ Hooks.on('setup', () => {
             systemEffect: true
         }))
     ];
+
+    const actorCommon = {
+        bar: ['resources.stress'],
+        value: []
+    };
+    const damageThresholds = ['damageThresholds.major', 'damageThresholds.severe'];
+    const traits = Object.keys(game.system.api.data.actors.DhCharacter.schema.fields.traits.fields).map(
+        trait => `traits.${trait}.value`
+    );
+    CONFIG.Actor.trackableAttributes = {
+        character: {
+            bar: [...actorCommon.bar, 'resources.hitPoints', 'resources.hope'],
+            value: [
+                ...actorCommon.value,
+                ...traits,
+                ...damageThresholds,
+                'proficiency',
+                'evasion',
+                'armorScore',
+                'scars',
+                'levelData.level.current'
+            ]
+        },
+        adversary: {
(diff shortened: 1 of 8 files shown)


CarlosFdez:
For that ugly bit, what exactly are you trying to solve, so I can repro it and see if I can come up with an alternative?

WBHarry:
If you check it in main, you'll find that the autocomplete list gives you resources with bad string paths, like:
Hit Points will give you --> `resources.hitPoints`
Stress will give you --> `resources.stress`
 Users trust clicking on these, and bringing in an active effect with those bricks a charcer since we're overwriting a schema with a number 2 or some such.
 
 ---
 What's desired is that this autocomplete should get `resources.<resource>.max`, and nothing else from resources.
 Meanwhile, the TokenConfig Resources tab where you can set up your resource bars need to get `resources.<resource>` (what we're currently getting and don't want in the autocomplete)
 
 ---
 Aside from managing to get the resources with actual schema defined max values (hit points and stress), there is a further complication in that `hope` does not have `max` defined in the schema, and is therefore not recognised as a `bar` value.
It's set up this way due to scars wanting to modify the maximum of hope if I recall.

PR #1639, opened 2026-02-05
[Fix] Gridless Hover Distance

WBHarry:
Made gridless distances lean towards being in the lower range by applying `Math.floor` to the calculated distance.
We could make it just `Math.round` to give it just a little bit of wiggleroom - but I think it makes for a better experience to give the wider interpretation of being in range to make melee a less annoying distance to get into.

### DIFF
--- a/module/canvas/placeables/token.mjs
+++ b/module/canvas/placeables/token.mjs
@@ -1,4 +1,4 @@
-import DhMeasuredTemplate from "./measuredTemplate.mjs";
+import DhMeasuredTemplate from './measuredTemplate.mjs';
 
 export default class DhTokenPlaceable extends foundry.canvas.placeables.Token {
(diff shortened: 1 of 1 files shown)


WBHarry:
Melee can still be weird the closer you get to being on the diagonal - but I don't think we should care too much, really. Gridless is the mode you use when you don't care about distances typically 🤔

CarlosFdez:
We could give this a shot, though it will not work as well for > 5 ft squares, since it relies on rounding down to a value. Looks like the workaround I did for diagonal didn't work. We can give this a shot.

If this doesn't work, we can round down to multiples of grid spaces (or quarter grid spaces?) and see how that behaves later.

CarlosFdez (approved):

PR #1665, opened 2026-02-15
[Feature] 1663 - Token ActiveEffect Scaling

### LINKED ISSUE #1663: [Feature] Beastform transformation
**Description**
When using the Beastform feature and selecting a form, the character gets an effect called "Beastform transformation".
That effect also appears on the upper left corner of the token.

When a form has a tinya size, the effect appears with normal size, taking a lot of space. 
Could something be done to resize the effect icon as well if a token decreases or increases its size? 

I am guessing this also affects other effects placed on tokens, not only Beastforms

Image

Image

WBHarry:
Fixes #1663 
Changed so that active effect images on a canvas token have their size altered in accordance with token size.
Unsure if this is a good thing, but the effects are like this:

**Previously**
image

**After Change**
image

### DIFF
--- a/module/canvas/placeables/token.mjs
+++ b/module/canvas/placeables/token.mjs
@@ -44,6 +44,35 @@ export default class DhTokenPlaceable extends foundry.canvas.placeables.Token {
         this.renderFlags.set({ refreshEffects: true });
     }
 
+    _refreshEffects() {
+        const s = canvas.dimensions.uiScale;
+        let i = 0;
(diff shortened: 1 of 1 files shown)


Ikraik:
Hmm, I suggest we explore more solutions but let's keep the PR active till other solutions are proposed and tested.

Ikraik:
@WBHarry suggestion is that we keep this on draft PR like we did for Sidebar. I don't think this will be approved soon.

PR #1667, opened 2026-02-17
[V14] Effect Stacking

WBHarry:
Added stacking functionality for effects that is enabled in the ActionEffect config. Effects can reference `@stacks` as a value for use in their `changes`.
image
image

**Display**
The stack value is shown if stacking is enabled, and if the current stacks are more than 1.
The value is shown in the top-right EffectsDisplay, and on the effect icon on the canvas token.
image

**Effect Creation**
My thinking is that when an active effect is attempted to be applied to an actor:
- If the effect does **not** already exist on the actor, it is applied. (Checked via `effect.origin`)
- If the effect is not set up to stack and it already exists, then it's not applied and creation is aborted.
- If the effect is set up to stack and it already exists, then we increase the stack count of the existing effect and abort creating a new one.

### DIFF
--- a/module/documents/actor.mjs
+++ b/module/documents/actor.mjs
@@ -122,6 +122,14 @@ export default class DhpActor extends Actor {
         }
     }
 
+    _onUpdateDescendantDocuments(parent, collection, documents, changes, options, userId) {
+        if (collection === 'effects') {
+            ui.effectsDisplay.render();
+        }
+
+        super._onUpdateDescendantDocuments(parent, collection, documents, changes, options, userId);
+    }
+
     async updateLevel(newLevel) {
         if (!['character', 'companion'].includes(this.type) || newLevel === this.system.levelData.level.changed) return;
 


WBHarry:
Added stacking on the following Compendium Items:
**No Mercy** (Warrior Feature)
**Rogue's Dodge** (Rogue Feature)
**Rage Up** (Blade Domain Card)
**Corrosive Projectile** (Sage Domain Card)

Ikraik:
Styling felt weird on stacking where the container was uneven at times. maybe no padding on top left for the number container? Might need molilo's input but it could be a personal preference situation for me.

CarlosFdez:
Few problems off the bat:

- Deleting when stacking to 0 might be unsafe. Depending on where the effect came from, it might be good to disable when hitting 0 instead, treating enabling as "setting to 1". Perhaps its possible to distinguish between temporary effects and stuff tied to domain cards and such?
- Effects do not rerender on actor and actor item update. Doing so might also remove the need for all the render calls.

WBHarry:
Hm, yeah. Agreed that we should not be deleting non-temporary effects when going to 0 stacks. 
There -is- a case for not doing it on temporary either - but in that case it'll get very annoying to actually get rid of the effect, since the most common usecase will probably be that 0 stacks means that it's over with.

WBHarry:
Ah, right. Actually, both the Effects Display and the canvas token effects do not show permanent effects to begin with, so it's a non issue.
It'd be meaningless after all to show a permanent icon on token/effectDisplay for something you -always- have.

CarlosFdez:
Ah, if it doesn't show non-temporary, then its fine. Then its just the update issue. Adding something to the actor's _onUpdateDescendantDocuments() (and checking that the collection is active effects and its the actor being viewed) might do it?

module/documents/activeEffect.mjs
```diff
@@ -108,6 +108,20 @@ export default class DhActiveEffect extends foundry.documents.ActiveEffect {
             update.img = 'icons/magic/life/heart-cross-blue.webp';
         }
 
+        const existingEffect = this.actor.effects.find(x => x.origin === data.origin);
```
CarlosFdez, line 111: What happens if two different effects have the same origin? Shouldn't this test a bit more? Or am I misunderstanding origin?

module/documents/activeEffect.mjs
```diff
@@ -108,6 +108,20 @@ export default class DhActiveEffect extends foundry.documents.ActiveEffect {
...
+        const stacks = data.system?.stacking?.enabled;
+        if (existingEffect && !stacks) return false;
+
+        if (existingEffect && stacks) {
+            const incrementedValue = existingEffect.system.stacking.value + 1;
+            await existingEffect.update({
+                'system.stacking.value': existingEffect.system.stacking.max
+                    ? Math.min(incrementedValue, existingEffect.system.stacking.max)
+                    : incrementedValue
+            });
```
CarlosFdez, line 119:
```suggestion
            await existingEffect.update({
                'system.stacking.value': Math.min(incrementedValue, existingEffect.system.stacking.max ?? Infinity)
            });
```

module/documents/activeEffect.mjs
WBHarry: Origin would be the original effect. IE, you have the warriors No Mercy feature in the SRD Compendium, when trying to apply your No Mercy a second time, it would say that `existingEffect=true` because there exists an effect on the actor with the same origin (`compendium.<id>.Item.<Id>.<something.something>`).

I don't think that check in particular can go wrong - but I guess we should maybe do some other check aswell to try and establish equality if the origin is missing 🤔

module/documents/activeEffect.mjs
```diff
@@ -184,10 +198,17 @@ export default class DhActiveEffect extends foundry.documents.ActiveEffect {
             } catch (_) {}
         }
 
-        const evalValue = this.effectSafeEval(itemAbleRollParse(key, parseModel, effect.parent));
+        const evalValue = this.effectSafeEval(this.effectRollParse(key, parseModel, effect.parent, effect));
         return evalValue ?? key;
     }
 
+    static effectRollParse(value, actor, item, effect) {
```
CarlosFdez: Maybe we can do without? This looks simple enough to keep in getChangeValue().

module/data/activeEffect/baseEffect.mjs
```diff
@@ -80,6 +80,17 @@ export default class BaseEffect extends foundry.data.ActiveEffectTypeDataModel {
                     initial: CONFIG.DH.GENERAL.range.melee.id,
                     label: 'DAGGERHEART.GENERAL.range'
                 })
+            }),
+            stacking: new fields.SchemaField({
```
CarlosFdez: this is a bit of a deviation and may be tricky, but if this field was nullable and omitted enabled, how much trickier would all of this be to implement?

WBHarry: Hm...probably not much. I can give it a shot.
It would read a bit cleaner to see that `stacking = undefined` in the data object rather than being there and being enabled=false. Enabled false does somewhat imply that it might get enabled later, which isn't the intention.

module/documents/activeEffect.mjs
CarlosFdez: Ah, I assumed origin was the item or actor it came from, rather than some sort of source id. Alrighty then!

module/documents/activeEffect.mjs
WBHarry: :+1:

module/data/activeEffect/baseEffect.mjs
WBHarry: I changed it around to that. Works :+1:

WBHarry:
> Ah, if it doesn't show non-temporary, then its fine. Then its just the update issue. Adding something to the actor's _onUpdateDescendantDocuments() (and checking that the collection is active effects and its the actor being viewed) might do it?

Worked like a charm.
```js
_onUpdateDescendantDocuments(parent, collection, documents, changes, options, userId) {
    if (collection === 'effects') {
        ui.effectsDisplay.render();
    }

    super._onUpdateDescendantDocuments(parent, collection, documents, changes, options, userId);
}
```

PR #1674, opened 2026-02-21
Conditional Effects

WBHarry:
image


WBHarry:
It's currently added to:
**Adrenaline** (Rogue Subclass Feature)
**Rise To The Challenge** (Warrior Subclass feature)
**Conjure Shield** (Wizard Subclass Feature)
**On The Brink** (Bone Domain Card)

Ikraik:
hmm I'm gonna test if Bard Rally can be unified into a single feature now with CAE. Will let you know if I faced any difficulty

WBHarry:
CAE isn't really related to that. It's for turning on/off an effect on an actor.

The thing with Bard Rally is that we currently have two effects because the math on the `changes` part of the Effect hasn't been possible to reconsile into a single one. CAE doesn't help with that.

Ikraik:
image

Something to resolve?

Ikraik:
image

Dunno why but the styling is off.. plus on status do we plan to modify that to reflect any custom statuses? Or is that gonna be on per module basis?

WBHarry:
> Something to resolve?

Yup, fixed. :+1:

WBHarry:
> Dunno why but the styling is off.. plus on status do we plan to modify that to reflect any custom statuses? Or is that gonna be on per module basis?

Improved the styling
image

Regarding custom statuses, that should work without a hitch. We're just making use of foundry's condition list. Adding custom stuff is done to that, so we'll automatically have it.

Ikraik:
image

Not sure why but the toggle should push flexible and paired to active effects if im not wrong, or did we change the logic somehow?

CarlosFdez:
I think the way the hiding and showing of elements is done is going to cause us trouble in the future. I have two ideas we can try:

- Set some data elements on the conditional-container, that controls the visibility of sub-elements. Add a new css class (conditional-conditional-field is a mouthfull, maybe something else?) on the elements to default to hidden unless certain things are set.
- Create a partial for the conditional, and re-render and replace the segment after any change.

Ikraik:
Conflicts to be resolved.

Issue #1631, opened 2026-02-03
[Bug] Hover distance


roguedevjake:
Token distance on hover doesn't include the height of the token. e.g. a token next to another but 200ft. in the air will read as "melee"

PR #1675, opened 2026-02-22
[Fix] Support elevation in token distance hovering and fix error when overlapping

CarlosFdez:
Closes https://github.com/Foundryborne/daggerheart/issues/1631 and https://github.com/Foundryborne/daggerheart/issues/1679. This uses the configured grid setting.

This assumes that the token height is the greater of the width distance and height distance. In v14 I believe this will be an actual setting on the token.

### DIFF
--- a/daggerheart.mjs
+++ b/daggerheart.mjs
@@ -420,10 +420,7 @@ const updateActorsRangeDependentEffects = async token => {
             // Get required distance and special case 5 feet to test adjacency
             const required = rangeMeasurement[range];
             const reverse = type === CONFIG.DH.GENERAL.rangeInclusion.outsideRange.id;
-            const inRange =
-                required === 5
-                    ? userTarget.isAdjacentWith(token.object)
-                    : userTarget.distanceTo(token.object) <= required;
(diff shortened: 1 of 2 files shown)


WBHarry:
Awesum that the calculation takes the adversary height into consideration. Yeah, v14 has height consideraiton in core itself.

Non-technically speaking, the height distance seems off by 5 for most core measurement settings. 
`equidistant` and `approximate` both show the distance as `melee` when adjacent to another token that is 10 feet in the air, which doesn't seem right. Running it as `alternating` makes the distance show as `Very Close`.
image

I believe you were talking about the specificities of this in Discord, but yeah I don't feel great about the default foundry setting leading to an unexpected result 🤔

WBHarry:
Or I suppose this happens because the token on the ground has a height of 5 which means they are in reality just 5 diagional:ish feet apart from top to bottom between tokens so to speak. So probably a feature rather than an issue 🙃

CarlosFdez:
This PR now fixes three different things but they all stem from the same function so I can't really split them.

PR #1677, opened 2026-02-22
[Fix] 1671 - Compendium Context Menues

### LINKED ISSUE #1671: [Bug] Items in compendiums cannot have their Actions/Effects deleted
**Describe the bug**
While editing an item in a compendium, added Actions / Effects cannot be deleted from the item.

The “Delete” option is missing entirely along with “Use Item” / “En-/Disable Effect”.

**To Reproduce**
1. Create an item in a compendium (the screenshots show a feature, but it is a general issue)
2. Add an Action
3. Right click or use the kebab menu
4. Observe no “Delete” option
5. Add an Effect
6. Right click or use the kebab menu
7. Observe no “Delete” option

**Expected behavior**
I would expect full features on the item sheet in compendiums, including the option to delete Actions and Effects.

**Screenshots**
Compendium:

Image

Image

World sidebar:

Image

Image

WBHarry:
Fixes #1671 
Since context menu `condition` deciding wether to show a contextmenu option is `sync`, it cannot accurately decide whether to show certain context menu options based on `fromUUIDSync`.

I added two `data-` properties to the DOM that can be checked on instead to give the accurate options.

### DIFF
--- a/module/applications/sheets/api/application-mixin.mjs
+++ b/module/applications/sheets/api/application-mixin.mjs
@@ -433,7 +433,7 @@ export default function DHApplicationMixin(Base) {
                     icon: 'fa-solid fa-lightbulb',
                     condition: element => {
                         const target = element.closest('[data-item-uuid]');
-                        return !target.dataset.disabled && target.dataset.subType !== 'beastform';
+                        return !target.dataset.disabled && target.dataset.itemType !== 'beastform';
                     },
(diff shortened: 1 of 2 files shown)


CarlosFdez (approved):
Looks good to me, though data-item-type instead of data-sub-type might be more clear to a reader maybe if you wanna try that? If you disagree though keep what you have.

WBHarry:
No, I think that's better. Changed it :+1:

PR #1678, opened 2026-02-22
[Fix] 1676 - Horde Damage Fix

### LINKED ISSUE #1676: [Bug] Tangle Bramble Swarm Crush Action has wrong damage formula
The Crush action should deal 2d6+8 damage, not 1d6 damage. EDIT: Issue seems to specifically affect tokens of Tangle Bramble Swarms that were spawned using the Tangle Bramble (minion) reaction "Drain and Multiply".

Image

Ver. 1.7.1

WBHarry:
Fixes #1676
Fixed so that horde damage reduction is only applied to the standard attack.
Went ahead and added a boolean flag on Adversary standard attacks used for this. It's probably useful in other situations aswell to distinguish it from any other attacks.

### DIFF
--- a/module/data/actor/adversary.mjs
+++ b/module/data/actor/adversary.mjs
@@ -86,8 +86,7 @@ export default class DhpAdversary extends DhCreature {
                         amount: 1
                     },
                     roll: {
-                        type: 'attack',
-                        isStandardAttack: true
+                        type: 'attack'
                     },
                     damage: {
                         parts: [


CarlosFdez:
is there any reason the boolean can't be added in data preparation? That might be safer overall.

WBHarry:
Hm...yeah, true. It's not something that should ever be edited or changed, so it's reasonable to just prepare it as true on the standard attacks.

PR #1685, opened 2026-02-25
[Feature] Damage Iterrable Rework

WBHarry:


Ikraik:
This is for v.1.7.3 feature release right?

WBHarry:
Well, it's for whenever it gets in, really. 🙃

CarlosFdez:
Let me look at this more carefully soon, I've been busy but I can prioritize this.

module/data/fields/action/damageField.mjs
```diff
@@ -286,6 +287,7 @@ export class DHDamageData extends DHResourceData {
         return {
             ...super.defineSchema(),
             base: new fields.BooleanField({ initial: false, readonly: true, label: 'Base' }),
+            // direct: new fields.BooleanField({ initial: false, label: 'DAGGERHEART.CONFIG.DamageType.direct.name' }),
```
CarlosFdez:
```suggestion
```

Looks like you considered flattening the structure one level, though I guess if you've abandoned that might as well remove the leftovers.

WBHarry: Removed :+1:

WBHarry:
I can't seem to solve the issue of `foundry.utils.isEmpty` not recognising `IterableObject` as an object. Tried a bunch of assigning and methods, but wouldn't work out.

I can solve it by adding a getter/function `isEmpty` to IterableObject and using that instead, but it won't have Foundry generally recognising `IterableObject` as an object for any other potential purposes. This may not be an issue though 🤷 

---

If you want to look at it, a good example to work of is `Demon Of Jealousy` and it's `What's Yours Is Mine` feature. Using the attack action there will fail to make the roll message with targets and everything because `baseAction.hasDamage` uses `foundry.utils.isEmpty`

PR #1691, opened 2026-02-28
[Fix] 1687 - Item Transfer Without Permissions

### LINKED ISSUE #1687: [Bug] Party actor items disappearing
**Describe the bug**
If you set the party actor permissions for players as "Observer" and they try to move an item from their sheet to the Party Actor sheet, they get an error about permissions but the item is lost and disappears. 

**To Reproduce**
1. Create a party actor and put a character in there
2. Make the party actor ownership permissions as observer
3. Try, as the character, to move an item to the party actor
4. See error and item disappearing

**Expected behavior**
Items should not disappear.

**Setup Information:**
 - OS: Windows
 - Browser Firefox
 - Foundry Version v13 b351
- System Version  v.1.7.2

WBHarry:
Fixes #1687 
Fixed so that transfering items completly aborts if lacking permissions. 

We could potentially make this work via socket delegating to any active GM if we -do- want to allow players to always toss on items even if they only have observer permission - but I don't think that's a good idea.

### DIFF
--- a/module/applications/sheets/api/base-actor.mjs
+++ b/module/applications/sheets/api/base-actor.mjs
@@ -294,13 +294,14 @@ export default class DHBaseActorSheet extends DHApplicationMixin(ActorSheetV2) {
 
         /* Handling transfer of inventoryItems */
         if (item.system.metadata.isInventoryItem) {
-            if (!this.document.testUserPermission(game.user, 'OWNER', { exact: true }))
+            if (!this.document.testUserPermission(game.user, 'OWNER', { exact: true })) {
                 return ui.notifications.error(
                     game.i18n.format('DAGGERHEART.UI.Notifications.lackingItemTransferPermission', {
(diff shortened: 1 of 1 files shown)


CarlosFdez:
Delegate to the GM is how the party actor and shops works in pf2e tho. But we decided that party should be owned by players and we don't have a shop actor.

CarlosFdez (approved):
Code looks fine to me. Personally I would prefer multi-line if statements to use curly braces, but that's not part of an established style guide.

It looks like you also moved deleting to happen after creation, and I don't see any issues that would arise from that. Perhaps it might even prevent some.

WBHarry:
I agree it's neater with brackets, changed it. Not sure why, but I've been using more non-bracket ifs lately.

I think if we want to look at more GM delegation, we can do that as a separate task at some point, as I imagine there may be quite a few places in that case.

PR #1707, opened 2026-03-05
[Fix] 1696 - Homebrew Fixes

### LINKED ISSUE #1694: [Feature] Saving new homebrew item feature should only require 1 click
Currently, if you add a new item feature under the homebrew menu, you need to click "save" twice: once on the feature window and once on the homebrew setting window. If either of these are missed (by clicking X to close the window instead), the new item feature is not saved and disappears, which makes it very easy to accidentally lose progress.

Image

### LINKED ISSUE #1695: [Bug] Adding an effect to a homebrew item feature deletes previous unsaved edits
1. Create a new item feature in the homebrew menu
2. Change the name of the feature, icon, description, etc.
3. Before saving changes, navigate to the effects tab and add a new effect
4. The feature name, icon, and description will be reset to their defaults

### LINKED ISSUE #1696: [Bug] Attempting to edit a new effect in the homebrew item features menu causes error message
Attempting to edit "New Effect" created when making a new effect in the homebrew item features menu has no effect and sends this error to the console:

Image

WBHarry:
Fixes #1694
Think it's a good idea. Changed it so that saving an ItemFeature makes it directly update the game setting so it's -actually- saved.

Fixes #1695
Since creating effects neccesitates updating the game setting, it will always lose any unsaved changes. So I changed it so that changes are directly saved to the game setting aswell. I don't really see the harm.

Fixes #1696 
SettingsActiveEffectConfig was out of date, making it error.
Moved the logic into a static method on ActiveEffectConfig so that both will always use the same logic to get their changeChoices for the autocomplete.

### DIFF
--- a/module/applications/sheets-configs/activeEffectConfig.mjs
+++ b/module/applications/sheets-configs/activeEffectConfig.mjs
@@ -35,6 +35,10 @@ export default class DhActiveEffectConfig extends foundry.applications.sheets.Ac
         }
     };
 
+    /**
+     * Get ChangeChoices for the changes autocomplete. Static for use in this class aswell as in settings-active-effect-config.mjs
+     * @returns {ChangeChoice { value: string, label: string, hint: string, group: string }[]}
+     */
     static getChangeChoices() {
         const ignoredActorKeys = ['config', 'DhEnvironment', 'DhParty'];
 
--- a/module/applications/sheets-configs/setting-feature-config.mjs
+++ b/module/applications/sheets-configs/setting-feature-config.mjs
@@ -73,11 +73,10 @@ export default class SettingFeatureConfig extends HandlebarsApplicationMixin(App
         return context;
     }
 
-    static async updateData(event, element, formData) {
+    static async updateData(_event, _element, formData) {
         const data = foundry.utils.expandObject(formData.object);
-        foundry.utils.mergeObject(this.move, data);
-        await this.settings.updateSource({
-            [`${this.movePath}`]: this.move
+        await this.updateMove({
+            [`${this.movePath}`]: data
         });
 
         this.render();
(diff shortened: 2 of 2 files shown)


module/applications/sheets-configs/activeEffectConfig.mjs
```diff
@@ -4,6 +4,38 @@ export default class DhActiveEffectConfig extends foundry.applications.sheets.Ac
...
+                { id: 'details', icon: 'fa-solid fa-book' },
+                { id: 'settings', icon: 'fa-solid fa-bars', label: 'DAGGERHEART.GENERAL.Tabs.settings' },
+                { id: 'changes', icon: 'fa-solid fa-gears' }
+            ],
+            initial: 'details',
+            labelPrefix: 'EFFECT.TABS'
+        }
+    };
+
+    static getChangeChoices() {
```
CarlosFdez, line 42: I think it'd be handy to add a comment to this so that its meaning is clear. I can see us easily forgetting what this does.

module/applications/sheets-configs/setting-feature-config.mjs
```diff
@@ -76,6 +76,9 @@ export default class SettingFeatureConfig extends HandlebarsApplicationMixin(App
     static async updateData(event, element, formData) {
         const data = foundry.utils.expandObject(formData.object);
         foundry.utils.mergeObject(this.move, data);
+        await this.settings.updateSource({
+            [`${this.movePath}`]: this.move
+        });
```
CarlosFdez, line 80: Out of curiousity (and not because there's anything wrong here) which of the issues is this particular line trying to solve?

CarlosFdez:
All of those issues make sense to me.

module/applications/sheets-configs/activeEffectConfig.mjs
WBHarry: Added a comment:
```
/**
 * Get ChangeChoices for the changes autocomplete. Static for use in this class aswell as in settings-active-effect-config.mjs
 * @returns {ChangeChoice { value: string, label: string, hint: string, group: string }[]}
 */
```

module/applications/sheets-configs/setting-feature-config.mjs
WBHarry: Yeah, this was a bit confusing.

I went through and replaced all updates to run through a single method now:
```
async updateMove(update) {
    await this.settings.updateSource(update);
    this.move = foundry.utils.getProperty(this.settings, this.movePath);
}
```

PR #1711, opened 2026-03-07
[Fix] Hit/Miss Display

WBHarry:
Added an automation setting that decides if players should get to see information in chat that would otherwise demand Observer permission - such as if an attack hit or missed an Adversary.

### DIFF
--- a/system.json
+++ b/system.json
@@ -2,7 +2,7 @@
     "id": "daggerheart",
     "title": "Daggerheart",
     "description": "An unofficial implementation of the Daggerheart system",
-    "version": "1.8.0",
+    "version": "1.9.0",
     "compatibility": {
         "minimum": "13.346",
         "verified": "13.351",


templates/ui/chat/parts/target-part.hbs
```diff
@@ -1,6 +1,6 @@
 <div class="roll-part target-section dice-roll" data-action="expandRoll">
     <div class="roll-part-header"><div><span>{{pluralize currentTargets.length "DAGGERHEART.GENERAL.Target"}}</span></div></div>
-    {{#if isGM}}
+    {{#if (or isGM (not automationSettings.hideObserverPermissionInChat))}}
```
CarlosFdez: Is this re-rendered per player?

WBHarry: It's currently not rerendered. Needs a reload.

CarlosFdez: I mean is it a problem that perms are checked in the template on a chat thing. Wouldn't visibility only be based on whomever is posting it, not viewing it?

Some rolls in foundry render dynamically per player, but I don't know if this is that.

WBHarry: The issue at hand is that some GM's (I don't understand this), do not want the hit/miss info to -ever- be shown when players are attacking Adversaries. Instead they want to verbally say that it's a hit or miss every time.
Normally I might ignore this - but since it's been like that up until now, I feel less good about just removing it and always showing.

CarlosFdez: That doesn't preclude gms seeing if it hit/miss though. This may be something that needs to be done per player javascript side. But I haven't followed how this is rendered so I asked. I'll check it later I guess.

WBHarry: Ah, I get what you're saying now. Yeah, I'm pretty sure this is rendered dynamically per player since `isGM` is available (and used for this - so GMs always see)
I'll check

CarlosFdez: Most messages are not rendered dynamically per player. But roll messages in foundry often are if the content is blank. Depends on how this one is done.

WBHarry: Yeah, it's definately dynamic in our roll messages. GM ends up seeing it and player does not 👍

CarlosFdez:
This may be cross project habits, but this seems to be a metagame setting, not standard automation. Its sortof like a world appearance setting. (Edit: its kinda similar to the spotlight queue, though still a bit different).

It might be fine if its a new tab in the automation settings, is there anything else that is more focused on showing/hiding things that GMs may or may not want players to know that we can put alongside it?

Where it is is *fine*, and we may not have a better place since there's no other metagame setting that I can find, but that's probably why the setting name is so long.

WBHarry:
> This may be cross project habits, but this seems to be a metagame setting, not standard automation. Its sortof like a world appearance setting. (Edit: its kinda similar to the spotlight queue, though still a bit different).
> 
> It might be fine if its a new tab in the automation settings, is there anything else that is more focused on showing/hiding things that GMs may or may not want players to know that we can put alongside it?
> 
> Where it is is _fine_, and we may not have a better place since there's no other metagame setting that I can find, but that's probably why the setting name is so long.

I -did- initially put it in Appearance, but then I remembered that that is Client settings - and it's not intended to be something each player gets to/has to set.
So we'd have to make a new bit of settings data in that case. (which might not be bad, since I think we have some other questionable things in automation)

WBHarry:
Had a look over all the settings. The following are ones I think could be argued don't really fit where they are (questionmark signifies that it's a stretch 😆):
```js
-- Automation Settings --
Summary Messages ?
Hide Chat Info
Damage Reduction Rules Default?
Show Resource Change ScrollTexts

-- Appearance Settings --
Show Token Distance On Hover ?
```

CarlosFdez:
Show Token Distance On Hover is definitely a stretch given a ruler does the same thing. Hide Chat Info *definitely* is a metagame setting. That means we have 2 and can go ahead with a metagame section or tab. Either works! Show Resource Change Scrolltexts might be fine as well, if it fits better than where it is.

Settings for GMs to hide things that are already automated felt weird next to settings that actually do things.

WBHarry:
Introduced Metagaming Settings. Atm it just has this new setting, but I think it's a good menu that we'll expand on.

PR #1714, opened 2026-03-08
[Feature] Custom Resources

WBHarry:
Added the ability for modules to register custom resources that will be shown through a locked tooltip accessible by a menu button next to the hope resource.

### DIFF
--- a/module/data/settings/Homebrew.mjs
+++ b/module/data/settings/Homebrew.mjs
@@ -219,8 +219,8 @@ export default class DhHomebrew extends foundry.abstract.DataModel {
                     result[key].id = key;
                     return result;
                 }, {}),
-                ...config.base,
                 ...config.custom,
+                ...config.base,
             });
         }
     }


module/applications/sheets/actors/character.mjs
```diff
@@ -942,6 +952,82 @@ export default class CharacterSheet extends DHBaseActorSheet {
...
+        /* Update resource symbols */
+        const section = target.closest('.resource-section');
+        for (const element of section.querySelectorAll('.resource-value')) {
+            if (Number.parseInt(element.dataset.value) <= value) {
+                element.querySelector('.full').classList.remove('hidden');
+                element.querySelector('.empty').classList.add('hidden');
+            } else {
+                element.querySelector('.full').classList.add('hidden');
+                element.querySelector('.empty').classList.remove('hidden');
+            }
```
CarlosFdez: Untested but try this;

```suggestion
            const showFull = Number.parseInt(element.dataset.value) <= value;
            element.querySelector('.full').classList.toggle('hidden', !showFull);
            element.querySelector('.empty').classList.toggle('hidden', showFull);
```

module/applications/sheets/actors/character.mjs
```diff
@@ -942,6 +952,82 @@ export default class CharacterSheet extends DHBaseActorSheet {
...
+
+        Hooks.once(CONFIG.DH.HOOKS.hooksConfig.lockedTooltipDismissed, () => {
+            resourceManager.classList.toggle('inverted');
+        });
+
+        for (const element of html.querySelectorAll('.resource-value'))
+            element.addEventListener('click', CharacterSheet.resourceUpdate.bind(this));
+    }
+
+    static async resourceUpdate(event) {
```
CarlosFdez: Couldn't this be a non-static function called updateResource()?

module/applications/sheets/actors/character.mjs
WBHarry: Tested and it works. Changed to it :+1:

module/applications/sheets/actors/character.mjs
WBHarry: Thought I had some issue with it being non-static, but testing it now it worked.
Changed the name aswell :+1:

WBHarry:
I added the ability to make inworld homebrew resources in the Homebrew System Menu.
I did it in a separate PR that goes against this one:
https://github.com/Foundryborne/daggerheart/pull/1718

daggerheart.mjs
```diff
@@ -97,6 +97,68 @@ Hooks.once('init', () => {
...
+        images: {
+            full: {
+                value: 'fa-solid fa-burger'
+            },
+            empty: {
+                value: 'fa-regular fa-burger'
+            }
+        }
+    };
+
```
CarlosFdez: Were these intended to be test resources not intended to be committed?

WBHarry: Yup. They're getting removed before we actually merge into main.
Figured if someone was running the branch it'd be nice not to have to know the setup.

module/applications/sheets/actors/character.mjs
```diff
@@ -942,6 +952,78 @@ export default class CharacterSheet extends DHBaseActorSheet {
...
+
+        Hooks.once(CONFIG.DH.HOOKS.hooksConfig.lockedTooltipDismissed, () => {
+            resourceManager.classList.toggle('inverted');
+        });
+
+        for (const element of html.querySelectorAll('.resource-value'))
+            element.addEventListener('click', this.updateResource.bind(this));
+    }
+
+    async updateResource(event) {
```
CarlosFdez: Sorry after saying this on the previous thing, but I just realized this is a sheet method taking an event, not a thing on the character. `onUpdateResource` might be a better name for it. Would be fine if its private.

WBHarry: Mm, that's fair. We can make that change 😎

PR #1719, opened 2026-03-09
[Feature]1672 - Character Feature Compendium Browse

### LINKED ISSUE #1672: [Bug] Adding a Feature to a sheet does not ask about the Compendium Browser
**Describe the bug**
At least for character sheets, the system asks if you want to create an item or use the compendium browser whenever you hit the `+` on a section:

Image

For Features, it does not; instead it immediately creates a new feature.

**To Reproduce**
1. Go to a sheet (character or adversary)
2. Click on the `+` in the Features section of the Features tab
3. Observe new Feature being added to the sheet instead of asking

**Expected behavior**

Features can very much be created and managed in compendiums. It would make sense to browse those when adding features to a sheet.

WBHarry:
Fixes #1672 

Fixed so that the new feature button in the character sheet can take you to the compendium browser.

### DIFF
--- a/module/applications/sheets/api/application-mixin.mjs
+++ b/module/applications/sheets/api/application-mixin.mjs
@@ -691,6 +691,9 @@ export default function DHApplicationMixin(Base) {
                 case 'weapon':
                     presets.folder = 'equipments.folders.weapons';
                     break;
+                case 'feature':
+                    presets.folder = 'features';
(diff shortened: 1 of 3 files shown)


CarlosFdez (approved):
While the ifThen in the template is scary, I dont have an alternative. Works for me.

PR #1731, opened 2026-03-14
[Fix] adversary feature type not rendering

CarlosFdez:
Removing the if in item-tags was the fix, everything else is just formatting. Something about the way the if is setup makes the check for partial-block fail.


WBHarry (approved):
Works 👍 
Wasn't really any point to an if check on the partial block either. If it would be null, then it would just show nothing in handlebars regardless.

PR #1732, opened 2026-03-15
[V14] TagTeamRoll Rework

WBHarry:


module/data/fields/action/effectsField.mjs
```diff
@@ -27,7 +27,7 @@ export default class EffectsField extends fields.ArrayField {
     static async execute(config, targets = null, force = false) {
         if (!config.hasEffect) return;
         let message = config.message ?? ui.chat.collection.get(config.parent?._id);
-        if (!message) {
+        if (!message && !config.skips.createMessage) {
```
CarlosFdez, line 30: what's the purpose of the createMessage skips? Is that related to the tag team behavior?

module/documents/actor.mjs
```diff
@@ -509,6 +510,30 @@ export default class DhpActor extends Actor {
         return await rollClass.build(config);
     }
 
+    async traitDiceRoll(trait, options = {}) {
```
CarlosFdez: Perhaps `rollTrait()`? Something more verby?

styles/less/dialog/tag-team-dialog/initialization.less
```diff
@@ -0,0 +1,59 @@
...
+                    opacity: 0.4;
+                }
+
+                .member-name {
+                    position: absolute;
+                }
+            }
+        }
+
+        .intiator-container {
```
CarlosFdez: Typo, you mean initiator-container?

styles/less/dialog/tag-team-dialog/sheet.less
```diff
@@ -1,178 +1,281 @@
...
+                    //     }
+
+                    //     img {
+                    //         height: 80px;
+                    //     }
+                    // }
+
+                    // .delete-button i {
+                    //     font-size: 40px;
+                    // }
```
CarlosFdez: Should prob kill the commented out stuff

module/data/fields/action/effectsField.mjs
WBHarry: It's not directly related, but the behavior hadn't been input in `effectsField`.
`config.skips` are meant to be generic to avoid certain behaviors.

In the case of TagTeamRolls, the rolls we make in the dialog should not be creating ChatMessages. So I found out it was missing here 🙃

styles/less/dialog/tag-team-dialog/sheet.less
WBHarry: I deded it 😎

styles/less/dialog/tag-team-dialog/initialization.less
WBHarry: You have no idea how many times I've misspelled it in this branch. it's just become a hang-up as the word keeps coming up 😆 

Fixed.

module/documents/actor.mjs
WBHarry: Yah, that's better language. Changed it :+1:

WBHarry:
Ah, now I know what I kept forgetting.
I need to look at Advantage/Disadvantage as that's not being shown atm.

But otherwise I agree. It doesn't hurt at all to bring it into the main v14 branch so it potentially gets more testing and errors are noticed.

PR #1733, opened 2026-03-15
[Fix] ItemBrowser Equipment ItemTags

WBHarry:
Fixed error with missing partial block


WBHarry:
It's definately strange. Typically the one thing you -don't- have to worry about with handlebars is falsy values, as they're always just quietely tossed. But appearantly not in this case 🙃

Issue #1052, opened 2025-08-22
[Bug] Trait name consitencies (and tooltips)


cptn-cosmo:
**Describe the bug**
Currently in character creation the trait names are written in full and not abbreviated like on the character sheet.  They both should be the same with an option to switch between abbreviated and full. Also the character creation should show the same tooltips for the traits as the character sheet (or just display the contents by default since that window can be a bit taller if needed. That would also make the use button a big taller and better looking for the suggested traits?

**Screenshots**

<img width="526" height="258" alt="Image" src="https://github.com/user-attachments/assets/0680a176-7015-4e7b-9b0e-52db460aeb31" />

Ikraik:
Resolved and merged in v14 release

PR #1737, opened 2026-03-18
[Fix] Fix CSS trait consistency in Character Creator and Sheet

Ikraik:
## Description

While I haven't added in a toggle, I added in full trait names as it would be consistent with the flow so far, and abbreviations is not mainstream yet even though it is understandable. Will revisit this in a later issue but for now traits fit in correctly within the sheet and character creator. It partially tackles issue #1052 if user prefers to use full trait names. Probably will see how to enforce the toggle as per issue.

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

- [x] Manual testing

image
image

## Checklist

## Additional Comments

In draft PR until toggle is added. Something to be discussed.

### DIFF
--- a/styles/less/dialog/character-creation/selections-container.less
+++ b/styles/less/dialog/character-creation/selections-container.less
@@ -240,16 +240,16 @@
                 display: flex;
                 align-items: center;
                 justify-content: space-evenly;
-                gap: 2px; //gap reduced to accomadate new font size and keep the same overall spacing between trait containers
+                gap: 2px;
 
                 .trait-container {
                     span {
-                        font-size: var(--font-size-10); // reduced font size to fit inside the smaller shield and maintain readability
+                        font-size: var(--font-size-10);
                     }
-                    width: 65px; //width and height increased to accomadate font size
+                    width: 65px;
                     height: 65px;
                     background: url(../assets/svg/trait-shield.svg) no-repeat;
-                    background-size: 100%; //scaling the shield to fit the container
+                    background-size: 100%;
 
                     div {
                         filter: drop-shadow(0 0 3px black);
--- a/styles/less/sheets-settings/character-settings/sheet.less
+++ b/styles/less/sheets-settings/character-settings/sheet.less
@@ -20,20 +20,20 @@
             display: flex;
             align-items: center;
             justify-content: space-evenly;
-            gap: 2px; //gap reduced to accomadate new font size and keep the same overall spacing between trait containers
+            gap: 2px;
 
             .trait-container {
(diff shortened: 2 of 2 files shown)


Ikraik:
wait. 41? da heck?

Ikraik:
OH no. it went to main... and I made the branch on v14-Dev 🫥

Ikraik:
ah! Thank you @WBHarry

styles/less/dialog/character-creation/selections-container.less
```diff
@@ -240,12 +240,17 @@
...
-                    width: 60px;
-                    height: 60px;
+                    span {
+                        font-size: var(--font-size-10); // reduced font size to fit inside the smaller shield and maintain readability
+                    }
+                    width: 65px; //width and height increased to accomadate font size
+                    height: 65px;
                     background: url(../assets/svg/trait-shield.svg) no-repeat;
+                    background-size: 100%; //scaling the shield to fit the container
+                    padding-right:2px; // trait-shield is not centered so a quick padding fix for centered text
```
CarlosFdez: Why not adjust the trait shield to remove that right padding?

styles/less/sheets-settings/character-settings/sheet.less
```diff
@@ -20,16 +20,23 @@
             display: flex;
             align-items: center;
             justify-content: space-evenly;
-            gap: 8px;
+            gap: 2px; //gap reduced to accomadate new font size and keep the same overall spacing between trait containers
 
             .trait-container {
-                width: 60px;
-                height: 60px;
+                width: 65px; //width and height increased to accomadate font size
```
CarlosFdez: I don't know about these sorts of comments, they're comments on what the code was relative to the previous version. Comments committed to the code should be evergreen, and imo best saved if something isn't obvious.

Ikraik: It was to remind myself and revert when needed.

styles/less/dialog/character-creation/selections-container.less
Ikraik: I did and it caused more pixel differences which were a headache and also was lazy to do so. In the long run its something to fix.

Ikraik:
@CarlosFdez more changes to be added. The PR is still in draft

Ikraik:
I set it for review since it is a makeshift solution. It partially solves the issue. I will add a toggle for other areas although I'm not sure how it will affect adversely there. Looking into it, it had to be case by case since css changes are on area basis like Character sheet base, character creator and wrench dialog part.

PR #1744, opened 2026-03-21
[V14] Armor System ArmorScore

WBHarry:
Readded so that armor items have their system defined armor instead of using an ActiveEffect.

- The ScrollingTexts are currently screwy when you change the total armor value enough that it updates multiple effects/armors at the same time.
- I haven't updated the SRD Armors to this yet. Want to verify the solution first 🙃

### DIFF
--- a/src/packs/items/armors/armor_Advanced_Chainmail_Armor_LzLOJ9EVaHWAjoq9.json
+++ b/src/packs/items/armors/armor_Advanced_Chainmail_Armor_LzLOJ9EVaHWAjoq9.json
@@ -5,6 +5,10 @@
   "_id": "LzLOJ9EVaHWAjoq9",
   "img": "icons/equipment/chest/breastplate-banded-steel-gold.webp",
   "system": {
+    "armor": {
+      "current": 0,
+      "max": 6
+    },
     "description": "",
     "actions": {},
     "attached": [],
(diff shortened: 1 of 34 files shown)


module/applications/sheets/actors/character.mjs
```diff
@@ -999,49 +1003,49 @@ export default class CharacterSheet extends DHBaseActorSheet {
...
+                ...change,
+                value: change.type === 'armor' ? { ...change.value, current: newCurrent } : change.value
+            }));
+
+            await effect.update({ 'system.changes': newChanges });
+        }
 
         const container = target.closest('.slot-bar');
         for (const armorSlot of container.querySelectorAll('.armor-slot i')) {
-            const marked = !decreasing && Number.parseInt(armorSlot.dataset.index) < newCurrent;
```
WBHarry, line 1039: I reverted this because the logic ended up wrong. When you deselected armor in the `ArmorManagement` overlay it would deslect all the icons visually no matter how many you clicked off.

CarlosFdez: I tested it personally and it was right? Could be the actual logic of marked was wrong, but using toggle(class, force) is significantly less verbose. Could it be that marked was undefined? It only works if the param is true or false.

Even this would be better and is effectively the same as what you have now, though:
```ts
armorSlot.classList.toggle('fa-shield', !decreasing && index < newCurrent);
armorSlot.classList.toggle('fa-shield-halved', decreasing && index >= newCurrent);
```

CarlosFdez: Oh wait, hmm, its not the same. Is there a hypothetical scenario where neither happens?

PR #1746, opened 2026-03-23
[V14] 1743 - Damage Update Error

### LINKED ISSUE #1743: [V14] [Bug] DamageData update error
Any updates to an Actor/Item that contains `DHDamageData` (damageField.mjs) will cause an error in the update.
Image

Few ways to trigger this.
1) Try to make an update in an Adversary's EditSheet - the system.attack will cause an error due to its damage field.
2) Make a Damage Action on a feature and try to add a damage type in the Effects tab.

WBHarry:
Fixes #1743
Fixed DamageParts causing errors on update

There is still an issue in that if you try to delete a damage part in the ActionConfig, it won't be deleted until you reload foundry. Not sure why 🤔 
It's in `action-base-config` `static removeDamage` --> Ends up in `actionField` `async update`

### DIFF
--- a/module/applications/sheets-configs/action-base-config.mjs
+++ b/module/applications/sheets-configs/action-base-config.mjs
@@ -316,7 +316,8 @@ export default class DHActionBaseConfig extends DaggerheartSheet(ApplicationV2) 
         const callback = (_, button) => {
             const data = this.action.toObject();
             const type = choices[button.form.elements.type.value].value;
-            const part = game.system.api.fields.ActionFields.DamageField.getDefaultDamagePart(type);
+            const part = this.action.schema.fields.damage.fields.parts.element.getInitialValue();
+            part.applyTo = type;
             data.damage.parts[type] = part;
             this.constructor.updateForm.bind(this)(null, null, { object: foundry.utils.flattenObject(data) });
         };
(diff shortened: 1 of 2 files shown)


WBHarry:
I don't get what's going on here. The `system.actions.<id>.damage.parts.hitPoints: _del` update just won't affect the data until you F5. Both in memory and `source` does not show the change having happened 🤯

WBHarry:
`IterableTypedObjectField` doesn't -seem- to be the issue to me. I tested setting up a random one on Character and tried updating/deleting keys on that one without issues.

module/applications/sheets-configs/action-base-config.mjs
```diff
@@ -312,8 +316,7 @@ export default class DHActionBaseConfig extends DaggerheartSheet(ApplicationV2)
         const callback = (_, button) => {
             const data = this.action.toObject();
             const type = choices[button.form.elements.type.value].value;
-            const part = { applyTo: type };
-            if (this.action.actor?.isNPC) part.value = { multiplier: 'flat' };
+            const part = game.system.api.fields.ActionFields.DamageField.getDefaultDamagePart(type);
```
CarlosFdez: Can't use getInitialValue()?

WBHarry: Hm..might be cleaner to set up an initial value on the field, yeah. Currently it's undefined.
I'll move it over there :+1:

WBHarry: Yup: `this.action.schema.fields.damage.fields.parts.element.getInitialValue()` worked like a charm out of the box. For some reason I thought it wouldn't resolve to anything.

PR #1750, opened 2026-03-25
[Feature] 1740 - Beastform Info

### LINKED ISSUE #1740: [Feature] Beastform Info Display
**Description**
Beastform is a main feature of one of the core classes, but currently a player using a Druid has no way to easily read all of the info about their current beastform in one place.

A player playing will therefore have to have a separate file or pdf open with their beastform information (I.e, core rule book or SRD) in order to make informed decisions on how to use their beastform.

Idea 1 is to create something like a feature or context-aware section of the sheet that gets filled in with all of the beastform info that you currently have on.

Idea 2 (which is not mutually exclusive with idea 1, but probably simpler to implement at first) is to have a beastform browser. Similar to what you get when you click the beastform transformation, but that you can click and simply open the beastform browser to read the different beastforms. This would be a good stopgap as the player then would have a per-beastform way to read the info (barring the ones that are modifications on other beastforms, like legendary beast)

WBHarry:
Fixes #1740 
Improved the EffectDisplay tooltip of the beastform effect to show the info about the active beastform
image

### DIFF
--- a/module/systemRegistration/handlebars.mjs
+++ b/module/systemRegistration/handlebars.mjs
@@ -36,6 +36,7 @@ export const preloadHandlebarsTemplates = async function () {
         'systems/daggerheart/templates/actionTypes/summon.hbs',
         'systems/daggerheart/templates/actionTypes/transform.hbs',
         'systems/daggerheart/templates/settings/components/settings-item-line.hbs',
+        'systems/daggerheart/templates/ui/tooltip/parts/beastformData.hbs',
         'systems/daggerheart/templates/ui/tooltip/parts/tooltipChips.hbs',
         'systems/daggerheart/templates/ui/tooltip/parts/tooltipTags.hbs',
         'systems/daggerheart/templates/dialogs/downtime/activities.hbs',
@@ -49,7 +50,6 @@ export const preloadHandlebarsTemplates = async function () {
         'systems/daggerheart/templates/ui/itemBrowser/itemContainer.hbs',
         'systems/daggerheart/templates/scene/dh-config.hbs',
         'systems/daggerheart/templates/settings/appearance-settings/diceSoNiceTab.hbs',
-        'systems/daggerheart/templates/ui/tooltip/parts/beastformData.hbs',
         'systems/daggerheart/templates/sheets/activeEffect/typeChanges/armorChange.hbs'
     ]);
 };


module/data/item/beastform.mjs
```diff
@@ -97,20 +97,19 @@ export default class DHBeastform extends BaseDataItem {
...
     /* -------------------------------------------- */
 
-    get beastformAttackData() {
-        const effect = this.parent.effects.find(x => x.type === 'beastform');
+    static getBeastformAttackData(effect) {
         if (!effect) return null;
 
-        const traitBonus =
-            effect.system.changes.find(x => x.key === `system.traits.${this.mainTrait}.value`)?.value ?? 0;
+        const mainTrait = effect.system.changes.find(x => x.key === 'system.rules.attack.roll.trait')?.value;
```
CarlosFdez, line 108: This might benefit from foundry.utils.expandObject() maybe?

module/documents/tooltipManager.mjs
```diff
@@ -31,12 +31,34 @@ export default class DhTooltipManager extends foundry.helpers.interaction.Toolti
...
                 effect = {
                     ...effectData,
-                    name: game.i18n.localize(effectData.name),
-                    description: game.i18n.localize(effectData.description ?? effectData.parent.system.description)
+                    name: game.i18n.localize(effectData.name)
                 };
+
+                if (effectData.type === 'beastform') {
+                    const beastformData = {
+                        features: effectItem.parent.items.filter(x => effectItem.system.featureIds.includes(x.id)),
```
CarlosFdez: It looks like you write the enriched description into the item, and then rename it to avoid overwrite. It might be better to avoid mutations and instead map this to some data, even if its only toObject()? Or at least a list of objects that are `{ item, description }` objects.

module/data/item/beastform.mjs
WBHarry: Hm. In what way? Running foundry.utils.expandObject on `effect.system.changes`  doesn't really do anything since it's an array

CarlosFdez: no you're right, I'm mixing up active effects and changes.

module/documents/tooltipManager.mjs
WBHarry: True! I changed it around so that we build a `feature.toObject()` list with the enrichedDescriptions
```js
const beastformData = {
    features: [],
    advantageOn: effectData.system.advantageOn,
    beastformAttackData: game.system.api.data.items.DHBeastform.getBeastformAttackData(effectItem)
};

const features = effectItem.parent.items.filter(x => effectItem.system.featureIds.includes(x.id));
for (const feature of features) {
    const featureData = feature.toObject();
    featureData.enrichedDescription = await feature.system.getEnrichedDescription();
    beastformData.features.push(featureData);
}
```

Ikraik:
Do you think this can merge with beastform evolution PR?

WBHarry:
Well, it doesn't need to merge with it. They're two different PRs that are both going to go into v14-Dev. If either impacts the other, that should show up in a merge conflict once the first is in 🙃

PR #1751, opened 2026-03-25
[Feature] Beastform Evolution Handling

WBHarry:
Added so that beastform actions can have trait bonuses to be selected while transforming

### DIFF
--- a/module/applications/dialogs/beastformDialog.mjs
+++ b/module/applications/dialogs/beastformDialog.mjs
@@ -10,6 +10,12 @@ export default class BeastformDialog extends HandlebarsApplicationMixin(Applicat
         this.selected = null;
         this.evolved = { form: null };
         this.hybrid = { forms: {}, advantages: {}, features: {} };
+        this.modifications = {
+            traitBonuses: configData.modifications.traitBonuses.map(x => ({
+                trait: null,
+                bonus: x.bonus
+            }))
+        };
 
         this._dragDrop = this._createDragDropHandlers();
     }
@@ -28,6 +34,7 @@ export default class BeastformDialog extends HandlebarsApplicationMixin(Applicat
             selectBeastform: this.selectBeastform,
             toggleHybridFeature: this.toggleHybridFeature,
             toggleHybridAdvantage: this.toggleHybridAdvantage,
+            toggleTraitBonus: this.toggleTraitBonus,
             submitBeastform: this.submitBeastform
         },
         form: {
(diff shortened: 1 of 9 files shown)


module/applications/dialogs/beastformDialog.mjs
```diff
@@ -156,11 +167,14 @@ export default class BeastformDialog extends HandlebarsApplicationMixin(Applicat
 
     canSubmit() {
```
CarlosFdez, line 168: Try making this function the following. The suggestions feature isn't really working well. I don't play a druid so I don't really know what to test. It looks like you always return false if modifications are not finished, so you can early return. note that `[].every(() => false)` returns true, so you don't have to check if its empty.

```js
        const modificationsFinished = this.modifications.traitBonuses.every(x => x.trait);
        if (!modificationsFinished) return false;
            
        if (this.selected) {
            switch (this.selected.system.beastformType) {
                case 'normal':
                    return true;
                case 'evolved':
                    return this.evolved.form;
                case 'hybrid':
                    const selectedAdvantages = Object.values(this.hybrid.advantages).reduce(
                        (acc, form) => acc + Object.values(form).length,
                        0
                    );
                    const selectedFeatures = Object.values(this.hybrid.features).reduce(
                        (acc, form) => acc + Object.values(form).length,
                        0
                    );

                    const advantagesSelected = selectedAdvantages === this.selected.system.hybrid.advantages;
                    const featuresSelected = selectedFeatures === this.selected.system.hybrid.features;
                    return advantagesSelected && featuresSelected;
            }
        }

        return false;
```

module/applications/dialogs/beastformDialog.mjs
```diff
@@ -292,6 +313,24 @@ export default class BeastformDialog extends HandlebarsApplicationMixin(Applicat
                         }
                     }
 
+                    const beastformEffect = selected.effects.find(x => x.type === 'beastform');
+                    while (app.modifications.traitBonuses.length) {
```
CarlosFdez, line 317: Unless the ordering matters (thanks pop), I think you can instead do a for loop and then empty the array (arr.length = 0 actually clear it fsr, or you could reassign). I think its best to avoid while unless absolutely necessary, they're p.dangerous on projects with contributors especially.

CarlosFdez: Could also be a reduce actually to return new beastFormEffect.changes

module/data/fields/action/beastformField.mjs
```diff
@@ -28,8 +28,21 @@ export default class BeastformField extends fields.SchemaField {
...
                         );
                     },
-                    hint: 'DAGGERHEART.ACTIONS.Config.beastform.exactHint'
+                    label: 'DAGGERHEART.ACTIONS.Config.beastform.exact.label',
+                    hint: 'DAGGERHEART.ACTIONS.Config.beastform.exact.hint'
                 })
+            }),
+            modifications: new fields.SchemaField({
+                traitBonuses: new fields.ArrayField(
+                    new fields.SchemaField({
```
CarlosFdez, line 37: Are there any other fields you envision here. Looks like this could be an array of trait bonus strings.

WBHarry: I -could- imagine having a restriction field. IE, limiting what trait this bonus is allowed to be put towards.

WBHarry: Not something I particularly care for adding atm or anything, but something like that could be a possibility for some usecase.

CarlosFdez: if there's a possibility then its fine, I was just wondering why only one thing in the schema.

PR #1755, opened 2026-03-28
[V14] [Feature] Spotlight Without Combat

WBHarry:
A token having a turnmarker is now determined by a global setting, and not wether its a combatants turn


module/applications/ui/combatTracker.mjs
```diff
@@ -187,6 +188,16 @@ export default class DhCombatTracker extends foundry.applications.sidebar.tabs.C
             round: this.viewed.round + 1
         });
         await combatant.update(update);
+
+        if (!combatant.token) return;
+        clearPreviousSpotlight();
```
CarlosFdez:
```suggestion
        if (combatant.token) clearPreviousSpotlight();
```

PR #1757, opened 2026-03-30
[V14] Message Rolls Rework

WBHarry:


CarlosFdez:
Should FateRoll.dHope and FateRoll.dFear return null if the type doesn't match?

CarlosFdez:
Done checking. So far it seems fine, though I did this https://github.com/Foundryborne/daggerheart/pull/1758 to simplify the modifier retrieval (getting the parenthetical modifiers is actually quite easy!). Personally I would like it if we could do without AdvantageDie and DisadvantageDie at the very least, but I won't bother too much about it.

module/applications/ui/chatLog.mjs
```diff
@@ -209,27 +210,17 @@ export default class DhpChatLog extends foundry.applications.sidebar.tabs.ChatLo
...
-                'rolls': [parsedRoll]
+            const rerollDice = message.system.roll.dice[target.dataset.dieIndex];
+            await rerollDice.reroll(`/r1=${rerollDice.total}`, {
+                liveRoll: {
+                    roll: message.system.roll,
+                    actor: message.system.actionActor,
+                    isReaction: message.system.roll.options.actionType === 'reaction'
+                }
+            });
+            // await message.system.roll._evaluate();
```
CarlosFdez:
```suggestion
```

WBHarry:
> Should FateRoll.dHope and FateRoll.dFear return null if the type doesn't match?

Mm..I'm a bit scared of messing with it right now. I think it can be looked at next time we mess with this, since it doesn't actually negatively effect anything.

PR #1758, opened 2026-03-30
Compute modifiers using deterministic terms

CarlosFdez:
Move getter to top level, and use isDeterministic for computation


CarlosFdez:
Turns out the naughty setters are used indirectly in an updateRollConfiguration() function. I'll need to see if I can clean that up somehow, since the setters being asymmetrical are a design problem.

EDIT: Maybe for another time, that's a more involved problem. Just the modifier retrieval then.

PR #1769, opened 2026-04-02
[Fix] Template Scene Awareness

WBHarry:
Fixed two things relating to our placeable templates and scenes

- All grid values are now taken from the currently viewed scene `canvas.scene` instead of `game.scenes.active`, since a user could be viewing and interacting with an inactive scene.
- The templates now care about the scene overrides for daggerheart distances rather than just always using the global setting.

### DIFF
--- a/module/enrichers/TemplateEnricher.mjs
+++ b/module/enrichers/TemplateEnricher.mjs
@@ -65,9 +65,9 @@ export const renderMeasuredTemplate = async event => {
 
     let baseDistance = getTemplateDistance(range);
 
-    const sceneGridSize = canvas.scene?.grid.size ?? CONFIG.Scene.documentClass.schema.fields.grid.fields.size.initial;
-    const sceneGridDistance =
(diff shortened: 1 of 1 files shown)


module/enrichers/TemplateEnricher.mjs
```diff
@@ -63,14 +63,9 @@ export const renderMeasuredTemplate = async event => {
...
-    let baseDistance = range;
-    if (Number.isNaN(Number(range))) {
-        baseDistance = game.settings.get(CONFIG.DH.id, CONFIG.DH.SETTINGS.gameSettings.variantRules).rangeMeasurement[
-            range
-        ];
-    }
+    let baseDistance = getTemplateDistance(range);
 
-    const dimensionConstant = game.scenes.active.grid.size / game.scenes.active.grid.distance;
+    const dimensionConstant = (canvas.scene?.grid.size ?? 100) / (canvas.scene?.grid.distance ?? 5);
```
CarlosFdez: Is divide by undefined safe? Didn't check

PR #1770, opened 2026-04-03
[Feature] 1766 - Group Attack

### LINKED ISSUE #1766: [Feature] Group Attack Implementation
**Description**
deal n x m + z damage, where n is the amount of currently selected tokens.
So for example n + 4 would deal 4 damage for each selected adversary token etc...

> I would imagine it as using the ability and being prompted to select tokens onscreen (similar to how you place tokens when using summon), with the currently selected ones pre-selected.  People have random stuff selected all the time, so just going off that straight-up is liable to make it wrong a lot of the time.

> (A simple number input would) be better for people playing without a canvas. That's not really what I want to be the primary design focus though.
We could check if foundry is running in no-canvas mode and pop an input in that case.

WBHarry:
Fixes #1766 
Implemented group attack on action damage

### DIFF
--- a/module/config/generalConfig.mjs
+++ b/module/config/generalConfig.mjs
@@ -70,6 +70,14 @@ export const range = {
     }
 };
 
+export const groupAttackRange = {
+    melee: range.melee,
+    veryClose: range.veryClose,
+    close: range.close,
+    far: range.far,
+    veryFar: range.veryFar
+};
+
 /* circle|cone|rect|ray used to be CONST.MEASURED_TEMPLATE_TYPES. Hardcoded for now */
 export const templateTypes = {
     CIRCLE: 'circle',
--- a/module/data/action/baseAction.mjs
+++ b/module/data/action/baseAction.mjs
@@ -264,6 +264,7 @@ export default class DHBaseAction extends ActionMixin(foundry.abstract.DataModel
             hasSave: this.hasSave,
             onSave: this.save?.damageMod,
             isDirect: !!this.damage?.direct,
+            damageOptions: this.damage?.groupAttack ? {} : null,
             selectedMessageMode: game.settings.get('core', 'messageMode'),
             data: this.getRollData(),
             evaluate: this.hasRoll,
@@ -279,6 +280,26 @@ export default class DHBaseAction extends ActionMixin(foundry.abstract.DataModel
                 ...(configOptions.skips ?? {})
             }
         };
+
+        if (this.damage) {
+            config.isDirect = this.damage.direct;
+
+            const groupAttackTokens = this.damage.groupAttack
+                ? game.system.api.fields.ActionFields.DamageField.getGroupAttackTokens(
+                      this.actor.id,
+                      this.damage.groupAttack
+                  )
+                : null;
+
+            config.damageOptions = {
+                groupAttack: this.damage.groupAttack
+                    ? {
+                          nr: Math.max(groupAttackTokens.length, 1),
+                          range: this.damage.groupAttack
+                      }
+                    : null
+            };
+        }
 
         DHBaseAction.applyKeybindings(config);
         return config;
--- a/module/data/action/baseAction.mjs
+++ b/module/data/action/baseAction.mjs
@@ -264,7 +264,6 @@ export default class DHBaseAction extends ActionMixin(foundry.abstract.DataModel
             hasSave: this.hasSave,
             onSave: this.save?.damageMod,
             isDirect: !!this.damage?.direct,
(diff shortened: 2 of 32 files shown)


CarlosFdez:
For variable names, I think we can do better than "nr"?

WBHarry:
> For variable names, I think we can do better than "nr"?

Hm..I guess maybe `groupAttack.damageMultiplier`?
It's the number of minions that are involved, and thus the multiplier for the damage.

CarlosFdez:
That can work. I can also see `attackers`, `numAttackers`, `size` working too maybe.

WBHarry:
Changed it to `numAttackers` so it's clear it's a number and not an array of tokens/actors :+1:

module/data/fields/action/damageField.mjs
```diff
@@ -224,6 +229,27 @@ export default class DamageField extends fields.SchemaField {
...
+        if (!canvas.scene) return [];
+
+        const targets = Array.from(game.user.targets);
+        const { custom } = CONFIG.DH.GENERAL.sceneRangeMeasurementSetting;
+        const sceneMeasurements = canvas.scene?.flags.daggerheart?.rangeMeasurement;
+        const globalMeasurements = game.settings.get(
+            CONFIG.DH.id,
+            CONFIG.DH.SETTINGS.gameSettings.variantRules
+        ).rangeMeasurement;
+        const rangeSettings = sceneMeasurements?.setting === custom.id ? sceneMeasurements : globalMeasurements;
```
CarlosFdez: This looks like something that can become a getter in the scene by this point? I think we do this sort of thing in a few other locations.

WBHarry: Hm, I think it'd have to become a utility method then.
If I do something like this:
```js
 get rangeSettings() {
      const { custom } = CONFIG.DH.GENERAL.sceneRangeMeasurementSetting;
      const sceneMeasurements = this.flags.daggerheart?.rangeMeasurement;
      const globalMeasurements = game.settings.get(
          CONFIG.DH.id,
          CONFIG.DH.SETTINGS.gameSettings.variantRules
      ).rangeMeasurement;
      return sceneMeasurements?.setting === custom.id ? sceneMeasurements : globalMeasurements;
  }
```

Then we have the issue that we won't be able to get the global setting when the scene is undefined (happens on world loading), since the method is now on the scene, which is undefined.

CarlosFdez: What are the situations we would want to get the normal settings when there is no scene, but get the scene settings if there is one? This for example can't do anything if there are no tokens. Same with the tooltips and AE automation stuff.

WBHarry: We definately don't care about actually getting something. It's more so that the methods being run end up being run before the scene has been initialized. ActiveTemplate and such are set up before the scene is for some reason.
But, it's fair enough to check if the methods return null and aborting.

module/data/action/baseAction.mjs
```diff
@@ -264,6 +264,7 @@ export default class DHBaseAction extends ActionMixin(foundry.abstract.DataModel
             hasSave: this.hasSave,
             onSave: this.save?.damageMod,
             isDirect: !!this.damage?.direct,
+            damageOptions: this.damage?.groupAttack ? {} : null,
```
CarlosFdez: The way this is broken up is a bit odd to read and it took me a bit to figure out what was going on.  Would it be possible to combine this somehow? I could see an IIFE with an early return working, but maybe you have something else in mind.

module/config/generalConfig.mjs
```diff
@@ -70,6 +70,14 @@ export const range = {
     }
 };
 
+export const groupAttackRange = {
+    melee: range.melee,
+    veryClose: range.veryClose,
+    close: range.close,
+    far: range.far,
+    veryFar: range.veryFar
+};
```
CarlosFdez, line 79: Nothing to fix here with it as it is, but one more point for using remeda :P

CarlosFdez:
It might be nice to have a tooltip over the target button to clarify what its for, but it works very well, nice job.

module/data/action/baseAction.mjs
WBHarry: Hm...yeah I'm not really sure why I bothered to declare damageOptions here. It should work to just rely on the later `if`. I think it's preferable the `damageOptions` key doesn't exist on the config at all if it's not an action with `this.damage`

WBHarry: I removed the initial declaration.
We could always extract the if to some `setupDamageConfig` kinda deal to modify the config if that's cleaner.

PR #1771, opened 2026-04-03
[Feature] 1614 - Add Evasion To Party Resources

### LINKED ISSUE #1614: [Feature] Add Evasion to the Party Overview
**Description**
On the Party Actor, there is currently no Evasion been shown.

Including it would enable the GM to have a better overview, since Evasion is often referenced during play. This could be specially valuable for groups that do no use automation to handle targeting and so on.

**Screenshots**

Image

WBHarry:
Fixes #1614 
Added evasion to party resources
image

### DIFF
--- a/styles/less/sheets/actors/party/resources.less
+++ b/styles/less/sheets/actors/party/resources.less
@@ -155,6 +155,26 @@ body.game:is(.performance-low, .noblur) {
                         }
                     }
 
+                    .stat-section {
+                        position: relative;
+                        display: flex;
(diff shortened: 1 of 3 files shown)


CarlosFdez:
I think we may be able to reduce on styles, the stuff being added to evasion looks like the rest already, but I think party resources could use a redesign regardless.

WBHarry:
That's fine with me. Always open for a different design idea. 

I guess we bring this in so we have a full data set for any such efforts.

PR #1772, opened 2026-04-03
[Fix] Multiterm Armor Effect Max

WBHarry:
Fixed so that multi term expressions get evaluated into a single number.
EX: Bare bones is set to max: `3 + @system.traits.strength.value`. After `itemAbleRollParse` that ends up as `3 + 2` which isn't a number, so it doesn't get applied. Now I use a roll to grab a total number. Might be another way to do it, don't recall.

### DIFF
--- a/module/data/activeEffect/changeTypes/armor.mjs
+++ b/module/data/activeEffect/changeTypes/armor.mjs
@@ -44,7 +44,8 @@ export default class ArmorChange extends foundry.abstract.DataModel {
         label: 'Armor',
         defaultPriority: 20,
         handler: (actor, change, _options, _field, replacementData) => {
-            const parsedMax = itemAbleRollParse(change.value.max, actor, change.effect.parent);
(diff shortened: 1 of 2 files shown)


module/data/activeEffect/changeTypes/armor.mjs
```diff
@@ -44,7 +44,8 @@ export default class ArmorChange extends foundry.abstract.DataModel {
         label: 'Armor',
         defaultPriority: 20,
         handler: (actor, change, _options, _field, replacementData) => {
-            const parsedMax = itemAbleRollParse(change.value.max, actor, change.effect.parent);
+            const baseParsedMax = itemAbleRollParse(change.value.max, actor, change.effect.parent);
+            const parsedMax = new Roll(baseParsedMax).evaluateSync().total;
```
CarlosFdez, line 48: This is fine, but we should probably refactor such that this problem is fixed more foundationally soon. Remind me if I forget.

PR #1784, opened 2026-04-07
[Feature] Redesign and merge party members and resources tabs

CarlosFdez:
There are probably things that can be improved on, and I'm 50/50 on whether we should even show traits (some people want them). But please give it a go and let me know if this possibly a direction we should go.

image


WBHarry:
It's really neat. The hope and stats fitting snuggly together tickles my brain right.

WBHarry:
I think we lost the edit and delete functionality as is
image

**Edit**: Ah, no i see. The name and the x next to it :+1:

WBHarry:
I realise this was probably a space consideration, but I don't think writing `1hp` ... `2hp` looks very good for the damage thresholds :thinking:
image

WBHarry:
Do we slot in Evasion underneath the traits perhaps?

CarlosFdez:
1hp/2hp/3hp was an idea I stole from DH+. I considered MIN / MAJ / SEV but would that be any better?

Evasion is the shield to the top right of the portrait. [9] in that case. Was it not very visible? I was hoping it was visible enough that a tooltip would explain it on the offchance someone didn't understand it.

WBHarry:
Ah, I see. I imediately thought of armor - in spite of then also seeing armor along with the other resources 🙃 
It is probably good as is, yeah. I'm quite tired tonight, and moving the cursor over there once should clarify it. It's the sort of thing you learn, and then know without issues.

WBHarry:
In regards to the damge thresholds, there probably isn't really a better shorthand label no.
After trying it, I think it might make it visually better if we have a space in the `hp` label though (top 2)
image

CarlosFdez:
Sure we can give that a try. I was also down to try MIN/MAJ/SEV if that looked better

WBHarry (approved):
Either way, I think it's a much better information layout, and combinin the tabs is also great so we don't have the functionality spread out :+1:

Something we could look at as a followup at some point would be to add in the custom resource drop down from the Character sheet so that could also be available here.

CarlosFdez:
I gave increasing the hp a try, but I needed to increase the gap a smidge, and now its too close to the resources. Though we can maybe increase the gap to the right of the portrait to compensate. You didn't test it with double digit thresholds.

image

CarlosFdez:
With shorthand labels instead:
image

PR #1785, opened 2026-04-08
[Feature] Group Roll Rework

### LINKED ISSUE #1276: [Feature] More info on the Group Rolls
**Description**
When making a group roll and selecting which trait do you want to use, the traits are shown as a chat message. 
You can then click on the dice to roll and change the trait. It will keep the original trait's name. Currently there is not way to know the formula used for a group roll and which trait is used if you can change it afterwards.

My suggestions:
Could the formula for each roll be shown?
Maybe the trait can be updated in the chat message if you select another one after setting the group roll? 
Or perhaps if you choose a trait when creating a group roll you can't change it afterwards? 

**Screenshots**

Image

### LINKED ISSUE #1450: [Bug] Group roll reaction rolls being deleted
**Describe the bug**
When selecting the option to make a group roll and the players that are assisting make reaction rolls, the options to click pass/fail disappear and the result is deleted/restarted if another reaction roll from the group roll is made before the GM sets the previous one to pass or fail

**To Reproduce**
1. Set a group roll with its leader and two-three assisting players
2. One assisting player makes a reaction roll but you don't click pass or fail after the result is shown.
3. A second assisting player makes a reaction roll.
4. See that the options pass/fail from the first reaction roll disappear and the roll is lost and you have to redo it.

**Expected behavior**
If the players roll, regardless of waiting or doing them at the same time, the options should not disappear.

**Setup Information:**
 - OS: Windows
 - Browser Firefox
 - Foundry Version V13 b351
- System Version v.1.3.2

### LINKED ISSUE #1496: [Feature] Group Rolls as Reactions
When supporting a group roll, only the character nominated to make the main roll makes a Duality Roll. The rest of the supporting characters should make Reaction Rolls according to DH Core p97.

Request Foundryborne to automatically default to Reaction Roll so that PCs don't accidently make a Duality Roll then have to manually remove the Hope or Fear from the check.

Win 11
Chrome Version 143.0.7499.111 (Official Build) (64-bit)
Foundry v13 Build 351
Foundryborne 1.4.4

Image

WBHarry:
Fixes #1276 
Fixes #1450 
Fixes #1496
Group Roll rework in the same style of the reworked Tag Team Roll


CarlosFdez (approved):
Anything I can comment on is minor code stuff. Like I think can maybe make the result into a partial, and handle the leader/member disjunction by adding a data-member for aiding id or "leader" at the top level and merging rerollDice and rerollLeaderDice and such. But the gains are small.

I have ideas for refinements, but they can always be additional features or improvements.

WBHarry:
> Anything I can comment on is minor code stuff. Like I think can maybe make the result into a partial, and handle the leader/member disjunction by adding a data-member for aiding id or "leader" at the top level and merging rerollDice and rerollLeaderDice and such. But the gains are small.
> 
> I have ideas for refinements, but they can always be additional features or improvements.

Lets handle improvements as future improvements then whenever inspiration strikes. I got it tested with a player, and the rerendering works without issues 🥳

PR #1799, opened 2026-04-14
Fix Include Item Damage error in v14 damage system

### LINKED ISSUE #1794: [Bug] Include Item Damage Not Working
In v14 we changed the damage system to only allow a single instance of damage to a resource. We forgot to handle the case with `Include Item Damage` on damage actions on Weapons though.

It's currently producing an error in `attackAction.prepareData` as `this.damage.parts.unshift(new DHDamageData(baseDamage));` fails.

---
We can either:
1) Change this behavior to instead be `Use Item Damage` - IE, toggle this if you want to use the weapon's damage, meaning you don't change it at all. This would be the easiest, and I could see it being fair enough.

2) If we do want to allow Item Damage + a bonus in the action config, then we have to consider that the item damage and what's added can have different dice sizes. This means that we would have to smash them together as a custom formula.
It could be done either in `attackAction.prepareData`, or perhaps more neatly by rendering the ActionConfig differently when the checkbox has been clicked. It could auto switch to the formula view and show the `item.damage` as a locked initial part of the formula which you then add to.

codebytaki:
- Change behavior from 'Include Item Damage' to 'Use Item Damage'
- Replace action damage with weapon base damage instead of unshift
- Fixes error where unshift fails due to v14 single damage instance limit
- Update label to reflect new behavior

Fixes Foundryborne/daggerheart#1794

---
name: Pull Request
about: Create a new pull request
title: '[PR] <Insert Title here>'
labels: pr
assignees: ''
---

Is this a community PR? Please go to preview tab and click [here](?expand=1&template=community_pull_request_template.md). If not, delete this line.

## Description

- Fixes #(issue)
- Closes #(issue)

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Code cleanup/refactor
- [ ] Documentation update
- [ ] Test coverage
- [ ] Dependency update
- [ ] Configuration change

## How Has This Been Tested?

- [ ] Manual testing
- [ ] Other:

## Checklist

- [ ] My code follows the project style guidelines
- [ ] I have performed a self-review of my code
- [ ] My changes generate no new warnings or errors

## Additional Comments

---

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,34 +1,49 @@
 ---
-name: Bug report
-about: Create a bug report to help us identify issues and resolve them
-title: "[Bug] <Insert Title here> "
-labels: bug
-type: bug
+name: 🐛 Bug Report
+about: Create a report to help us improve
+title: '[BUG] '
+labels: 'bug'
 assignees: ''
 ---
 
-**Describe the bug**
-A clear and concise description of what the bug is.
+## Bug Description
 
-**To Reproduce**
-Steps to reproduce the behavior:
+<!-- A clear and concise description of what the bug is -->
+
+## Steps To Reproduce
+
+<!-- Steps to reproduce the behavior -->
+
 1. Go to '...'
 2. Click on '....'
 3. Scroll down to '....'
 4. See error
 
-**Expected behavior**
-A clear and concise description of what you expected to happen.
+## Expected Behavior
 
-**Screenshots**
-If applicable, add screenshots to help explain your problem.
+<!-- A clear and concise description of what you expected to happen -->
 
-**Setup Information:**
- - OS: [e.g. iOS, Windows]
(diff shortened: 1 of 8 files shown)


WBHarry:
It's possible we will go down this route of fully overriding with the Item Damage, but it's not decided yet.
We do also have some other things in the pipe that might effect how we want to solve it.

Going to decline it, but we do have it on our radar.

codebytaki:
Okay, I will try some other method and fix it with multimethod.

WBHarry:
We've recently changed our contribution policy to not accept undiscussed solutions.
We've done this for various reasons,

But the result is that we'd like any ideas for changes to be discussed in an Issue. It means potential contributors don't waste their time on something we don't want to implement, are already working on, or that we want to do in a different way - and it also makes us feel a lot better by not having to decline things we've never seen before 🙃 

---
So if you have an idea of how the solution would be structured you can speak to that in the linked issue.

PR #1802, opened 2026-04-14
[Feature] Advantage Default Dice

WBHarry:
Added rules for default advantage and disadvantage dice

### DIFF
--- a/module/applications/dialogs/d20RollDialog.mjs
+++ b/module/applications/dialogs/d20RollDialog.mjs
@@ -156,12 +156,7 @@ export default class D20RollDialog extends HandlebarsApplicationMixin(Applicatio
         }
         if (this.config.uses) this.config.uses = foundry.utils.mergeObject(this.config.uses, rest.uses);
         if (rest.roll?.dice) {
-            Object.entries(rest.roll.dice).forEach(([key, value]) => {
-                if(key === 'advantageFaces')
-                    this.roll[key] = Number.parseInt(value);
-                else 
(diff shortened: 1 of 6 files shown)


module/applications/dialogs/d20RollDialog.mjs
```diff
@@ -153,7 +157,10 @@ export default class D20RollDialog extends HandlebarsApplicationMixin(Applicatio
         if (this.config.uses) this.config.uses = foundry.utils.mergeObject(this.config.uses, rest.uses);
         if (rest.roll?.dice) {
             Object.entries(rest.roll.dice).forEach(([key, value]) => {
-                this.roll[key] = value;
+                if(key === 'advantageFaces')
+                    this.roll[key] = Number.parseInt(value);
```
CarlosFdez: You can completely avoid this if you add ` data-dtype="Number"` to the select in rollSelection.hbs. Afterwards I think you can maybe do `this.roll = foundry.utils.mergeObject(this.roll, rest.roll.dice)` and it should work maybe.

WBHarry: Ah..I think I put it on the select helper yesterday and not on the select. i was confused why it didn't work 😆

WBHarry: Using `data-dtype` and `foundry.utils.mergeObject`. Neater now :+1:

CarlosFdez (approved):

PR #1803, opened 2026-04-14
[Feature] Active Party

WBHarry:
- A Party can now be set as the `active` party via the context menu in the actor's tab. If no party is active and a new one is created, that one is autmatically marked as party.
- Made the active PartySize available as a parameter for formulas. I added this in our `replaceFormulaData` wrapper so that it's globally available anywhere a formula is evaluated.
- I fixed so that Summon Actions can actually evaluate rollData properties as part of updating the 2 SRD entries that use party size.

### DIFF
--- a/module/applications/sidebar/tabs/actorDirectory.mjs
+++ b/module/applications/sidebar/tabs/actorDirectory.mjs
@@ -103,9 +103,7 @@ export default class DhActorDirectory extends foundry.applications.sidebar.tabs.
                     if (!actor) throw new Error('Unexpected missing actor');
 
                     const currentActiveParty = game.actors.find(x => x.type === 'party' && x.system.active);
-                    if (currentActiveParty)
-                        await currentActiveParty.update({ 'system.active': false });
-
+                    await currentActiveParty?.update({ 'system.active': false });
                     await actor.update({ 'system.active': true });
                     ui.actors.render();
                 }
--- a/module/applications/sidebar/tabs/actorDirectory.mjs
+++ b/module/applications/sidebar/tabs/actorDirectory.mjs
@@ -46,50 +46,71 @@ export default class DhActorDirectory extends foundry.applications.sidebar.tabs.
 
     _getEntryContextOptions() {
         const options = super._getEntryContextOptions();
-        options.push({
-            name: 'DAGGERHEART.UI.Sidebar.actorDirectory.duplicateToNewTier',
-            icon: `<i class="fa-solid fa-arrow-trend-up" inert></i>`,
-            condition: li => {
-                const actor = game.actors.get(li.dataset.entryId);
-                return actor?.type === 'adversary' && actor.system.type !== 'social';
-            },
-            callback: async li => {
-                const actor = game.actors.get(li.dataset.entryId);
-                if (!actor) throw new Error('Unexpected missing actor');
+        options.push(
+            {
+                name: 'DAGGERHEART.UI.Sidebar.actorDirectory.duplicateToNewTier',
+                icon: `<i class="fa-solid fa-arrow-trend-up" inert></i>`,
+                condition: li => {
+                    const actor = game.actors.get(li.dataset.entryId);
+                    return actor?.type === 'adversary' && actor.system.type !== 'social';
+                },
+                callback: async li => {
+                    const actor = game.actors.get(li.dataset.entryId);
+                    if (!actor) throw new Error('Unexpected missing actor');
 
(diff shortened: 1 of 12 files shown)


module/applications/sidebar/tabs/actorDirectory.mjs
```diff
@@ -46,50 +46,71 @@ export default class DhActorDirectory extends foundry.applications.sidebar.tabs.
...
+                    const actor = game.actors.get(li.dataset.entryId);
+                    return actor && actor.type === 'party' && !actor.system.active;
+                },
+                callback: async li => {
+                    const actor = game.actors.get(li.dataset.entryId);
+                    if (!actor) throw new Error('Unexpected missing actor');
+
+                    const currentActiveParty = game.actors.find(x => x.type === 'party' && x.system.active);
+                    if (currentActiveParty)
+                        await currentActiveParty.update({ 'system.active': false });
```
CarlosFdez:
```suggestion
                    await currentActiveParty?.update({ 'system.active': false });
```

CarlosFdez:
Would active party not be better as a world setting actually? You can still make a getter for active

WBHarry:
> Would active party not be better as a world setting actually? You can still make a getter for active

Hm. In a way I like it, but at the same time it does mean that we should `_preDelete` check if the party was active to clear that setting, since it won't go automatically anymore.
I guess there is a pretty neat benefit on the other side of things as you never have to worry about unsetting a currently active party and ending up with multiple somehow since the setting specifices the only one. 

Second point wins out I think :+1:

WBHarry:
Changed it to work on `game.settings.get(CONFIG.DH.id, CONFIG.DH.SETTINGS.gameSettings.ActiveParty)`

CarlosFdez:
> setting

Even then, a setting pointing to an actor that doesn't exist wouldn't be a disaster. You kinda have to check that it exists anyways since there may be no actor at all.

module/data/actor/party.mjs
```diff
@@ -44,6 +48,23 @@ export default class DhParty extends BaseDataActor {
         }
     }
 
+    _onCreate(data, options, userId) {
+        super._onCreate(data, options, userId);
+
+        if (game.user.isActiveGM && !game.settings.get(CONFIG.DH.id, CONFIG.DH.SETTINGS.gameSettings.ActiveParty)) {
```
CarlosFdez:
```suggestion
        const activeParty = game.actors.get(game.settings.get(CONFIG.DH.id, CONFIG.DH.SETTINGS.gameSettings.ActiveParty);
        if (game.user.isActiveGM && !activeParty)) {
```

This should be more resilient to module shenaniganry and have a bit better intent maybe

module/documents/actor.mjs
```diff
@@ -125,6 +125,8 @@ export default class DhpActor extends Actor {
                 game.system.registeredTriggers.unregisterItemTriggers(token.actor.items);
             }
         }
+
+        if(this.system._preDelete() === false) return false;
```
CarlosFdez: This should be using super._preDelete() instead and passing in the args. And perhaps be earlier on so that item triggers aren't unregistered if deletion is cancelled. Though perhaps triggers should be unregistered in onDelete instead?

module/helpers/utils.mjs
```diff
@@ -189,7 +189,15 @@ export const getDeleteKeys = (property, innerProperty, innerPropertyDefaultValue
 
 // Fix on Foundry native formula replacement for DH
 const nativeReplaceFormulaData = Roll.replaceFormulaData;
-Roll.replaceFormulaData = function (formula, data = {}, { missing, warn = false } = {}) {
+Roll.replaceFormulaData = function (formula, baseData = {}, { missing, warn = false } = {}) {
```
CarlosFdez, line 192: Not for this PR but it might be good to see if its possible to replace the Roll class itself with a subclass that overrides this method.

module/systemRegistration/settings.mjs
```diff
@@ -189,4 +189,11 @@ const registerNonConfigSettings = () => {
         config: false,
         type: SpotlightTracker
     });
+
+    game.settings.register(CONFIG.DH.id, CONFIG.DH.SETTINGS.gameSettings.ActiveParty, {
+        scope: 'world',
+        config: false,
```
CarlosFdez, line 195: Funny enough crucible actually makes this setting configurable! PF2e does not though and presents other ways. No need to change here, just commenting on it.

CarlosFdez (approved):

PR #1804, opened 2026-04-15
[Refactor] Increase reuse of gold and inventory styling

CarlosFdez:
No new features or fixes, just a refactor to try to make it easier to add inventory and gold to new sheet types in the future.

`currency-section` was renamed to `gold-section` as well.

### DIFF
--- a/styles/less/sheets/actors/party/inventory.less
+++ b/styles/less/sheets/actors/party/inventory.less
@@ -3,51 +3,6 @@
 
 .application.sheet.daggerheart.actor.dh-style.party {
     .tab.inventory {
-        .search-section {
-            display: flex;
-            gap: 10px;
-            align-items: center;
-
-            .search-bar {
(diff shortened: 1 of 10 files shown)


styles/less/sheets/actors/party/inventory.less
```diff
@@ -3,51 +3,6 @@
...
-                    border: none;
-                    outline: 2px solid transparent;
-                    transition: all 0.3s ease;
-                    padding: 0 20px;
-
-                    &:hover {
-                        outline: 2px solid light-dark(@dark, @golden);
-                    }
-
-                    &:placeholder {
```
CarlosFdez, line 29: This never worked (needs an extra `:`) but I think it looks better without. Something to keep in mind tho.

WBHarry (approved):
Really neat :+1:

PR #1806, opened 2026-04-16
[Feature] Add hotkey for party sheet toggle

CarlosFdez:


module/systemRegistration/settings.mjs
```diff
@@ -52,6 +52,27 @@ export const registerKeyBindings = () => {
...
+
+    game.keybindings.register(CONFIG.DH.id, CONFIG.DH.SETTINGS.keybindings.partySheet, {
+        name: _loc('DAGGERHEART.SETTINGS.Keybindings.partySheet.name'),
+        hint: _loc('DAGGERHEART.SETTINGS.Keybindings.partySheet.hint'),
+        editable: [{ key: "KeyP" }],
+        onDown: () => {
+            const controlled = canvas.ready ? canvas.tokens.controlled : [];
+            const selectedParty = controlled.find((c) => c.actor?.type === 'party')?.actor;
+            const party = selectedParty ?? game.actors.party;
+            if (!party) return;
```
WBHarry, line 64: We should show an error message if no party could be found to open up rather than quietely returning.

WBHarry (changes requested):

module/systemRegistration/settings.mjs
CarlosFdez: Should it? I'm mimicking the behavior of the built in character sheet toggle, which doesn't error. in game.mjs `toggleCharacterSheet`.

WBHarry: Hm..the party button isn't shown unless there -is- a party and the button 100% works though, no? 🤔

CarlosFdez: Its a globally registered hotkey, not a button. Only reason I see we should error is that perhaps it'd improve discoverability if someone presses P when there is no active party or someone accidentally presses P and gets the error pop up and learns its a thing that way.

WBHarry: I meant the button in the character sheet. The reason that doesn't need an error is that if it's shown, it works.

If you bind the hotkey, it may do nothing, which isn't overly helpfull to the user. If it said something like: "No active party found, rightclick one to set it or go to the system settings.", then the user gets why and we don't get a #Support question 🙃

CarlosFdez: oh, I see. I wasn't referring to any buttons. I was referring to how the "C" hotkey doesn't error if there is no character, so I don't know yet why "P" should be different in this respect.

WBHarry: Well, it's fine without. If we do find a lot of players confused, then I guess we can change it at that time :+1:

WBHarry (approved):

PR #1807, opened 2026-04-16
[Feature] Show notification when invalid item types are added to actors

CarlosFdez:
Internally this refactors isItemValid() to use metadata for the inventory items and to exist on all actor types. The createDocuments() implementation now checks if all items were filtered out, and if so, it will warn about the first item type in the attempted list. This will get the right results in most cases, since batch adding is normally a module invoked thing, not drag/drop.

image

Also stealth fixes an error in the ancestry sheet on drop that's of minor importance.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -213,6 +213,9 @@
                     "headerTitle": "Adversary Reaction Roll"
                 }
             },
+            "Base": {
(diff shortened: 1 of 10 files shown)


WBHarry (approved):
Don't see any issues. Good localisation of the logic :+1:

PR #1808, opened 2026-04-16
[Feature] Make all item types quantifiable in the party

CarlosFdez:
This is experimental and should be tested, but if the general idea works, this is a step for loot and merchant support.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -213,6 +213,9 @@
                     "headerTitle": "Adversary Reaction Roll"
                 }
             },
+            "Base": {
+                "CannotAddType": "Cannot add {itemType} items to {actorType} actors."
+            },
             "Character": {
                 "advantageSources": {
                     "label": "Advantage Sources",
--- a/module/applications/dialogs/itemTransfer.mjs
+++ b/module/applications/dialogs/itemTransfer.mjs
@@ -38,13 +38,15 @@ export default class ItemTransferDialog extends HandlebarsApplicationMixin(Appli
         originActor ??= item?.actor;
         const homebrewKey = CONFIG.DH.SETTINGS.gameSettings.Homebrew;
         const currencySetting = game.settings.get(CONFIG.DH.id, homebrewKey).currency?.[currency] ?? null;
+        const max = item?.system.quantity ?? originActor.system.gold[currency] ?? 0;
 
         return {
             originActor,
             targetActor,
             itemImage: item?.img,
             currencyIcon: currencySetting?.icon,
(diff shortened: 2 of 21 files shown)


WBHarry:
I'm not sure we need to remove `metadata.isQuantifiable` from the items.
Currently the branch still has 
```
if (this.metadata.isQuantifiable)
            schema.quantity = new fields.NumberField({ integer: true, initial: 1, min: 0, required: true });
```
In `item/base.mjs`. I don't think that's neccessarily a bad thing, as it's a single place for modifying the schema rather than adding a `quantity` field to every model.

CarlosFdez:
The branch removed it, it was just still in the comment.

So the goal is to support weapons/armor being quantifiable everywhere except character sheets. This was your quote:

> Ah, I suppose Weapon/Armor doesn't actually set isQuantitiable. Which -is- sensible for Characters due to equip/unequip behavior. 
> Solution would probably be to set it to isQuantifiable, and somehow avoid the quantity input field on Characters in particular.

However, setting every inventory type item to isQuantifiable (and hiding the quantity on character) is the same as just having the quantity field on all of them. Instead I added a list of quantifiable types in the actor's metadata (effectively moving quantifiability to be an actor side concern), so on characters weapons/armor are not quantifiable, but on parties (and later merchants) they are.

If you have another idea I suppose let me know. This just seemed the cleanest version of your suggestion.

CarlosFdez:
Here's an example of how it works

https://github.com/user-attachments/assets/014988f7-5f93-4208-99a0-923d55586ca9

WBHarry:
No I agree with the implementation. I think that works :+1:

I was just saying that now we've added this field on all items that can have a quantity:
image

Previously this was done by `metadata.isQuantifiable`. Maybe it should be renamed to `metadata.hasQuantity`, but my point was that since it's a something present on so many items, it could be added to schema in `item/base` by looking at `metadata.hasQuantity` instead of adding it in different places.

CarlosFdez:
Ah that's what you mean. We can do that, though I think there is some misdirection in that solution. I actually tried to see if they all inherited from some inventory item first but nah.

We can make it use the already existing isInventoryItem instead.

WBHarry:
If you think it's more clear to add it directly on the individual data models, then that's fine with me.
In that case we should tidy away the `base/item.mjs` code though:
```
static defineSchema() {
   ...
   ...
   if (this.metadata.isQuantifiable)
            schema.quantity = new fields.NumberField({ integer: true, initial: 1, min: 0, required: true });
}
```

WBHarry (approved):
:+1:

PR #1810, opened 2026-04-17
[Fix] party sheet resources in light mode

CarlosFdez:
The header buttons still need fixes but I lack familiarity so I'm not touching that, only the parts that I broke.

image


WBHarry (approved):
:+1:

PR #1811, opened 2026-04-17
fix: 2 improvements across 2 files

tomaioo:
## Summary

fix: 2 improvements across 2 files

## Problem

**Severity**: `High` | **File**: `daggerheart.d.ts:L13`

Inside `declare global`, `const fromUuid = foundry.utils.fromUuid;` and `const fromUuidSync = foundry.utils.fromUuidSync;` use runtime initializers. In declaration files (`.d.ts`), ambient declarations must not include value initializers. This can cause TypeScript declaration parsing/emit failures.

## Solution

Replace with ambient typed declarations, e.g. `const fromUuid: typeof foundry.utils.fromUuid;` and `const fromUuidSync: typeof foundry.utils.fromUuidSync;`.

## Changes

- `daggerheart.d.ts` (modified)
- `gulpfile.js` (modified)

### DIFF
--- a/daggerheart.d.ts
+++ b/daggerheart.d.ts
@@ -10,8 +10,8 @@ declare global {
      * This class manages the registration and execution of hooked callback functions.
      */
     class Hooks extends foundry.helpers.Hooks {}
-    const fromUuid = foundry.utils.fromUuid;
-    const fromUuidSync = foundry.utils.fromUuidSync;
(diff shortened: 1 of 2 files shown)


WBHarry:
Two things:

- This is a no-AI project.
- We recently changed our contribution policy. We do not consider PRs opened without having made a discussion issue on what wants to be achieved and how first.

PR #1814, opened 2026-04-18
[Feature] Updates to inventory icons, context menus, action use, and observer visibility

CarlosFdez:
Refactor to facilitate item usage on more actor types, and fixes some bugs and quirks with icon and context menus as a result. It also fixes some deprecation warnings.

- The globe was turned into a pencil.
- Send to chat is now available in the party sheet
- More `<i>` elements are now inert. This is better for screen readers.
- Visibility of elements as well as the options available are now more resilient to the observer status across all sheets. It is now possible for observers to interact with the sheet, though still not edit anything.
- Actions and items no longer show that they are usable from the party sheet

image

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -3233,6 +3233,7 @@
             "Tooltip": {
                 "disableEffect": "Disable Effect",
                 "enableEffect": "Enable Effect",
+                "edit": "Edit",
                 "openItemWorld": "Open Item World",
                 "openActorWorld": "Open Actor World",
                 "sendToChat": "Send to Chat",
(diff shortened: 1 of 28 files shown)


CarlosFdez:
Clicking on the image icon will still do an action, but I want to handle that in a separate PR. That one is a bit more involved since I need to be sure actions are still editable despite not preparing them.

WBHarry (approved):
Great improvement :+1:

I don't think it's a big deal that you can click on an item action to perform the action - since the ActionList is still set to empty, so it'll just become a `ToChat`.

The two things I saw that could perhaps more pertinently be improved would be that an observer can roll an Attack roll from the sidebar, aswell as being able to do trait rolls from the shields.

---
I'm goind to accept this PR as is though, so the base improvement is in.

PR #1815, opened 2026-04-18
[Feature] Action Areas

WBHarry:
Added the ability for `Actions` to define `Areas`.
- An area can either just be a template used as a marker, or it can add on effects that will be applied to those that enter.

- Effects can now optionally define which token dispositions they should be applied to

---
Going to make a pass on the SRD to update it with areas once this PR is considered good 🙃


module/applications/sheets-configs/action-config.mjs
```diff
@@ -52,9 +56,20 @@ export default class DHActionConfig extends DHActionBaseConfig {
...
+
+        const { areaIndex, index } = button.dataset;     
+        let effectId = null;  
+        if (areaIndex !== undefined) {
+            effectId = this.action.area[areaIndex].effects[index];
+            const data = this.action.toObject();
+            data.area[areaIndex].effects.splice(index, 1);
+            this.constructor.updateForm.bind(this)(null, null, { object: foundry.utils.flattenObject(data) });
+        }
+        else {
```
CarlosFdez:
```suggestion
        } else { 
```

We should maybe consider eslint so that we don't accidentally mess up formatting.

module/applications/sheets-configs/action-config.mjs
```diff
@@ -52,9 +56,20 @@ export default class DHActionConfig extends DHActionBaseConfig {
...
         if (!this.action.effects) return;
-        const index = button.dataset.index,
+
+        const { areaIndex, index } = button.dataset;     
+        let effectId = null;  
+        if (areaIndex !== undefined) {
+            effectId = this.action.area[areaIndex].effects[index];
+            const data = this.action.toObject();
+            data.area[areaIndex].effects.splice(index, 1);
+            this.constructor.updateForm.bind(this)(null, null, { object: foundry.utils.flattenObject(data) });
```
CarlosFdez:
```suggestion
            this.constructor.updateForm.call(this, null, null, { object: foundry.utils.flattenObject(data) });
```

Test that I didn't mess it up of course. There's call and apply, and I think call is this + normal args and apply is this + args as array.

lang/en.json
```diff
@@ -109,6 +109,12 @@
                     "customFormula": "Custom Formula",
                     "formula": "Formula"
                 },
+                "area": {
+                    "sectionTitle": "Areas",
+                    "shape": "Shape",
+                    "size": "Size"
+
```
CarlosFdez:
```suggestion
```

module/applications/sheets-configs/action-config.mjs
```diff
@@ -52,9 +56,20 @@ export default class DHActionConfig extends DHActionBaseConfig {
...
+        }
+        else { 
             effectId = this.action.effects[index]._id;
-        this.constructor.removeElement.bind(this)(event, button);
+            this.constructor.removeElement.bind(this)(event, button);
```
CarlosFdez:
```suggestion
            this.constructor.removeElement.call(this, event, button);
```

module/applications/sheets-configs/action-settings-config.mjs
```diff
@@ -31,21 +31,38 @@ export default class DHActionSettingsConfig extends DHActionBaseConfig {
...
     static removeEffect(event, button) {
         if (!this.action.effects) return;
-        const index = button.dataset.index,
+        const { areaIndex, index } = button.dataset;
+        let effectId = null;
+        if (areaIndex !== undefined) {
+            effectId = this.action.area[areaIndex].effects[index];
+            const data = this.action.toObject();
+            data.area[areaIndex].effects.splice(index, 1);
+            this.constructor.updateForm.bind(this)(null, null, { object: foundry.utils.flattenObject(data) });
```
CarlosFdez:
```suggestion
            this.constructor.updateForm.call(this, null, null, { object: foundry.utils.flattenObject(data) });
```

module/applications/sheets-configs/action-config.mjs
WBHarry: Tested and it works all good. Updated :+1:

module/applications/sheets-configs/action-config.mjs
WBHarry: :+1:

module/applications/sheets-configs/action-settings-config.mjs
WBHarry: :+1:

module/data/activeEffect/baseEffect.mjs
```diff
@@ -93,6 +93,12 @@ export default class BaseEffect extends foundry.data.ActiveEffectTypeDataModel {
                     max: new fields.NumberField({ integer: true, label: 'DAGGERHEART.GENERAL.max' })
                 },
                 { nullable: true, initial: null }
+            ),
+            targetDispositions: new fields.SetField(
+                new fields.NumberField({
+                    choices: CONFIG.DH.GENERAL.simpleDispositions
+                }),
+                { label: 'Affected Dispositions' }
```
CarlosFdez: Missing translation

module/documents/chatMessage.mjs
```diff
@@ -137,6 +137,10 @@ export default class DhpChatMessage extends foundry.documents.ChatMessage {
             element.addEventListener('click', this.onApplyEffect.bind(this))
         );
 
+        html.querySelectorAll('.action-areas').forEach(element =>
+            element.addEventListener('click', this.onCreateAreas.bind(this))
+        );
+
```
```suggestion
        for (const element of html.querySelectorAll('.action-areas')) {
            element.addEventListener('click', this.onCreateAreas.bind(this));
        }

```

CarlosFdez:
I am moderately uncertain of naming the array of areas "area", and it comes to a head as a bit weird when doing the chat message (you are assigning area to areas). In a language like javascript, lack of type info means that confusing an array for a singular thing could be dangerous. But I am also hesitant to rename that, since action.area does flow a lot better, even if you have to do action.area.length 🙃

CarlosFdez:
That said, making the actions a dropdown stemming from a single button was actually a great idea.

module/canvas/placeables/regionLayer.mjs
```diff
@@ -95,4 +95,53 @@ export default class DhRegionLayer extends foundry.canvas.layers.RegionLayer {
...
+        const length = baseDistance;
+        const radius = length;
+
+        const shapeWidth = type === line.id ? 5 * dimensionConstant : type === rectangle.id ? length : undefined;
+        const shapeType = type === inFront.id ? cone.id : type;
+
+        const { width, height } = game.canvas.scene.dimensions;
+        return {
+            x: width / 2,
+            y: height / 2,
```
CarlosFdez:
```suggestion
        return {
            ...canvas.mousePosition,
```

For defaults, this may lead to a better result.

module/canvas/placeables/regionLayer.mjs
```diff
@@ -95,4 +95,53 @@ export default class DhRegionLayer extends foundry.canvas.layers.RegionLayer {
...
+            base: {
```
CarlosFdez: Maybe this should only be set if its an emanation. You could probably add a pick helper in module/helpers/utils.py below sortBy, and then pick ["width", "height", "x", "y", "shape"] from some parent token. That should handle certain cases.

A lot of complexity looks to be from handling many types of shapes. A base set that gets extended in a switch statement may simplify some things.

module/canvas/placeables/regionLayer.mjs
```diff
@@ -95,4 +95,53 @@ export default class DhRegionLayer extends foundry.canvas.layers.RegionLayer {
...
+                type: 'token',
+                x: 0,
+                y: 0,
+                width: 1,
+                height: 1,
+                shape: game.canvas.grid.isHexagonal ? CONST.TOKEN_SHAPES.ELLIPSE_1 : CONST.TOKEN_SHAPES.RECTANGLE_1
+            },
+            t: shapeType,
```
CarlosFdez: Removable I think.

module/documents/chatMessage.mjs
```diff
@@ -249,6 +253,55 @@ export default class DhpChatMessage extends foundry.documents.ChatMessage {
...
+        };
+
+        if (this.system.action.area.length === 1) createArea(this.system.action.area[0]);
+        else if (this.system.action.area.length > 1) {
+            /* Pop a selection. Possibly a context menu? */
+            new foundry.applications.ux.ContextMenu.implementation(
+                event.target,
+                '.action-areas',
+                this.system.action.area.map((area, index) => ({
+                    name: area.name,
```
```suggestion
                    label: area.name,
```

name is deprecated and will throw a warning. However the context menu threw a bug for me when I clicked the same button multiple times. If you can fix it go for it, but I have a suspicion that you may have to use game.tooltip.createLockedTooltip() instead.

https://github.com/user-attachments/assets/cae168b7-c4c4-456e-8f8f-406985a9830c

WBHarry: Got it fixed. It's a general improvement as this behavior would also have been seen on the SceneNavigation scene environment button - and just generally being wastefull to close/open the same dialog repeatidly.

Just tossed on this check in our `contextMenu.triggerContextMenu` method:
```js
if (ui.context?.selector === selector) return;
```

module/canvas/placeables/regionLayer.mjs
WBHarry: It was indeed removable :+1:

module/canvas/placeables/regionLayer.mjs
WBHarry: Works well :+1:

module/documents/chatMessage.mjs
WBHarry: Replaced both `name` and `callback`. They were actually both deprecated. Now it's `label` and `onClick` :+1:

WBHarry:
> I am moderately uncertain of naming the array of areas "area", and it comes to a head as a bit weird when doing the chat message (you are assigning area to areas). In a language like javascript, lack of type info means that confusing an array for a singular thing could be dangerous. But I am also hesitant to rename that, since action.area does flow a lot better, even if you have to do action.area.length 🙃

I initially named it `areas`, then I looked at the fact that we have `action.cost` and felt it stuck out.

I'm fine changing it over to areas if you think it's neater. Two wrongs don't make a right and all that 🙃

module/data/activeEffect/baseEffect.mjs
WBHarry: Fixed :+1:

module/canvas/placeables/regionLayer.mjs
WBHarry: I went with a switch case. It -does- make it a lot more readable:
```js
static getTemplateShape({ type, angle, range, direction } = {}) {
        const { line, rectangle, inFront, cone, circle, emanation } = CONFIG.DH.GENERAL.templateTypes;

        /* Length calculation */
        const { grid, distance } = CONFIG.Scene.documentClass.schema.fields.grid.fields;
        const sceneGridSize = canvas.scene?.grid.size ?? grid.size.initial;
        const sceneGridDistance = canvas.scene?.grid.distance ?? distance.getInitialValue();
        const dimensionConstant = sceneGridSize / sceneGridDistance;

        const rangeNumber = Number(range);
        const settings = canvas.scene?.rangeSettings;
        const baseDistance =
            (!Number.isNaN(rangeNumber) ? rangeNumber : settings ? settings[range] : 0) * dimensionConstant;

        const length = baseDistance;
        /*----*/

        const shapeData = {
            ...canvas.mousePosition,
            type: type,
            direction: direction ?? 0,
        };

        switch(type) {
            case rectangle.id:
                shapeData.width = length;
                shapeData.height = length;
                break;
            case line.id:
                shapeData.length = length;
                shapeData.width = 5 * dimensionConstant;
                break;
            case cone.id:
                shapeData.angle = angle ?? CONFIG.MeasuredTemplate.defaults.angle;
                shapeData.radius = length;
                break;
            case inFront.id:
                shapeData.angle = '180';
                shapeData.radius = length;
                shapeData.type = cone.id;
                break;
            case circle.id:
                shapeData.radius = length;
                break;
            case emanation.id:
                shapeData.radius = length;
                shapeData.base = {
                    type: 'token',
                    x: 0,
                    y: 0,
                    width: 1,
                    height: 1,
                    shape: game.canvas.grid.isHexagonal ? CONST.TOKEN_SHAPES.ELLIPSE_1 : CONST.TOKEN_SHAPES.RECTANGLE_1
                };
                break;
        }

        return shapeData;
    }
```

WBHarry:
Eh, I went ahead and changed it from `area` to `areas`. It's just more logical. 🤷

module/applications/dialogs/actionSelectionDialog.mjs
```diff
@@ -57,7 +57,7 @@ export default class ActionSelectionDialog extends HandlebarsApplicationMixin(Ap
 
     /** @inheritDoc */
     async _prepareContext(options) {
-        const actions = this.#item.system.actionsList.map(action => ({
+        const actions = this.#item.system.actions.map(action => ({
```
CarlosFdez: I prefer actions as a name, but actionList filters for usability (see weapon.mjs), whereas I think actions does not. There's probably better ways to handle it, but actionList is the with filter which is probably what we need to do in the dialog.

module/canvas/placeables/regionLayer.mjs
```diff
@@ -95,4 +95,61 @@ export default class DhRegionLayer extends foundry.canvas.layers.RegionLayer {
...
+                shapeData.angle = '180';
+                shapeData.radius = length;
+                shapeData.type = cone.id;
+                break;
+            case circle.id:
+                shapeData.radius = length;
+                break;
+            case emanation.id:
+                shapeData.radius = length;
+                shapeData.base = {
```
CarlosFdez, line 142: Should this receive the token that the card belongs to to use its dimensions, or are you handling that some other way? If we don't need the measurements of the origin token, then we could maybe omit this.

CarlosFdez (approved):

PR #1816, opened 2026-04-19
[Feature] Class Custom Levelup Options

WBHarry:
- Added in so that a class can declare custom levelup options on the different tiers.
- Added a levelup option for `dice` so that the index of used dice can be upgraded via levelup options. Primarily intended for the Combo Die, but it could be used as an option to improve your default advantage die or such.
- Moved all Dice rules in under `system.rules.dice`


CarlosFdez:
If possible I'd like to keep the API the same, and use faces instead of an index. What's the reason for the index?

EDIT: If its done in code, putting die sizes in an array and then doing `Math.min(DIE_SIZES.length - 1, DIE_SIZES.indexOf(blah) + 1)` should be sufficient. But if its an AE it might be more difficult.

WBHarry:
The reason for an index is that it's consistent and workable that an active effect does `+1` to get to the next dice size - rather than it being variable. (d10 -> d12 = 2, d12 -> d20 ) = 8)

CarlosFdez:
Is it intended for an AE to do the bump here? It looked like bumps are in code, and those are easy enough to handle.

WBHarry:
It's intended for AE's, but also with this PR intended to be able to raise/lower by 1 to support levelup options.

Ikraik:
@WBHarry we keep this open?

WBHarry:
Yup.

CarlosFdez (approved):
🙈

PR #1823, opened 2026-04-21
[Fix] Easier to see light theme header buttons

CarlosFdez:
Fixes the hard to see buttons on the party and item sheet header. I don't know what to call the color value, its kinda imitating the parchment color as a flat color, but for a one off light theme only thing, its maybe fine?

image


WBHarry (approved):
:+1:

PR #1824, opened 2026-04-21
[Feature] Redesign group roll dialog

CarlosFdez:
Missing approve/unapprove, modifier rendering, reset buttons, and localization. I am waiting for the reset/delete button to be decided on before I finish this PR.

image


WBHarry:
It's really neat when actually using it now :+1: :+1:

Advantage Dice and Disadvantage Dice are not currently getting a dice visual, as you can see on Rangeroni's roll
image

WBHarry:
I'm not a 100% about this one, but I did find that even with the `opacity: 0.4` on the unselected `Success`|`Failure` icons it was still not overwhelmingly clear which was selected due to the small icon size and background.

Would it be a good idea to add a border around the chosen one?
image

CarlosFdez:
How about a highlight?

WBHarry:
Could definately work! 
It's also possible it's just a me thing. It's not exactly major, just how I experienced it while clicking around.

CarlosFdez:
Issues should be a bit better resolved now. While the selected one should pop a bit more now, at the end of the day there is still the +/- indicator, so we do have to balance usability with visibility.

WBHarry (approved):

PR #1827, opened 2026-04-22
1794 - Include Item Damage

### LINKED ISSUE #1794: [Bug] Include Item Damage Not Working

WBHarry:
Fixes #1794 
Went ahead and made IncludeItemDamage work again, smashing the formulas together

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -119,8 +119,7 @@
                 "deleteTriggerContent": "Are you sure you want to delete the {trigger} trigger?",
                 "advantageState": "Advantage State",
                 "damageOnSave": "Damage on Save",
-                "useDefaultItemValues": "Use default Item values",
-                "itemDamageIsUsed": "Item Damage Is Used"
+                "useDefaultItemValues": "Use default Item values"
             },
             "RollField": {
                 "diceRolling": {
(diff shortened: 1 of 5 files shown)


CarlosFdez:
I won't poke at the sheet but everything else looks fine.

CarlosFdez (approved):

WBHarry:
Ooh, right. No I only removed an `#if` case in the template. Git just thinks I changed absolutely everything because of that :laughing:

PR #1832, opened 2026-04-24
[Fix] class domains when setting max domains to null or 0

CarlosFdez:
The settings say that 0 or null is unlimited, but 0 wasn't allowed nor was it unlimited if null.

### DIFF
--- a/module/data/actor/character.mjs
+++ b/module/data/actor/character.mjs
@@ -410,14 +410,11 @@ export default class DhCharacter extends DhCreature {
     }
 
     get loadoutSlot() {
-        const loadoutCount = this.domainCards.loadout?.length ?? 0,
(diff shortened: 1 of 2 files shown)


CarlosFdez:
Main thing I'm unsure of is that we could do the same for maxLoadout, but the display of max when rendered is easier to test if max has no value. Testing for is infinity in handlebars is trickier than checking for is true.

Could be that all uses of maxDomains should instead check if its truthy before restricting. I'll leave it to your judgement.

CarlosFdez:
I am realizing that it might be that the hint was only for maxLoadout and wasn't intended for class domains (unlimited hope/fear aren't available and don't make sense either), and I think I'm remembering that we were maybe thinking of removing the max class domains setting.

WBHarry (approved):
I think we may want to consider that yeah. But for now at least, this is better behavior at least :+1:

PR #1834, opened 2026-04-24
[Fix] Unstoppable Fix

WBHarry:
Fixed so that ORIGIN effect values are handled differently depending on if they target the item or the origin

### DIFF
--- a/module/documents/activeEffect.mjs
+++ b/module/documents/activeEffect.mjs
@@ -197,7 +197,7 @@ export default class DhActiveEffect extends foundry.documents.ActiveEffect {
         const stackingParsedValue = effect.system.stacking
             ? Roll.replaceFormulaData(key, { stacks: effect.system.stacking.value })
             : key;
-        const evalValue = itemAbleRollParse(stackingParsedValue, actorModel, itemModel);
+        const evalValue = this.effectSafeEval(itemAbleRollParse(stackingParsedValue, actorModel, itemModel));
         return evalValue ? evalValue : key;
     }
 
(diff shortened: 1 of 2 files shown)


module/documents/activeEffect.mjs
```diff
@@ -172,23 +172,31 @@ export default class DhActiveEffect extends foundry.documents.ActiveEffect {
...
                 const originEffect = foundry.utils.fromUuidSync(effect.origin);
                 const doc =
                     originEffect.parent?.parent instanceof game.system.api.documents.DhpActor
                         ? originEffect.parent
                         : originEffect.parent.parent;
-                if (doc) parseModel = doc;
+                if (doc) {
+                    if (doc instanceof game.system.api.documents.DHItem) {
+                        itemModel = doc;
+                        key = key.replaceAll(/origin\.@/gi, '');
```
CarlosFdez, line 187:
```suggestion
                        key = key.replaceAll(/^origin\.@/gi, '');
```

module/documents/activeEffect.mjs
```diff
@@ -172,23 +172,31 @@ export default class DhActiveEffect extends foundry.documents.ActiveEffect {
...
+                    } else {
+                        actorModel = doc;
+                        key = key.replaceAll(/origin\.@/gi, '@');
```
CarlosFdez, line 190:
```suggestion
                        key = key.replaceAll(/^origin\.@/gi, '@');
```

Test that it works, but this should prevent some midstring replace oopsies

PR #1835, opened 2026-04-24
[Fix] origin change values and prioritize item options

CarlosFdez:
Fixes unstoppable, supercedes https://github.com/Foundryborne/daggerheart/pull/1834

Since this change is much more foundational, I would like help testing this. In short, these are the changes:
1) When doing the origin replacement, we removed origin from the change key instead of the evaluated value. Now we replace origin on the value instead. This was also done by #1834 
2) We ensure that an actor is assigned to actor and item is assigned to item when passed to itemAbleRollParse. We prioritize the actor document over the actor system model as it's roll data is a superset.
3) itemAbleRollParse now prioritizes item roll data over actor props IF the item is an item document. Those are a superset of the actor roll data and shouldn't cause any regressions.

After this unstoppable works without any changes to the expression. In the future we can make it always retrieve the origin item if the origin item is on the same actor, though I don't know of a quick way to do that. effect.item is null for some reason as well. After that, ORIGIN would only be required for effects that come from other actors.

### DIFF
--- a/module/data/actor/base.mjs
+++ b/module/data/actor/base.mjs
@@ -1,6 +1,6 @@
 import DHBaseActorSettings from '../../applications/sheets/api/actor-setting.mjs';
 import DHItem from '../../documents/item.mjs';
-import { getScrollTextData } from '../../helpers/utils.mjs';
+import { createShallowProxy, getScrollTextData } from '../../helpers/utils.mjs';
 
 const fields = foundry.data.fields;
 
@@ -180,8 +180,7 @@ export default class BaseDataActor extends foundry.abstract.TypeDataModel {
      * @returns {object}
      */
     getRollData() {
-        const data = { ...this };
-        return data;
+        return createShallowProxy(this);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


WBHarry:
Tested it in various situations, and found a few cases that needed changing that I pushed up:
- I changed it so that `item/base.mjs` `getRollData` uses `Object.assign` to set up it's data. Didn't know it before, but a quirk of Object.assign compared to spread is that it assigns over all getter function computed values to the object. And in this case that's perfect as we don't care about the underlying function for this snapshot. 
(Fixes Bare Bones not having access to `@tier` (Actor) for it's caculation)

- I changed so that the item model in `document/activeEffect.mjs` will be `effect.parent` if there is no origin. This is to handle all effects on items that have `transfer=true`, as the item model will be the actual parent of the effect there.
(Fixes the `Paired` effect of a short sword getting a null item model and a value of blank string)

- I returned the use of `this.effectSafeEval` to wrap the final value in `activeEffect.mjs`. I noticed in my original PR that for some reason that wasn't used anymore. It's to concatenate any math expressions so that for example the `1+1` value of a short sword's paired action becomes `2` instead.

WBHarry (approved):
Easier said than done - but not everything seems to work 🥳

PR #1839, opened 2026-04-26
[Fix] companion hope levelup feature and default stress

CarlosFdez:
This is an issue in v13 as well. Can we remove the hope increase in the creature's data prep? Why does companion even have hope?

### DIFF
--- a/module/config/resourceConfig.mjs
+++ b/module/config/resourceConfig.mjs
@@ -60,12 +60,6 @@ const companionBaseResources = Object.freeze({
         max: 3,
         reverse: true,
         label: 'DAGGERHEART.GENERAL.stress'
-    },
-    hope: {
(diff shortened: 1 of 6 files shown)


WBHarry:
Companions have hope because the daggerheart designers decided that they have a levelup option that increases the amount of hope the player character has.
So we keep track of the amount of hope the `companion` has so that the player character can gain it.

CarlosFdez:
> Companions have hope because the daggerheart designers decided that they have a levelup option that increases the amount of hope the player character has. So we keep track of the amount of hope the `companion` has so that the player character can gain it.

Maybe it used to use that? I'm looking at character and it seems like it just checks level data directly now.

WBHarry (approved):
:+1:

PR #1842, opened 2026-04-27
[Fix] console noise when starting a tag team dialog or group roll

CarlosFdez:
This one has no major rush

### DIFF
--- a/daggerheart.mjs
+++ b/daggerheart.mjs
@@ -343,7 +343,7 @@ Hooks.on(CONFIG.DH.HOOKS.hooksConfig.tagTeamStart, async data => {
         if (!party) return;
 
         const TagTeamDialog = game.system.api.applications.dialogs.TagTeamDialog;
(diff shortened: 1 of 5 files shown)


CarlosFdez:
Added a bugfix for something that should never happen but better safe than sorry.

WBHarry:
This seems to break the styling when starting a TagTeam roll and moving from Initialization to the main view. 
Assumably, this is because something goes wacky with the `onRender` when doing things this new way.

CarlosFdez:
I had refactors in the branch that I didn't commit because I figured it would work but I guess it didn't. That's actually concerning because the only thing I did here was fetch an existing dialog if one existed rather than create a new one. Looks like the re-render is buggy and it was being hidden by creating a new dialog.

WBHarry (approved):
Hm. Yeah changing it to remove the `onRender` handling fixes so that the styling doesn't go wacko. And since it's an objectively better way of doings things..win win 🙃

PR #1843, opened 2026-04-27
[Fix] Contextmenu Delete In Compendiums

WBHarry:
Fixed so that the delete option is available in the compendium.

The other instances of `doc?.isOwner` are fine as is, since they end up blocking things that should not be done in the compendium, such as trying to roll damage.


CarlosFdez:
Oh, I see, "sync". Yeah that makes sense, if unfortunate.

CarlosFdez (approved):

PR #1846, opened 2026-04-30
[Fix] Improve Class-Subclass Linkage

WBHarry:
This is very much unfinished as things remain to be done, but the linkage idea is there to be considered.

Some remaining things:
- How should the editing of class identifier be done? My thinking is that it should not be a updateOnChange edit thing. Probably a popup dialog where you enter a new identifier and then have to hit save to avoid repeated changes along the way to make the next point work better.
- Changing a classes identifier should either clear all existing subclasses, or perhaps better to migrate them? 
- Handle deleting of classes/subclasses


WBHarry:
One thing of note regarding the idea of using `uuid` instead,

That would make it non-resilient towards users dragging classes from/to compendiums since that makes a whole new item with a new uuid

CarlosFdez:
if we use source uuid as well, then imported items would continue to link properly. But not if you then drag it to a new compendium. So it depends on how important that last bit is.

WBHarry:
For me, it's fine as such. The most important part is that we gain the functionality where subclasses can be linked to classes without actually editing the class (so you can add to SRD ones etc).

Users will forever have an issue where they will start to create stuff inworld, and then pull it over to the compendium and thus brick the class they setup with items and subclasses no longer being linked. It's a hard behavior to prevent without people being burnt a few times 🤔

CarlosFdez:
I continued it, but ran into a new problem, which is more convincing me that uuid treatAs is the appropriate solution.

I moved configuration to the subclass itself, since the idea is that a subclass attaches to a class, not the other way around. However, a subclass cannot know *which* class item is the one that it should show in the UI. To get around this issue I made it save the image and name at the time of the link, however there is no verification of a broken link anymore nor will it update when the original item updates.

Furthermore, updating the class a subclass links to doesn't rerender the class sheet, but we may need to solve this no matter what solution we do.

If you can find a way to get around these issues with the current structure that I pushed up we are go mode, but swapping to a uuid might be simpler. Give it a try first though.

CarlosFdez:
At this point I think it mostly works though give it some final passes.

PR #1847, opened 2026-04-30
Fix detection of negative modifiers

CarlosFdez:
Fixes certain things like the group roll and I imagine the tag team dialog


WBHarry (approved):
:+1:

PR #1853, opened 2026-05-01
Fix retrieving parent documents when the model is null

CarlosFdez:
Supercedes https://github.com/Foundryborne/daggerheart/pull/1848
Fixes whatever that fixes and also fixes bonuses leading to erroring chat messages


WBHarry (approved):
Huzzah :+1:

PR #1855, opened 2026-05-02
[Fix] Player Created Regions

WBHarry:
Creating template regions from the action ChatMessage button wasn't working for players before.
It failed because we were erronously always passing in a Behavior array object to the `placeRegion` method even when there were no effects. Passing in an empty array instead made it work.

- Doing this, I learned that players are -never- allowed to make a Region with an attached Behavior regardless of permissions. To get around that I added a gmEmit method flow to those cases. 
- Separated `emitGMUpdate` and `emitGMCreate`
- I changed so that we pass in the currently viewed level as a `levels` array to `placeRegion` so that the region only belongs to the current level instead of showing up on every single one.

### DIFF
--- a/module/documents/chatMessage.mjs
+++ b/module/documents/chatMessage.mjs
@@ -258,7 +258,7 @@ export default class DhpChatMessage extends foundry.documents.ChatMessage {
             const effects = selectedArea.effects.map(effect => this.system.action.item.effects.get(effect).uuid);
             const { shape: type, size: range } = selectedArea;
             const shapeData = CONFIG.Canvas.layers.regions.layerClass.getTemplateShape({ type, range });
-            
+
             const scene = game.scenes.get(game.user.viewedScene);
             const level = scene.levels.find(x => x.isView);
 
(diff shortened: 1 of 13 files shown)


module/documents/chatMessage.mjs
```diff
@@ -259,27 +259,48 @@ export default class DhpChatMessage extends foundry.documents.ChatMessage {
...
+                ownership: { default: CONST.DOCUMENT_OWNERSHIP_LEVELS.NONE },
+                visibility: CONST.REGION_VISIBILITY.ALWAYS
+            };
+            const placeRegion = data => {
+                canvas.regions.placeRegion(data, { create: true });
+            };
+
+            const needsGMExecution = effects.length > 0;
+
+            if (needsGMExecution && !game.user.isGM) {
```
```suggestion
            // Regions with effects must be placed by the GM
            if (effects.length > 0 && !game.user.isGM) {
```

When the variable is only used once I prefer proximity (so no line between) but its also short enough that I think you can just include it in there and add a comment

WBHarry: That's fair. My thinking was that the condition might get more complicated if we go ahead and add in Aura's potentially (although thinking of it, prolly not)

We can do this and just pull it out if that ever happens :1+:

CarlosFdez (approved):

PR #1861, opened 2026-05-03
[Feature] Add setting to hide party stats

CarlosFdez:
Some GMs are weird and prefer to hide these things.

image


WBHarry (approved):
:+1:

PR #1862, opened 2026-05-03
[Feature] Exclude unusable non-weapons in character sidebar

CarlosFdez:
Mostly just hides armor unless its special usable armor

image


WBHarry (approved):
Makes sense to me. I never saw much of a use for showing it :+1:

PR #1864, opened 2026-05-04
[UI] Adjust spacing of character hope

CarlosFdez:
Minor UI change. The hope section feels too wide, and I think that might be because all font awesome icons became fixed width in v14. 

Before
image

After
image

### DIFF
--- a/styles/less/sheets/actors/character/header.less
+++ b/styles/less/sheets/actors/character/header.less
@@ -168,7 +168,7 @@
             .domains-section {
                 position: relative;
                 display: flex;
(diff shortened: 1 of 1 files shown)


CarlosFdez:
I pulled up v13, and it seems v13 is in between the two. The edge of the box goes up to the dot, whereas in v14 it goes up to the end of the letter A in ancestry, and in v14 after my change it goes to the end of y.

image

WBHarry (approved):
It works for me. It's better than the current wide spacing :+1:

PR #1865, opened 2026-05-04
[UI] Rework styling of traits section of the character sheet

CarlosFdez:
This should look better when the character sheet is resized, and also opens the door for having multiple spellcasting traits shown at once. The large spellcasting mouseover thing is poof.

image


WBHarry (approved):
I think this is better, yup :+1:

PR #1866, opened 2026-05-04
[UI] Avoid shifting elements when mousing over a tags

CarlosFdez:
https://github.com/user-attachments/assets/76a7fea0-4181-4eb0-a91c-399187096847


WBHarry (approved):
:+1:

PR #1871, opened 2026-05-05
Implement another traits redesign

CarlosFdez:
This is a second crack at it based on molilo's attempt. Compare with what exists now and make a choice I guess.

image

image


WBHarry:
Hm. Something about the way the "shield" goes from thick to -extremely- thin and then widdening again looks off to me :thinking:

WBHarry:
Is the white shadow on agility because of hovering?

CarlosFdez:
For the shield, poke molilo I guess. As for the white glow, I think that's just something behind the window as it's translucent.

WBHarry (approved):
:+1:

PR #1875, opened 2026-05-09
[Feature] Additional Trait Reworks

WBHarry:
Some initial, tentative trait display changes. Some attempts at the Character Edit Sheet.
It's tough to make it work there due to the constrained space. The idea would be that it could be nice to use the new trait shield that the main sheet uses. Mess with it as desired.


CarlosFdez:
I have an idea for how we can summon a bit more space. Though once https://github.com/Foundryborne/daggerheart/pull/1937 is in we'll have to rebase this.

PR #1877, opened 2026-05-10
[Feature] 1876 - Adversary Disposition Split

### LINKED ISSUE #1876: [Feature] "Friendly adversaries" / GM NPCs
**Description**
As a GM, I sometimes use an NPC/adversary to add a party member. The prototype token will be linked to the actor and disposition is set to Friendly. Currently, the encounter tracker still shows these actors as adversaries and counts them against the BP of the encounter.

WBHarry:
Fixes #1876 
Introduced a adversary category split for BP calculation and CombatTracker display based on token disposition
image

----
The combat tracker isn't currently rerendering based on changes to tokenData related to the combatants. Not sure if that's something we want to delve into or not.

A possibility to enable quicker changes of disposition (thinking primarily neutral to hostile might be a usecase) could be to add a disposition change button in the combat tracker, since there's space there 🤷‍♂️ 
image

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -409,7 +409,7 @@
                 "openCountdowns": "Countdowns",
                 "adversaryCategories": {
                     "friendly": "Friendly",
-                    "neutral": "Neutral"
+                    "adversaries": "Adversaries"
                 }
             },
             "CompendiumBrowserSettings": {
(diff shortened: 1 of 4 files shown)


CarlosFdez (approved):
Disposition swapping mid fight is maybe not something we need to explicitly here, and could perhaps also be accomplished by stuff in the adversary sheet to show (and possibly edit) the disposition. Maybe a little icon somewhere (happy face, straight face, frowny face, I dunno). What we have here avoids the NPC having to explicitly remove friendly npcs from the tracker if added accidentally, so its already pretty good.

PR #1878, opened 2026-05-11
Slightly improve visibility of countdowns and fix mixing light and dark modes

CarlosFdez:
Hide whitespace changes in the diff for a better representation

image


WBHarry (approved):
Thinks it's a good idea. Checked it out and it seems fine to me :+1:

PR #1879, opened 2026-05-13
Implement new button styling

CarlosFdez:
In development button style. Please test it before we merge it in to get your opinion.

The following things are missing at this time:
1) Proper light mode styling
2) Converting stuff to CSS variables
3) Hover should maybe be puncher.


WBHarry:
Hm, personally I don't like the duller color more than the current strong one. 
image

image

CarlosFdez:
I kinda agree. I guess see if you warm up to it.

So the main reason for the design is that we have a lot of gold. For example, try converting the character features to glassy. The headers are bright gold, the buttons are bright gold, the borders are bright gold. This is basically the main issue with our current sheet right now. Technically, the trait headers are buttons that are using the duller gold look already.

However, I don't know what the hierarchy should be and what can give. Maybe a darker button color?

Ikraik:
PR is stale and can be handled later.

PR #1880, opened 2026-05-14
[Feature] ActionConfig - Granular Outcomes

WBHarry:


module/applications/sheets-configs/action-base-config.mjs
```diff
@@ -299,10 +389,15 @@ export default class DHActionBaseConfig extends DaggerheartSheet(ApplicationV2)
         this.constructor.updateForm.bind(this)(null, null, { object: foundry.utils.flattenObject(data) });
     }
 
-    static addDamage(_event) {
+    static addDamage(_event, button) {
```
CarlosFdez: This is a general thing in the codebase, but do you mind if we rename actions like these to `#onAddDamage`? It differentiates from methods invoked to do things, and its also the convention used by the core foundry codebase.

WBHarry: I suppose we could, although it does feel like becoming the villain of my own story.

In the case of Foundry, I detest the use of `#`. It's always the reason we end up needing to copy some 5 extra methods when we want to override just 1, as it's not possible to reference the private methods from outside.
In our case it's probably less of an issue though as our classes and dialogs aren't intended to be overwritten.

CarlosFdez: the important thing is starting with `on`. If using # makes you uncomfortable we can avoid it. Let me know.

The main benefit of using # is being able to make changes without being concerned about something being part of a public interface. I am in agreement when functionality can't be replicated.

PR #1881, opened 2026-05-14
Add hot reload configuration

CarlosFdez:
Enable HMR in the foundry application settings first 
image

The field I added to the deploy is untested but it should work in theory.

Snuck myself into the authors list on recommendation, though I think having a single Foundryborne author would have been best.


WBHarry (approved):
:+1:

PR #1883, opened 2026-05-15
[Fix] Errors when updating max hp and implement bar update animations

CarlosFdez:
Fixes a pretty nasty error that can occur if you reset a character, import a character, or manually change max hp. When you do so, the canvas becomes unusable and you're no longer able to interact with it until you reload.

However, since the break was because bars can now animate, I figured I'd implement segment animations while I fix it.

https://github.com/user-attachments/assets/96f4d130-904d-4bb8-9ba6-f61e64d1ed8b


WBHarry (approved):
Huzzah :+1:

PR #1884, opened 2026-05-15
[Fix] selecting multiclass and multiple sheet issues

CarlosFdez:
Fixes a few things. Feel free to omit anything that seems minor from patch notes, or condense stuff

* Multiclass items causing an error when selected. We missed a spot in the previous release for subclass link stuff
* Moved multiclass selection before domain card selection
* Fixed height of multiclass selection row, and handle 3 or more domains and subclasses. Updating styling as well.
* Fixed responsiveness of tab when resized, removed max size rules
* Fixed card rendering when multiple lines (image should now resize appropriately)

Here's an example of most style fixes happening all at the same time.

image


WBHarry (approved):
:+1:

PR #1885, opened 2026-05-15
[Fix] Compendium browser sometimes spawning behind other windows

CarlosFdez:
Internally foundry uses a combination of zIndex and frontApp to handle bring to front. Us persisting and setting the zIndex caused the two to go out of sync starting in v14, which is why sometimes compendium browsers when spawning behind will not go back to front when you click them, forcing you to click on a tertiary window.


WBHarry (approved):
:+1:

PR #1887, opened 2026-05-15
Move scrollbar definition to global styling

CarlosFdez:
This is more or less the way foundry does it as well. If we want to do it globally including for main foundry app windows, we can override `--color-scrollbar` instead.


WBHarry (approved):
:+1:

PR #1890, opened 2026-05-15
[Fix] creating domain cards on character sheet

CarlosFdez:
And a refactor of the names of the properties I passed through.


CarlosFdez:
It doesn't seem like the error would be in v13 glancing at the code.

WBHarry (approved):
Great, yeah I didn't think of the case where DomainCard creation would require system :+1:

PR #1892, opened 2026-05-16
Lift definition of alt outcomes up one step

CarlosFdez:
Damage is now part of the alt outcome data, with the intent for a single alt outcome field to contain all variations. In essence this allows the altOutcomes data structure to drive more of the logic for rendering its subelements, which you can see in the update to getOutcomeTabs() and the outcomes tab template.

getOutcomeTabs() should probably be no longer a static method but now an internal data prep function for the context, but one thing at a time.

AltOutcome is in its own file to avoid cyclical imports. Actions point to sheets in their metadata, which makes things rather tricky for import order. Any action sheet importing from the damage field file directly will cause an error on startup.


CarlosFdez:
This is a merge to Harry's PR, so the one that decides is Harry I imagine once he's happy to return to it.

WBHarry:
Yah, I'm going to check on this again sometime soon. I've gotten a bit more thoughtful about the whole thing due to a few things. But lifting it up like this so Effects and Damage can be under the same level should definately be a good thing regardless

PR #1894, opened 2026-05-16
[Fix] TokenHUD Fixes

WBHarry:
Unsure if it's since start of Foundry V14 or in one of the subsequent patches - but localization and layout had changed in foundry's tokenHUD.

- Fixed so that the tokenHUD matches the current V14 foundry layout. 
- Fixed so that the tooltips have the correct translations


CarlosFdez (approved):
Looks good to me, though it may be wise to limit the foundry minimum version to whichever one included the eye icon on the token hud.

WBHarry:
Hm..I think I was wrong in thinking it was a different icon. It's just the classical `fa-solid fe-eye-slash` kinda deal.
Must be the font-size or something that looked different to me.

PR #1895, opened 2026-05-16
[Feature] Add tier/type headers and filters to environments in browser

### LINKED ISSUE #1888: [Feature] Add filter support for environments in compendium
A compendium that can filter environments based on tier and type would be fantastic.

Additionally, being able to filter by features, damage type etc would be incredible.

CarlosFdez:
Closes https://github.com/Foundryborne/daggerheart/issues/1888

### DIFF
--- a/module/config/itemBrowserConfig.mjs
+++ b/module/config/itemBrowserConfig.mjs
@@ -70,6 +70,49 @@ export const typeConfig = {
             }
         ]
     },
+    environments: {
+        columns: [
+            {
+                key: 'system.tier',
+                label: 'DAGGERHEART.GENERAL.Tiers.singular'
+            },
+            {
+                key: 'system.type',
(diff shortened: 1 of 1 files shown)


WBHarry (changes requested):
If we're adding on a filter for Environment difficulty, then we should probably add a column with that information aswell. 

Otherwise all good. It was quite empty here before 🙃

CarlosFdez:
Should we? I copied it over from adversaries, that don't have it in the columns. There are more filters than columns for most of these. Are environments an exception for that bit of info you think?

WBHarry:
Huh, I didn't think of the way Adversaries are set up. In that case it's definately not something that needs to be done here.
If desired it can be handled holistically at another time :+1:

WBHarry (approved):

PR #1896, opened 2026-05-17
[Fix] Uses of font awesome and adjust browser filter spacing

CarlosFdez:
I don't know what the other uses are, but it does fix the header icons in the browser. I also adjusted the spacing of the filter elements to hopefully flow a bit better.

image


WBHarry (approved):
:+1:

PR #1897, opened 2026-05-17
Further increase sheet opacity

CarlosFdez:
This increases the opacity from 90 (56%) to d0 (80%). Slightly lowers blur as it now overcompensates a bit.

image


WBHarry (approved):
:+1:

PR #1899, opened 2026-05-17
[Fix] party rerenders from shields and weapons

CarlosFdez:
Equips/unequips as well as updates in shields did not reflect in the party sheet. This extends it to all item updates, though there is a debounce to prevent overloading the render requests.


WBHarry (approved):
:+1:

PR #1900, opened 2026-05-17
[Fix] journey end calculation and hope reduction when gaining scars

CarlosFdez:
Fixes 3 scars being considered as marking the last hope slot. Also fixes a crash when updating scars in some cases.


WBHarry (approved):
Bit of oroboros with the scars lowering the max they're compared against before yeah 😆  :+1:

PR #1903, opened 2026-05-19
[UI/UX] Remove item type line from domain cards

CarlosFdez:
Removes the "Domain Card" label from the item, reducing the total height. I was originally gonna reduce the image height, but then I found out I mathed out the aspect ratios from the book wrong, we are actually correct.

We can reject this or redirect the change if we think this will make it harder on newbies.

image

### DIFF
--- a/module/applications/sheets/items/domainCard.mjs
+++ b/module/applications/sheets/items/domainCard.mjs
@@ -4,7 +4,7 @@ export default class DomainCardSheet extends DHBaseItemSheet {
     /**@inheritdoc */
     static DEFAULT_OPTIONS = {
         classes: ['domain-card'],
-        position: { width: 450, height: 700 }
+        position: { width: 450, height: 680 }
     };
 
     /** @override */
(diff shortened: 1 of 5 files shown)


moliloo:
The idea to have the type write there is because its a pattern followed in the cards items to have a label that clearly say what is that card.
If you think its not necessary we can remove, just explaining to you whats the idea to have this label there

image

CarlosFdez:
Makes sense, though the actual cards don't observe that pattern. It seems like in the book, Spell/Ability play that role.

image
image
image

We could do something like `Grimoire Domain Card` if we should dummy proof more than the in person game (nothing wrong with that), but we might need to move Domain + Level to be entirely to the header for that to look good, and that requires some card design tweaks.

cptn-cosmo:
I feel like listing the domain type is clear enough to see that it is a domain card, especially the domain cards wont ever be shown in the same place as other items

Ikraik:
Current conflict there. Resolve and I'll review again.

CarlosFdez:
Minor updates are gonna happen here as part of something else, so lets close this for now.

PR #1904, opened 2026-05-19
Create sourcemaps for css files

CarlosFdez:
This lets us see the less file the rule we're looking at is actually defined in. We don't exclude js sourcemaps from the final build, so I didn't exclude the css ones either. 

EDIT: looks like while the sourcemaps are in the *build*, they're not in the *deployment*, so I guess its fine.


WBHarry (approved):
:+1:

PR #1905, opened 2026-05-19
[UI/UX] Add minor visual tweaks to daggerheart setting dialogs

CarlosFdez:
This only impacts stuff like the homebrew/metagame/variant rule settings, not everything in general.

Does the following minor tweaks (and moves some CSS to the proper place):
- Fixes file-picker styles
- Gave a height to form group labels, leading to consistent height between select rows and checkbox rows
- Styled standard-form and applied to metagame, including a padding on top level labels similar to how fieldset does. I am 50/50 on removing the "gap" from fieldset and making that fieldset a standard-form.

image

image

### DIFF
--- a/styles/less/global/elements.less
+++ b/styles/less/global/elements.less
@@ -7,12 +7,12 @@
     input[type='text'],
     input[type='number'],
     textarea,
+    file-picker,
     .input[contenteditable] {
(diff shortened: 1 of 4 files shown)


WBHarry (approved):
:+1:

PR #1906, opened 2026-05-20
[Fix] finishing levelup with a multiclass

CarlosFdez:
Bit of a double trouble here:

* The multiclass logic wasn't updated to use linkedClass or fetchSubclasses. Either worked, but linkedClass is faster
* Items added via level up were not receiving a compendiumSource, making subclass belongs to checks not function

Also, when leveling below the minimum possible multiclass level (5), it also removes the multiclass item. This allows cleanup. There should probably be better ways to remove char creation stuff stuff if not "bound". Perhaps a context menu.


WBHarry (approved):
Awesome. Definately need this fixed :+1:

PR #1907, opened 2026-05-20
[Feature] Add way to delete unbound character creation items

CarlosFdez:
Currently this only works on worlds with auto levelup disabled as well as multiclass items that are "left behind" after a failed levelup. But given the errors in circulation starting the last release, this may be handy.

Once we have a way to flag individual characters as manual vs automatic, this will see more use.

image

image


WBHarry:
Think this one will need a bit more carefull checking. After removing every piece of a multiclasses characters, I ended up as so:
image

WBHarry (approved):
:+1:

PR #1908, opened 2026-05-20
[Fix] Refocus armor slots when using arrow keys

CarlosFdez:
https://github.com/user-attachments/assets/4aaae43b-cb04-42e6-a6a0-268e4f74f877


WBHarry (approved):
:+1:

PR #1911, opened 2026-05-20
[Feature] Class/Multiclass Feature Split

WBHarry:
Changed so that Multiclass features and MulticlassSubclass features are displayed in separate fieldsets from the base class features in the character sheet.
We could either order it:

1) Class|Multiclass|Subclass|MulticlassSubclass

2) Class|Subclass|Multiclass|MulticlassSubclass

I went with **2** at the moment.
image

### DIFF
--- a/module/data/actor/character.mjs
+++ b/module/data/actor/character.mjs
@@ -588,11 +588,9 @@ export default class DhCharacter extends DhCreature {
             } else if (item.system.originItemType === CONFIG.DH.ITEM.featureTypes.community.id) {
                 communityFeatures.push(item);
             } else if (item.system.originItemType === CONFIG.DH.ITEM.featureTypes.class.id) {
-                if (item.system.multiclassOrigin) multiclassFeatures.push(item);
-                else classFeatures.push(item);
+                (item.system.multiclassOrigin ? multiclassFeatures : classFeatures).push(item);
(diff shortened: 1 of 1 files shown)


module/data/actor/character.mjs
```diff
@@ -586,9 +588,11 @@ export default class DhCharacter extends DhCreature {
             } else if (item.system.originItemType === CONFIG.DH.ITEM.featureTypes.community.id) {
                 communityFeatures.push(item);
             } else if (item.system.originItemType === CONFIG.DH.ITEM.featureTypes.class.id) {
-                classFeatures.push(item);
+                if (item.system.multiclassOrigin) multiclassFeatures.push(item);
+                else classFeatures.push(item);
```
CarlosFdez: These could maybe `(item.system.multiclassOrigin ? multiclassFeatures : classFeatures).push(item)` for all of them. If that feels obtuse, you can do `const collection = item.system.multiclassOrigin ? multiclassFeatures : classFeatures` and then add to that

WBHarry: Nah, I think that makes sense. Changed it to tertiaries :+1:

CarlosFdez (approved):
Looks good to me but I didn't test it. Merge if you're feeling lucky.

PR #1912, opened 2026-05-20
[Fix] 1898 - Reaction Roll Message Look

### LINKED ISSUE #1898: [Bug] Reaction Rolls lost their distinction in chat
**Describe the bug**
When rolling reaction rolls, the roll shows "with hope" or "with fear" in chat now.

**To Reproduce**
1. Roll a reaction roll
2. Look at the chat

Image

Image

**Setup Information:**
 - OS: Windows 11
 - Browser Chrome
 - Foundry Version v14.360
- System Version 2.2.5

WBHarry:
Fixes #1898 
Fixed so that Reaction roll chat messages do not mention 'withHope'/'withFear'. The logic was already there, it was just looking at the wrong variable path now.

### DIFF
--- a/templates/ui/chat/parts/roll-part.hbs
+++ b/templates/ui/chat/parts/roll-part.hbs
@@ -6,7 +6,7 @@
                 {{#if roll.isCritical}}
                     <span>{{localize "DAGGERHEART.GENERAL.criticalShort"}}</span>
                 {{else}}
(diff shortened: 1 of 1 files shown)


CarlosFdez:
I was encouraged to include on the aid rolls for group rolls because its nice, but it looks like we should maybe remove the with hope or with fear there too. Separate consideration though.

CarlosFdez (approved):

PR #1918, opened 2026-05-22
[Feature] V14 Cleanup

WBHarry:
- (TokenHUD) Changed to the updated Front/Back button icon. Added a new tokenHUD button from v14 we were missing: `LockToken`
- Fixed in the PrototypeTokenConfig token image preview added in v14
image

- Fixed translations in the combat tracker that had been moved
- Fixed so that our TokenSizes also set and sync the `token.depth` property
- Fixed so that our beastforms set `token.depth` when transforming and restores the tokens default depth when transforming back, same as it does for width/height.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -778,7 +778,9 @@
                 "title": "Group Roll"
             },
             "TokenConfig": {
-                "actorSizeUsed": "Actor size is set, determining the dimensions"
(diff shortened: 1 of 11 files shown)


CarlosFdez:
Conceptually looks good to me, and bunglebum is glorious. Untested tho, don't have time tonight.

CarlosFdez (approved):

PR #1923, opened 2026-05-24
[UI] Adjust actor sheet headers

CarlosFdez:
I tried to purty a bit and make the text more readable on lighter images. Also environment and party should be less tall now. 

Need a name for the whitish orangish white by now tho.

image

image


WBHarry (approved):
Good improvement 👍 
Unless there's some other input about it, then I think it's all good to bring in

PR #1928, opened 2026-05-24
[Feature] Full Rerolls

WBHarry:
Changed so that chat roll messages have a `Reroll Action` and `Reroll Damage` context menu option as applicable for doing a full reroll. This will take the place of the existing DamageRerollDialog.

Hopefully `DamageRoll.reroll` can be made a lot neater at some point, but with the current system I didn't figure something smarter out right now.


module/dice/dualityRoll.mjs
```diff
@@ -388,4 +384,58 @@ export default class DualityRoll extends D20Roll {
             if (currentCombatant?.actorId == config.data.id) ui.combat.setCombatantSpotlight(currentCombatant.id);
         }
     }
+
+    static updateResources(oldDuality, newDuality, actor) {
```
WBHarry: Moved from `DualtityDie` and made static so it can be used both here and in `DualityDie`

CarlosFdez: Could alternatively be a helper. Inside module/dice/helpers.js of course. It makes it more private though so ymmv.

WBHarry: Hm..it's not perfect, but I think I do prefer having it localized to something related to `Duality`. Even if we name a utility function "updateResourcesForDuality" or the like, I'm not sure it becomes more logical 🤔

CarlosFdez: that's why I suggested a dice specific helpers file in dice/helpers.js instead of using the global generic one.

WBHarry: Ah, that I can definately see, even if I can't read 🙃

module/data/chat-message/actorRoll.mjs
```diff
@@ -130,6 +130,40 @@ export default class DHActorRoll extends foundry.abstract.TypeDataModel {
...
+            partData.total = rerolled.total;
+            partData.roll = rerolled.toJSON();
+        }
+
+        if (game.modules.get('dice-so-nice')?.active) {
+            const rerollPromises = rerolls.map(roll => game.dice3d.showForRoll(roll, game.user, true));
+            await Promise.allSettled(rerollPromises);
+        } else {
+            foundry.audio.AudioHelper.play({ src: CONFIG.sounds.dice });
+        }
```
CarlosFdez: Might be time to make this a general helper. It looks like we do this quite often even before this PR.

Maybe `triggerChatRollFx`? Would take rolls as var args or an array, either or.

module/dice/dualityRoll.mjs
WBHarry: Moved it to `dice/helpers.mjs` :+1:

module/data/chat-message/actorRoll.mjs
WBHarry: Did this and updated all usecases. Caught a lot of cases where we didn't play an audio sound when DiceSoNice was not enabled aswell. :+1:

The cases where we do this logic for DualityDice have not been changed. They specifically modify the appearance of the dice, and it's dine differently in all the cases. 
I think that is better done as a separate task if desired.

CarlosFdez: Understandable

CarlosFdez:
Ran into an issue with rendering the dice. This is without dice so nice. Looks like the values are still fine though

https://github.com/user-attachments/assets/4cbaede6-5cb6-44c0-b6be-079f763bec86

WBHarry:
Fixed so the same denomination property used for the dice appearance still exists after reroll

CarlosFdez (approved):

PR #1929, opened 2026-05-25
[Fix] Advantage Dice Duplication

WBHarry:
I found an issue where adding an Advantage or Disadvantage dice has the AdvantageDice term in that `DhRoll` get duplicated upon rerolling.
It hasn't been an issue since advantage dice have never been possible to reroll before, but since that is about to get added, it's an issue beyond just technically being wrong.

- Fixed so that the Advantage/Disadvantage dice recieves the correct dice type, previously it ended up as just a `Die`.
- Changed `ApplyAdvantage` to work the same way as `ConstructBaseDice`, where the typing is corrected if the dice already exists, otherwise the dice are added to the `DhRoll` if it should be present. Previously it always added, even if the dice was already present.


CarlosFdez (approved):
Untested but I looked over it and conceptually it makes sense to me, and I see where the bug is. If you're uncertain about it I can test but this looks great to me.

PR #1930, opened 2026-05-25
[Feature] Show actor sheet when clicking actors in item browser

CarlosFdez:
Clicking on an actor in the browser now opens the sheet. There was nothing to expand for actors anyways. Should be easier to browse monsters and environments to throw against the party now.


WBHarry (approved):
Nice :+1:

PR #1932, opened 2026-05-26
[Refactor] Add emphatic color variable and set up scope overrides for core ones

CarlosFdez:
Some refactoring to use the core variables and get some variables that are more usage based. The "dark blue vs golden" main color for big shiny text was an easy one.

We don't override the color-text-primary, and use the base one as is, but I added a less variable for "type safety" since I suspect that's the reason we're using the other ones.


WBHarry (approved):
Doesn't seem to cause any issues, and a much nicer way to set the very common text colours :+1:

PR #1934, opened 2026-05-26
[Fix] usable checks on adversary features and locked compendium actors

CarlosFdez:
This fixes something i broke in v14 that works fine in v13.

image

This also makes all actions and items flagged as unusable (which means don't show roll icons) if inside of a locked compendium.


WBHarry (approved):
:+1:

PR #1935, opened 2026-05-26
[Fix] Disable character and adversary name inputs when read only

CarlosFdez:
Fixes adversary and characters names having editable inputs (and then throwing errors) in locked compendiums, and probably also fixes a similar issue for observers.


WBHarry (approved):
👍

PR #1937, opened 2026-05-26
[Refactor] Define more border and input color variables

CarlosFdez:
Also fixes a regression in main.

Some bonuses:
- Fixes the daggerheart sidebar menu having a golden border
- Slightly dialed back opacity increase. Very slightly
- Currency and biography inputs are no longer golden. That was kindof an unintentional side effect anyways


WBHarry (approved):
:+1:

PR #1939, opened 2026-05-27
[Fix] setting dialogs created from overriden light sheet actors

CarlosFdez:
Not a new bug. This occurs when the world is set to dark mode, you set a character to light mode, then open the character settings.


WBHarry (approved):
:+1:

PR #1940, opened 2026-05-28
[Fix] Preload class and subclass features for description

CarlosFdez:
Also snuck in a minor perf improvement for fetchSubclasses().

This hopefully allows classes and subclasses to get access to the full description when features are stored in compendiums. This isn't the ideal solution, but its the easy fast one.


WBHarry (approved):
:+1:

PR #1942, opened 2026-05-28
[Fix] tier adjustment on actions that use standard attack damage

CarlosFdez:
Treats actions with formulas that are an exact match for the attack's value without a description presence as intending to "use the standard attack". Something system wide to handle powered up attacks would be preferable, but this handles those cases in the meantime.

You can test with Cult Adept's Enervating Blast.


WBHarry (approved):
:+1:

PR #1943, opened 2026-05-29
Swap order of thresholds and resources in actor editor

CarlosFdez:
Minor thing that probably isn't worth a mention in the patch notes, but I tried to better match the statblock ordering to make it easier to data enter. Difficulty isn't as easy to move: there isn't space to put the token size where difficulty was. But something is better than nothing.

I was motivated from copying over homebrew statblocks and tripping over myself a bit.

image

image

Book ordering is 
tier -> type -> description -> motives -> difficulty -> thresholds -> hp/stress > attack -> experiences -> features

Current adversary order after this PR is
tier -> type -> difficulty -> description -> motives -> thresholds -> hp/stress -> attack -> experiences -> features

### DIFF
--- a/styles/less/sheets-settings/adversary-settings/sheet.less
+++ b/styles/less/sheets-settings/adversary-settings/sheet.less
@@ -7,7 +7,7 @@
         &.attack.active {
             display: flex;
             flex-direction: column;
-            gap: 16px;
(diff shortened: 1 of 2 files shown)


WBHarry (approved):
:+1:

PR #1945, opened 2026-05-29
[Refactor] Create index files for actor sheet styles

CarlosFdez:
Also renames actions.less to features.less to match the actual tab name. This gives more control of the imports to the actor sheet itself


WBHarry (approved):
:+1:

PR #1946, opened 2026-05-29
[UI] Make prosemirror button nicer

CarlosFdez:
Buttons are normally display flex, but foundry makes the styling block when hovered

Before
image

After
image


WBHarry (approved):
:+1:

PR #1948, opened 2026-05-30
[Fix] tag team roll results with mixed resource target types

CarlosFdez:
Easier to see the diff if whitespace is ignored. This also makes the sheet more resilient to errors in a few places but does not fix the styling quirks when your damage roll has stress.


WBHarry (approved):
Interesting that this never came up while testing it originally. Good fix :+1:

PR #1949, opened 2026-05-30
[Feature] NPC Actors

WBHarry:
Added the NPC Actor.
image

The separation of Adversary and NPC is solely on the basis that an NPC is a pure narrative device and does not have any direct combat functionality or resources. If you want to have an "NPC" with an attack, hit points, damage thresholds etc then you would still make an Adversary to servethat purpose
 
I have a further idea to enable optionally more detailed `Notes` for both Adversary and NPC, but that's a future task.


styles/less/sheets/index.less
```diff
@@ -8,6 +8,9 @@
 @import './actors/adversary/sidebar.less';
 @import './actors/adversary/effects.less';
 
+@import './actors/npc/sheet.less';
+@import './actors/npc/header.less';
```
CarlosFdez: I'm starting to move towards having index files. Since it lets the sheet itself control import order and makes it easier to add new ones if they're all in the same folder.

Ideally the index file would have the main sheet styles and then import any other stuff WITHIN the scss block rather than defining the top level style, but it is what it is.

WBHarry: I have been thinking about this when adding new stuff, but stuck to the established convention so far.

We can either use `index.less`, or work off the commonly found `sheet.less`. It sort feels like all other `less` files in the folder should be imported into that one.

system.json
```diff
@@ -244,6 +244,9 @@
             "adversary": {
                 "htmlFields": ["notes", "description"]
             },
+            "npc": {
+                "htmlFields": ["notes"]
```
CarlosFdez, line 248: Note description and motives? Adversary seems to be missing motives itself

WBHarry: Added in the missing `motives` and `description` :+1:

Hm. For adversary, it's actually defined as a stringfield - intended for just short things I guess. A bit inconsistent, but I felt a Prose editor worked well for the NPC:
`motivesAndTactics: new fields.StringField()`

WBHarry: image

CarlosFdez: Maybe we outta keep it as a string field and not open that genie unless we need to?

WBHarry: As in, keep it as a string field for Adversary - or you mean have it as a string field for NPC?

CarlosFdez: bothofum

WBHarry: Works for me. StringFields convert cleanly to HTMLFields if we do change our minds later
image

styles/less/sheets/index.less
WBHarry: Pulled in main with the `sheets/index.less` changes. Updates the NPC styles with an `index.less` in the same way :+1:

CarlosFdez: I think both, sheet.less should be index.less and then import the others. But I wasn't gonna fuss it.

styles/less/sheets/index.less
```diff
@@ -6,6 +6,7 @@
 @import './actors/character/index.less';
 @import './actors/companion/index.less';
 @import './actors/environment/index.less';
+@import './actors//npc/index.less';
```
CarlosFdez:
```suggestion
@import './actors/npc/index.less';
```

CarlosFdez (approved):

PR #1951, opened 2026-05-31
[Refactor] Replace scroll shadows with scroll animation timeline

CarlosFdez:
Replaces all masks that rely on padding with scroll timelines that only show the scroll shadows while you are scrolling. As a result, we now have top shadows and bottom shadows. Top shadows were often omitted since it required a top padding or it'd mess up the visuals. And a few minor padding and size fixes.

Scroll shadows are the fades that happen while scrolling:

image


WBHarry (approved):
Neat mixin solution :+1:

PR #1963, opened 2026-06-01
[Fix] ChatMessage Saves Pending Confirmation

WBHarry:
Changed so that a confirmation warning pops up when trying to apply effects to targets that haven't done their saves.
It's the same dialog shown when trying to apply damage to targets that haven't done their saves.

image

Possible we should look at prettying this one up, but that can always be done later 🤷‍♂️

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -711,9 +711,9 @@
             },
             "PendingReactionsDialog": {
                 "title": "Pending Reaction Rolls Found",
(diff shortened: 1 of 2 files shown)


CarlosFdez (approved):
looks fine to me. Bit of a copy paste but I suspect when refactored it'd instead be a new a template to try to make it look nicer, with both dialogs pointing to that same template.

WBHarry:
Yeah, I think this dialog could use a glowup to be sure. Maybe we can put molilo on the case 🧌

PR #1964, opened 2026-06-01
[Fix] Tooltip Color Scope

WBHarry:
Added DH style to tooltips to fix the coloring of ArmorManagement and ResourceManagement tooltip.
The new scoped color variables weren't hitting as it didn't match any of the rules.

**Prev**
image
image

**Changed**
image
image

Not 100% if it's right to tack it on in the `setStyle` method, but it doesn't seem to cause any trouble in other tooltips from what I've noticed.

### DIFF
--- a/module/applications/sheets/actors/character.mjs
+++ b/module/applications/sheets/actors/character.mjs
@@ -1045,7 +1045,7 @@ export default class CharacterSheet extends DHBaseActorSheet {
         game.tooltip.activate(target, {
             html,
             locked: true,
-            cssClass: 'bordered-tooltip',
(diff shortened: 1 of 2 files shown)


CarlosFdez:
Do we globally style all tooltips? I recall foundry had data-tooltip-class (which can be defined at any parent, even the outermost div), but if we style all tooltips regardless of source then this should be fine.

WBHarry:
Ah, yeah that's a lot better.

Removed the global styling and now specifically adding `dh-style` to ResourceManagement and ArmorManagement :+1:

PR #1968, opened 2026-06-03
[Fix] Allow removing empty string domains

CarlosFdez:
Doesn't need a changelog mention. This allows removal of erroneous domains added programmatically, like what I accidentally did while testing domain additions once upon a time.


WBHarry (approved):
:+1:

Issue #734, opened 2025-08-09
[Feature] Homebrew Settings - Gold


cptn-cosmo:
In the Homebrew settings for gold, allow setting the start amount new characters start with.

Ikraik:
Marking this for v1.1 as we can manually workaround this, but it is a good enhancement feature.

Ikraik:
I think it would be necessary to convert the actors' gold fields into a data model to better manage conversions, perhaps - Joaquin

CarlosFdez:
As its been done, I'm closing the issue!

PR #1969, opened 2026-06-04
[Feature] Configurable starting gold amounts in Homebrew settings

TrezzeIohan:
Hi, guys! This is my first time contributing to a project; any feedback is appreciated.

This pull request aims to solve issue https://github.com/Foundryborne/daggerheart/issues/734

Added `initialAmount` field to each currency type (coins, handfuls, bags, chests) in the Homebrew settings schema. Defaults match book values (0, 1, 0, 0).

- Homebrew.mjs: added `initialAmount` NumberField per currency; `_initializeSource` coerces empty submissions to 0; `refreshConfig()` syncs values to `CONFIG.DH.RESOURCE.character.initialCurrency`
- resourceConfig.mjs: added mutable `initialCurrency` object on `character` export
- actorField.mjs: added `CharacterGoldField` subclass that reads initial values from config at actor creation time
- character.mjs: switched from `GoldField` to `CharacterGoldField`
- settings.hbs: restructured currency section to CSS Grid with column headers ("Quantity Name" / "Starting Amount") instead of per-field inline labels
- settings.less: added `.currency-rows` grid styles
- en.json: added `quantityName` and `initialAmount` localisation keys

Screenshot from 2026-06-04 00-17-41

### DIFF
--- a/module/data/fields/actorField.mjs
+++ b/module/data/fields/actorField.mjs
@@ -126,4 +126,15 @@ class GoldField extends fields.SchemaField {
     }
 }
 
-export { attributeField, ResourcesField, GoldField, stressDamageReductionRule, bonusField };
+class CharacterGoldField extends GoldField {
+    getInitialValue(options) {
+        const base = super.getInitialValue(options);
+        const initialCurrency = CONFIG.DH.RESOURCE.character.initialCurrency;
+        for (const type of ['coins', 'handfuls', 'bags', 'chests']) {
+            base[type] = initialCurrency[type];
+        }
+        return base;
+    }
+}
+
+export { attributeField, ResourcesField, GoldField, CharacterGoldField, stressDamageReductionRule, bonusField };
--- a/module/data/fields/actorField.mjs
+++ b/module/data/fields/actorField.mjs
@@ -116,25 +116,29 @@ class ResourcesField extends fields.TypedObjectField {
 }
 
 class GoldField extends fields.SchemaField {
-    constructor() {
-        super({
-            coins: new fields.NumberField({ initial: 0, integer: true }),
-            handfuls: new fields.NumberField({ initial: 1, integer: true }),
-            bags: new fields.NumberField({ initial: 0, integer: true }),
-            chests: new fields.NumberField({ initial: 0, integer: true })
-        });
+    constructor({ initial, ...options } = {}) {
+        super(
+            {
+                coins: new fields.NumberField({ initial: 0, integer: true }),
+                handfuls: new fields.NumberField({ initial: 1, integer: true }),
+                bags: new fields.NumberField({ initial: 0, integer: true }),
+                chests: new fields.NumberField({ initial: 0, integer: true })
+            },
+            options
+        );
+        this._initialCallback = initial;
     }
-}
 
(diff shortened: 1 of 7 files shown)


CarlosFdez:
Hello, thank you for working on the PR. 

The idea seems sound, though I have a few observations.

1) Disabled denominations will still get the initial value, which will be revealed when un-disabling it. The initial value should maybe set the values for disabled ones to zero somewhere.
2) Instead of making a CharacterGoldField subclass, you could add args to the GoldField to pass to the second parameter of the schemafield during initialization. You could then set an initial value callback from within the character itself.
3) Instead of outright giving up the schema based form input stuff (since formGroup doesn't work), you could use formInput instead. That should work for the text inputs.
4) Quantity Name doesn't feel like the right name for the column. Perhaps `Denomination` or `Category Name`? That's what its called in the book (both terms are used at various points on page 104 of the CRB).

TrezzeIohan:
@CarlosFdez Thank you so much for your feedback!

I've addressed all the points you raised. Please feel free to leave any further comments.
Also, if this PR gets accepted, do you have any other issues you'd recommend I work on?


[Fix] Address PR review feedback for starting gold feature

- Disabled denominations now receive 0 at character creation
- Replace CharacterGoldField subclass with GoldField options callback;
  initial value logic now lives in character.mjs
- Use formInput instead of plain text inputs for denomination labels
- Rename "Quantity Name" column header to "Denomination"
- Remove initialCurrency from resourceConfig and currency sync from
  refreshConfig (no longer needed)

module/data/fields/actorField.mjs
```diff
@@ -116,13 +116,28 @@ class ResourcesField extends fields.TypedObjectField {
...
 
 class GoldField extends fields.SchemaField {
-    constructor() {
-        super({
-            coins: new fields.NumberField({ initial: 0, integer: true }),
-            handfuls: new fields.NumberField({ initial: 1, integer: true }),
-            bags: new fields.NumberField({ initial: 0, integer: true }),
-            chests: new fields.NumberField({ initial: 0, integer: true })
-        });
+    constructor({ initial, ...options } = {}) {
```
CarlosFdez: initial should be one of the properties supported on `DataField`. I'd imagine that would work as is without needing to set up a callback.

TrezzeIohan: I believe I've addressed your feedback. Could you take another look?

CarlosFdez: That was it yeah, thanks!

cptn-cosmo:
@TrezzeIohan It is a bit difficult recommending issues currently. If you haven't already, we'd recommend joining the projects discord server to discuss open issues and how to best approach them. If you have already feel free to ping me (same username) so you can get access to the dev channels.

CarlosFdez:
Code looks fine to me but I'm too busy to test rn.

WBHarry (approved):
Tested and checked. Is all good. Bringing in :+1:

module/data/fields/actorField.mjs
TrezzeIohan: Thank you for your time, Carlos

PR #1970, opened 2026-06-04
[Fix] visual quirk with blur in unfocused countdown

CarlosFdez:
Rather than blur being applied on header/content separately, its applied on the outermost div. This avoids the weird "dividing line" in the blur that occurs when an image is under it.


CarlosFdez:
I snuck in a minor fix for icon only mode and some refactors to make it scale with font size. I also replace id for data-countdown, since id is meant to be used once only on the entire DOM. I also replaced the events with the onX names we talked about.

WBHarry (approved):
Scrollbar begone 🪄 

Nice standardisation of action names and reducing to a single Id value.

PR #1972, opened 2026-06-05
[Fix] rerolling when hope/fear automation is enabled

CarlosFdez:
Another one for the I wish typescript wasn't a pain to use with foundry pile.


WBHarry (approved):
:+1:

PR #1973, opened 2026-06-05
[Fix] several issues with inline damage

CarlosFdez:
Fixes a few issues with inline damage rolls and does some quality of life

- Fix issue where clicking inline damage rolls didn't work
- Fix issue where scaling adversaries with inline damage didn't work
- Add QoL so that first param of damage roll is optional, and fix issue with optional first params. Now you can do `@Damage[3d6|type:physical]`


WBHarry (approved):
:+1:

PR #1975, opened 2026-06-05
[Housekeeping] Replace prettier with stylistic, improve types, and add no-undef rule

CarlosFdez:
Most changes are indentation based, since stylistic has different preferences. The main benefit is that now if you set up the symlinks so that foundry is linked up right, the types should more or less function and undeclared stuff should warn immediately.

No unused vars has a lot of violations so I didn't turn it on yet.

Other changes are:
- Removed some unused code that had a lot of errors in it
- Removed a function that was overwritten later in the fire
- Replaced reject() for return on the party


WBHarry:
Hm..yah, I tested the current branch, setting up the symlink and all.

The autocompletion of `CONFIG` is definately there which is great.
I think you prolly forgot to push up a package change from `eslint/prettier` to stylistic tho. I don't see any reference of it in `package.json`, and I'm definately getting that console junk and stuff.

`npm run lint` gives 1250 errors and a tun of garbo blank space in the terminal :upside_down_face:

CarlosFdez:
Did you run npm ci? Interesting that you're getting garbo blank space without prettier though. Maybe the garbo blank space is by something else? It went away when I replaced prettier. Oh well, I prefer how stylistic handles nested ternaries better. The 2 space indentation was annoying.

WBHarry (approved):
Huh, well I'll be.
I was not aware of the difference between `npm install` and `npm ci` until I looked it up now. I ran install before - but clearly the reinstall was needed.

It works. And I generally like non-forced indentation, so it's all good with me 👍

CarlosFdez:
Oh it still indents, but with a multiple of 4 instead of sometimes indenting by 2 because there's a `? ` in the expression.

I will get something in to make dev setup neater easier, but with this hopefully one type of error will cleanly not exist anymore, and now the jsdoc type stuff can actually work.

PR #1977, opened 2026-06-05
[Housekeeping] Styles Index:ification

WBHarry:
Went through all of the styles folders and made sure every folder uses an `index.less` file all the way down.

In cases where a folder only had a single `less` file such as `sheet.less`, I still added a `index.less`. I'm thinking in separation of concerns. If we do add new `less` files in this folder, then that shouldn't be a concern of the upper folders. Like this, we just import the new style sheet to the closest `index.less`.


styles/less/dialog/group-roll-dialog/sheet.less
```diff
@@ -0,0 +1,48 @@
+.daggerheart.dialog.dh-style.views.group-roll-dialog {
+    .window-content {
+        h1 {
```
WBHarry, line 3: This is the copied contents of `_common.less` that I removed. It was only used here, so felt sensible to just put the styles here instead.

styles/daggerheart.less
```diff
@@ -1,19 +1,12 @@
...
 @import './less/global/index.less';
-
 @import './less/ui/index.less';
-
 @import './less/ux/index.less';
 
 @import '../build/tagify.css';
 
-@import './less/utils/mixin.less';
+
```
CarlosFdez:
```suggestion
```

CarlosFdez:
YAY!

CarlosFdez (approved):

PR #1980, opened 2026-06-06
[Housekeeping] Fancier dev setup

CarlosFdez:
createSymlink is now a bit more dummy proof, and also can put the project repo symlink in the foundry systems folder.


.env.example
```diff
@@ -1,2 +1,2 @@
 FOUNDRY_MAIN_PATH=/path/to/foundry/resources/app/main.js
-FOUNDRY_DATA_PATH=/path/to/foundry/data
\ No newline at end of file
+FOUNDRY_DATA_PATH=/path/to/foundry/
```
CarlosFdez, line 2: there is no change in the actual data. It never worked with the actual data folder, only the pre data folder. This was misleading so I changed it.

tools/create-symlink.mjs
```diff
@@ -1,47 +1,70 @@
...
 }
 
-// Language files
-try {
-    await fs.promises.symlink(path.join(fileRoot, 'public', 'lang'), path.join('foundry', 'lang'));
-} catch (e) {
-    if (e.code !== 'EEXIST') throw e;
-}
+await createFoundrySymlink();
+console.log();
```
WBHarry: Empty console log.

WBHarry (approved):
:+1:

tools/create-symlink.mjs
CarlosFdez: its to add a space between them. Should it have a comment? Or we add an extra newline to the final console logs?

WBHarry: Ah, well it's not exactly a forward facing feature, so that should be fine then. I just had a snap reaction to a seemably meaningless log 🙃

CarlosFdez:
```suggestion
console.log(); // Add empty newline
```

PR #1982, opened 2026-06-07
[Fix] Companion Rolls

WBHarry:
- Fixed so that companion rolls don't error. 
- Fixed title of companion action rolls. The formUpdate for `d20rollDialog` was being run onClose despite being flagged not to do so. Fixed by `event.preventDefault` - the update was changing the `config.title`.

### DIFF
--- a/module/dice/d20Roll.mjs
+++ b/module/dice/d20Roll.mjs
@@ -100,16 +100,13 @@ export default class D20Roll extends DHRoll {
 
         this.options.roll.modifiers = this.applyBaseBonus();
 
-        const actorExperiences = this.options.roll.companionRoll
-            ? (this.options.data?.companion?.system?.experiences ?? {})
-            : (this.options.data.system?.experiences ?? {});
-        this.options.experiences?.forEach(m => {
-            if (actorExperiences[m])
-                this.options.roll.modifiers.push({
-                    label: actorExperiences[m].name,
-                    value: actorExperiences[m].value
-                });
-        });
+        let actorExperiences = this.options.data.system?.experiences ?? {};
+        if (this.options.roll.companionRoll) {
+            const companion = typeof this.options.data.companion === 'string' ? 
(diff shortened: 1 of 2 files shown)


module/dice/d20Roll.mjs
```diff
@@ -101,7 +101,7 @@ export default class D20Roll extends DHRoll {
         this.options.roll.modifiers = this.applyBaseBonus();
 
         const actorExperiences = this.options.roll.companionRoll
-            ? (this.options.data?.companion?.system.experiences ?? {})
+            ? (this.options.data?.companion?.system?.experiences ?? {})
```
CarlosFdez: what was companion that it wasn't null but didn't have a system?

WBHarry: Think it's lost at this point after the roll is finished and the chat message is made. Companion = undefined
We could look into trying to keep the companion data alive - but at the very least I don't think it serves any purpose atm once the roll has been made.

WBHarry: Oh right, nvm. Yeah companion.system was an ID for some reason I think. I'll check again

WBHarry: image

Makes sense. The `companion` field is a foreignUUID field, and the data at this point is objectified, so it doesn't do the lookup:
`companion: new ForeignDocumentUUIDField({ type: 'Actor', nullable: true, initial: null }),`

WBHarry: I guess it could be more proper to check what type it is, and do a `fromUUIDSync` as required.

CarlosFdez: huh, yeah, looks like that would be the correct fix.

WBHarry: Changed over to that logic :+1:

CarlosFdez:
Works well and looks good to me!

CarlosFdez (approved):

PR #1986, opened 2026-06-07
[Feature] System Styled GamePause

WBHarry:
Experimented with hope/fear pulsed text and the DhCompatible logo

https://github.com/user-attachments/assets/0d5b61f6-61d2-458c-92f6-c3b825c1ae58

### DIFF
--- a/styles/less/utils/colors.less
+++ b/styles/less/utils/colors.less
@@ -234,20 +234,6 @@
     }
 }
 
-@keyframes pause-pulse {
-    0% {
-        filter: drop-shadow(0 0 5px @secondary-blue);    
-    }
-
-    50% {
-        filter: drop-shadow(0 0 5px @golden);
-    }
-
-    100% {
-        filter: drop-shadow(0 0 5px @secondary-blue);   
-    }
-}
-
 // LESS variable versions of core foundry color variables
 @color-border: var(--color-border);
 @color-fieldset-border: var(--color-fieldset-border);
--- a/styles/less/utils/colors.less
+++ b/styles/less/utils/colors.less
@@ -234,6 +234,20 @@
     }
 }
 
+@keyframes pause-pulse {
+    0% {
+        filter: drop-shadow(0 0 5px @secondary-blue);    
+    }
+
+    50% {
+        filter: drop-shadow(0 0 5px @golden);
+    }
+
+    100% {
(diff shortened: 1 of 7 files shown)


CarlosFdez:
IMO its an improvement (not bad). Perhaps you could add some identifying css class or id so that the default class or a different subclass would not have the styling. That's the only thing I've noticed as maybe an issue.

WBHarry:
I changed it around so that we only do changes if there are no registered renderGamePause events - IE, any modules seeking to modify how gamePause works. Tested it with the `Alternative Pause Icon` module, and it works as normal with this change.

```js
export default class DhGamePause extends foundry.applications.ui.GamePause {
    async _onRender(context, options) {
        await super._onRender(context, options);
    
        /* Avoid altering the styling if a module has subscribed to the renderGamePause hook  */
        if (!Hooks.events.renderGamePause?.length) {
            this.element.classList.add('dh-style');
        }
    }

    /** @override */
        const context = await super._prepareContext(options);

        /* Avoid altering the gamepause context if a module has subscribed to the renderGamePause hook  */
        if (!Hooks.events.renderGamePause?.length) {
            context.spin = options.spin ?? false;
            context.icon = options.icon ?? 'systems/daggerheart/assets/logos/compatible_with_DH_logos-10.png';
        }

        return context;
    }
}
```

CarlosFdez:
It'd be nice if that module made its own subclass instead so we can use DEFULT_OPTIONS but it is what it is.

styles/less/utils/colors.less
```diff
@@ -234,6 +234,20 @@
     }
 }
 
+@keyframes pause-pulse {
```
CarlosFdez: Actually, since the colors used here are hard coded and this isn't a mixin, this could go in game pause I think. This file is for color variables for theming.

CarlosFdez (changes requested):

styles/less/utils/colors.less
WBHarry: That's fair :+1:

WBHarry:
> It'd be nice if that module made its own subclass instead so we can use DEFULT_OPTIONS but it is what it is.

Yeah, I ended up settling on this solution (which I stole from DnD5E after realisign that DEFAULT_OPTIONS doesn't work out with modules.

PR #1987, opened 2026-06-09
Fix linking world item subclasses

### LINKED ISSUE #1974: [Bug] Manually added subclasses for world item classes fail to validate correctly
**Describe the bug**
When a class and its subclass were both made as world items, you cannot manually add the class and then the subclass. The subclass will not recognize the manually added class as belonging to it.

**Setup Information:**
 - Daggerheart 2.3.1 on Foundry v14

CarlosFdez:
Closes https://github.com/Foundryborne/daggerheart/issues/1974

### DIFF
--- a/module/applications/characterCreation/characterCreation.mjs
+++ b/module/applications/characterCreation/characterCreation.mjs
@@ -441,9 +441,8 @@ export default class DhCharacterCreation extends HandlebarsApplicationMixin(Appl
 
         if (type === 'subclasses') {
             const classItem = this.setup.class;
-            const uuid = classItem?._stats.compendiumSource ?? classItem?.uuid;
             presets.filter = {
-                'system.linkedClass': { key: 'system.linkedClass', value: uuid }
+                'system.linkedClass': { key: 'system.linkedClass', value: classItem?.sourceUuid }
             };
         }
 
--- a/module/applications/sheets/actors/character.mjs
+++ b/module/applications/sheets/actors/character.mjs
@@ -356,7 +356,7 @@ export default class CharacterSheet extends DHBaseActorSheet {
 
             const levelups = Object.values(actor.system.levelData?.levelups) ?? [];
             const uuid = item.uuid;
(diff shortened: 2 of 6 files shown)


module/applications/characterCreation/characterCreation.mjs
```diff
@@ -441,9 +441,8 @@ export default class DhCharacterCreation extends HandlebarsApplicationMixin(Appl
 
         if (type === 'subclasses') {
             const classItem = this.setup.class;
-            const uuid = classItem?._stats.compendiumSource ?? classItem?.uuid;
             presets.filter = {
-                'system.linkedClass': { key: 'system.linkedClass', value: uuid }
+                'system.linkedClass': { key: 'system.linkedClass', value: classItem?.sourceUuid }
```
CarlosFdez, line 445: This is the part that I can't really figure out a solution for yet. So it turns out classItem is a plain object, but classItem.system is the DHClass type data model, and classItem.system.parent is a DHItem. I don't really understand why that's happening, but its preventing classItem?.sourceUuid from working. I could use some help investigating.

CarlosFdez: Maybe just a weird index thing?

WBHarry: Found the issue.
On drop at line 608 in `characterCreation.mjs` we are spreading the original drop object, which makes it a plain object. 

```js
else if (item.type === 'class' && event.target.closest('.class-card')) {
   this.setup.class = { ...item, effects: Array.from(item.effects).map(x => x.toObject()), uuid: item.uuid };
```
I can't see a reason why we need to do that however. The effects are ran through `toObject` later anyhow, and the uuid is already present.
My suggestions is to just change it to `this.setup.class = item;`. Seems to work fine like that. :+1:

CarlosFdez: Thanks for digging, I'll do just that.

WBHarry (approved):
Solid getter solution. Tested character creation and manual creation :+1:

PR #1989, opened 2026-06-09
[Housekeeping] Fix check for FOUNDRY_DATA_PATH

CarlosFdez:
Avoids names that end in Data from tripping the check, and makes the check actually do something.


WBHarry (approved):
:+1:

PR #1990, opened 2026-06-10
[Feature] CountdownsType Split

WBHarry:
- Added Short|Long toggle buttons in the Countdown UI to let users select which they want to have visible.
- When the value of a countdown is changed, an animation is shown on the element.


CarlosFdez:
So far, looks like icons only view has a regression. I'll check the rest soon.

image

WBHarry:
> So far, looks like icons only view has a regression. I'll check the rest soon.
> 

Good catch, another `#each` loop was introduced, and `../` was being used to access `iconOnly`. Changed to `@root.iconOnly` to fix :+1:

lang/en.json
```diff
@@ -1170,6 +1169,10 @@
                 "fear": "Fear",
                 "spotlight": "Spotlight"
             },
+            "CountdownBaseType": {
```
CarlosFdez: It looks like elsewhere you refer to this as the countdown type, but the label is CountdownBaseType. I'm wondering if instead CountdownType should be renamed to something else, like CountdownAdvancement.

WBHarry: Hm. We could do that, advancement is definately a better name for this:
```js
"CountdownType": {
    "actionRoll": "Action Roll",
    "characterAttack": "Character Attack",
    "characterSpotlight": "Character Spotlight",
    "custom": "Custom",
    "fear": "Fear",
    "spotlight": "Spotlight"
},
```

It would however mean that we have to do a migration.

WBHarry: Unless we do it everywhere -except- the DB fields ofc.

CarlosFdez: Would it? They're localization fields.

Looks like these are stored in `countdown.progress.type`. The normal types are stored in `countdown.type`.

WBHarry: Ah, you're right. The database stuff is actually named well, just these ones. I'll fix em up

CarlosFdez: Could be CountdownProgressType I guess.

WBHarry: I went with `CountdownProgressType` so it matches well with `CONFIG.DH.GENERAL.countdownProgressionTypes` :+1:

module/config/generalConfig.mjs
```diff
@@ -970,6 +959,19 @@ export const countdownAppMode = {
     iconOnly: 'icon-only'
 };
 
+export const countdownType = {
```
CarlosFdez: Is the convention for plural?

module/applications/ui/countdowns.mjs
```diff
@@ -1,5 +1,6 @@
 import { waitForDiceSoNice } from '../../helpers/utils.mjs';
 import { emitGMUpdate, GMUpdateEvent, RefreshType, socketEvent } from '../../systemRegistration/socket.mjs';
+import { SYSTEM } from './../../../module/config/system.mjs';
```
CarlosFdez: Uses of this should probably be CONFIG.DH instead to match up with everything else rather than directly import it.

module/applications/ui/countdowns.mjs
```diff
@@ -32,6 +39,7 @@ export default class DhCountdowns extends HandlebarsApplicationMixin(Application
         },
         actions: {
             toggleViewMode: DhCountdowns.#onToggleViewMode,
+            onToggleCountdownTypes: DhCountdowns.#onToggleCountdownTypes,
```
CarlosFdez: should likely be `toggleCountdownTypes: DhCountdowns.#onToggleCountdownTypes`. The convention is action -> onAction.

module/config/generalConfig.mjs
WBHarry: Hm, I think it's not perfectly adhered to either way currently actually. But it's at least mostly running by plural it seems like, so I changed it over to that :+1:

module/applications/ui/countdowns.mjs
WBHarry: Agreed!
The reason I had it imported is because in an earlier itteration I was setting up a `const` outside of the class, so it was set up before CONFIG was available.

Now that's no longer the case though, so I removed the import.

module/applications/ui/countdowns.mjs
WBHarry: That makes sense. Gonna program that into muh nogging 🧠 
Fixed it.

CarlosFdez (approved):

PR #1991, opened 2026-06-10
[Housekeeping] V14.364 Update

WBHarry:
- Raised compatability to foundry-364. 
- Swapped to new resourceBar color methodology introduced in this stable release. Foundry's `_getBarColors` reads from `CONFIG.Token.barConfig`. Should make it really easy for a module to swap colors by just overriding the config.


CarlosFdez:
Does _getBarColors() exist earlier? If it doesn't, we may have to bump min.

WBHarry:
Ah, that's true. `_getBarColors` was introduced in this version.

PR #1992, opened 2026-06-10
Bulk add items during character creation

CarlosFdez:
Also includes duplicateSource for character creation, which we need to start setting now for world items (but this doesn't set it for levelups).

### DIFF
--- a/module/applications/characterCreation/characterCreation.mjs
+++ b/module/applications/characterCreation/characterCreation.mjs
@@ -525,17 +525,16 @@ export default class DhCharacterCreation extends HandlebarsApplicationMixin(Appl
         };
 
         // Inner function to create the base item data
-        // @todo double check the data/baseData splat. That doesn't work for system, but its the previous code
         async function createEmbeddedItemData(baseData) {
             const uuid = baseData.uuid ?? baseData._uuid
             const data = baseData instanceof Item ? baseData : await foundry.utils.fromUuid(baseData.uuid) ?? baseData;
             return {
-                ...data,
                 ...baseData,
                 id: data.id,
                 uuid: uuid,
                 _uuid: uuid,
                 effects: data.effects?.map(effect => effect.toObject()),
+                flags: baseData.flags ?? data.flags,
                 _stats: {
                     ...data._stats,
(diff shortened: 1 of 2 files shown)


module/applications/characterCreation/characterCreation.mjs
```diff
@@ -524,37 +524,53 @@ export default class DhCharacterCreation extends HandlebarsApplicationMixin(Appl
...
-        await createEmbeddedItemWithEffects(this.character, this.setup.subclass);
-        await createEmbeddedItemsWithEffects(this.character, Object.values(this.setup.domainCards));
+        // Inner function to create the base item data
+        // @todo double check the data/baseData splat. That doesn't work for system, but its the previous code
+        async function createEmbeddedItemData(baseData) {
+            const uuid = baseData.uuid ?? baseData._uuid
+            const data = baseData instanceof Item ? baseData : await foundry.utils.fromUuid(baseData.uuid) ?? baseData;
+            return {
+                ...data,
+                ...baseData,
```
CarlosFdez, line 532: I would like a second look at this. So this splat doesn't work for system data, it will always favor baseData. In the one case where baseData isn't an item (mixed ancestries), we already deep merge anyways. So I'm thinking of killing the fromUuid retrieval and just putting baseData here directly.

CarlosFdez: hmm, I guess primary ancestry effects. The splat is useless though outside of perhaps flags. I'm not sure why we pass uuid in though.

WBHarry: Passing uuid in was probably done since we did convert to simple objects before - and that has a tendency to lose the uuid property.

CarlosFdez: Passing in _uuid and uuid both seems odd tho. uuid also isn't actually stored data, its derived by a getter. It won't be there if you use toObject() for example.

CarlosFdez: I removed the data splat, but it looks like its still necessary because of effects.

WBHarry (approved):
Doesn't seem to cause any issues, and it's a lot better.

Tested character creation and leveling up with subclass improvement aswell as multiclassing :+1:

PR #1994, opened 2026-06-11
[Refactor] Inherit font family body on most standard components

CarlosFdez:
End result should be the same hopefully. The stuff I left behind aren't in use right now so I can't really verify them.

I'm assuming the reason the fonts file got flagged is end of line changes. I set up git attributes a while back to force conversion to ln instead of crln.

### DIFF
--- a/styles/less/utils/fonts.less
+++ b/styles/less/utils/fonts.less
@@ -19,7 +19,3 @@
 .dh-style {
     .dh-typography();
 }
-
-.dh-style fieldset {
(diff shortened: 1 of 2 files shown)


WBHarry:
I'm not spotting any immediate oddities in font family across text, so I'm going to say this is all good. We'll just have to correct later if we -do- find some place where it didn't cover everything.

WBHarry (approved):

PR #1996, opened 2026-06-12
[Fix] On Drop Compendium ActiveEffect

WBHarry:
Fixed so that it works to drop a compendium ActiveEffect on a sheet. 
Since it's form a compendium it wasn't the full data, just the UUID - now we fetch the data in that case.


CarlosFdez:
Looks like upstream instead does

```
const documentClass = foundry.utils.getDocumentClass(data.type);
const document = await documentClass.fromDropData(data);
```

but even then, you're doing document.data here (as data.data) which shouldn't work. This code also seems to no longer be necessary for item sheets, and is only for actor config now, which could maybe get its own implementation? I'd have to dig more into it.

PR #1997, opened 2026-06-13
[Fix] dragging AE from compendium or within same actor

CarlosFdez:
Supercedes https://github.com/Foundryborne/daggerheart/pull/1996 and fixes drag dropping AEs within the same sheet.

As of v14, item's now support drag/drop events, and `_onDropActiveEffect` checks for same actor. Using item.toDragData() handles pretty much most of the checks, though we may need to use `Document.fromDropData()` in some places I'm not aware of yet.

This won't work for dropping AEs on non-actor-item sheets. Actor/item setting apps don't accept AEs anyways, and action sheets don't support drag dropped AEs regardless of this change (but perhaps they should?).

### DIFF
--- a/module/applications/sheets/actors/adversary.mjs
+++ b/module/applications/sheets/actors/adversary.mjs
@@ -186,15 +186,6 @@ export default class AdversarySheet extends DHBaseActorSheet {
         });
     }
 
-    /** @inheritdoc */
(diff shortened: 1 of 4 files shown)


WBHarry (approved):
Great solution. All the red is beautiful 🥳

Issue #1981, opened 2026-06-07
[Feature] Loadout should be sortable


itteerde:
**Description**
Sorting works for loot, but not the Loadout. Would be great to have sorting there, too.

TrezzeIohan:
PR under Review.

Ikraik:
Marking this closed as this is fixed and merged

PR #2004, opened 2026-06-14
[Feature] Enable sorting for Loadout domain cards

TrezzeIohan:
## Summary

This PR aims to solve issue https://github.com/Foundryborne/daggerheart/issues/1981: 
Enable drag-and-drop sorting functionality for Domain Cards in the Loadout and Vault sections by reusing the exact pattern already implemented for Loot items.

## Changes

- Add \`data-item-id\` and \`draggable=\"true\"\` attributes to domain card template (\`domain-card-item.hbs\`)
- Add sorting by Foundry's \`sort\` property in the \`domainCards\` getter (\`character.mjs\`)

## Test plan

- [ ] Open character sheet with domain cards in Loadout
- [ ] Drag a card to reorder it — should move to new position
- [ ] Reload application — order should persist
- [ ] Move card to Vault and drag to reorder there too
- [ ] Verify Loot sorting still works (no regressions)

### DIFF
--- a/module/data/actor/character.mjs
+++ b/module/data/actor/character.mjs
@@ -411,8 +411,8 @@ export default class DhCharacter extends DhCreature {
 
     get domainCards() {
         const domainCards = this.parent.items.filter(x => x.type === 'domainCard');
-        const loadout = domainCards.filter(x => !x.system.inVault);
-        const vault = domainCards.filter(x => x.system.inVault);
+        const loadout = domainCards.filter(x => !x.system.inVault).sort((a, b) => a.sort - b.sort);
+        const vault = domainCards.filter(x => x.system.inVault).sort((a, b) => a.sort - b.sort);
 
         return {
             loadout: loadout,
--- a/templates/sheets/global/partials/domain-card-item.hbs
+++ b/templates/sheets/global/partials/domain-card-item.hbs
@@ -1,5 +1,5 @@
-<li class="card-item" data-item-uuid="{{item.uuid}}" data-type="domainCard">
-    <img src="{{item.img}}" data-action="useItem" class="card-img" />
+<li class="card-item" data-item-id="{{item.id}}" data-item-uuid="{{item.uuid}}" data-type="domainCard">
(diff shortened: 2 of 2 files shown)


TrezzeIohan:
The following images contain the behavior of inventory items being sortable (eg: being able to be dragged to another position) and the same behaviour being applicable to the loadout tab. First 2 are from the inventory tab, last ones are from the loadout tab.

Inventory Sorting 1
Inventory Sorting 2
Loadout Sorting 1
Loadout Sorting 2

CarlosFdez:
Looks fine, though did you manage to test portrait view? Does it work to drag not the image? If so, the draggable might not be doing anything.

I'm eventually planning on rewriting how sorting works here for multiple reasons (including for faster vault/recall), but this is fine and dandy for now.

TrezzeIohan:
@CarlosFdez I just tested it.
[Screencast from 14-06-2026 23:25:37.webm](https://github.com/user-attachments/assets/6fe3242b-09eb-4b51-8906-5dedc404cc6e)
[Screencast from 14-06-2026 23:25:37.webm](https://github.com/user-attachments/assets/0d766fa8-ed03-4ed4-8b93-723839365513)

First one tests portrait, the second tests dragging not from the image, but from the body of the list item.

Was that what you were asking or did you mean something else?

CarlosFdez:
Almost, what about name in list mode? Does drag/drop still work without draggable?

CarlosFdez:
ok, tested it, turns out that template isn't used in list mode. However it looks like if you drag from the name or from the buttons that pop up, dragging doesn't work (at least in chrome). Draggable should probably be set on the li. Otherwise works good.

EDIT: the reason why it shouldn't be just image is because someone may start dragging from the bottom of the image, which causes the controls to pop up and block the attempt to drag/drop.

TrezzeIohan:
Sorry for the wait
[Dragging From Texts From Card View.webm](https://github.com/user-attachments/assets/a8717552-0595-4ed6-8c4f-b8d6fd5e029d)

CarlosFdez (approved):
Looks good to me, thanks for the PR.

PR #2005, opened 2026-06-14
[Housekeeping] Enable no unused vars and add more types

CarlosFdez:
Should hopefully prevent a certain class of error. It looks like the fateTypeLabel wasn't needed.

This allows item.actor and weapon.parent types to function as well. I added a vscode settings file despite that not usually being recommended so that by default find by reference will go to the file, not the types file.


WBHarry (approved):
:+1: No unused is definately nice.

PR #2006, opened 2026-06-15
[Fix] Weapon Effects Rules Correction

WBHarry:
Changed so that weapon effects are not just applied to attack/damage of the weapon itself

I think I had missunderstood the following rule
image
I thought they meant that the rules of a feature only applies to the weapon. 
Now I think they mean that the rules on the weapon (that can be whatever) are unique for that weapon (IE, read the rules on the weapon and apply them, you won't find them in any general rules section)

- This solves the issue that a weapon couldn't have a homebrew effect giving you +1 to spellcasting.
- It also means that something like `Reliable` on a rapier or `Powerfull` on a Greatstaff will apply the bonus to hit and damage respectively to spells being cast and such.

### DIFF
--- a/module/data/action/baseAction.mjs
+++ b/module/data/action/baseAction.mjs
@@ -343,13 +343,11 @@ export default class DHBaseAction extends ActionMixin(foundry.abstract.DataModel
 
         return Array.from(await actor.allApplicableEffects({ noTransferArmor: true, noSelfArmor: true })).filter(
             effect => {
-                /* Effects on weapons only ever apply for the weapon itself */
-                if (effect.parent.type === 'weapon') {
-                    /* Unless they're secondary - then they apply only to other primary weapons */
-                    if (effect.parent.system.secondary) {
-                        if (effectParent?.type !== 'weapon' || effectParent?.system.secondary) return false;
-                    } else if (effectParent?.id !== effect.parent.id) return false;
-                }
+                /* Effects on secondary weapons only ever apply for the primary weapon */
+                const isSecondaryWeaponEffect = effect.parent.type === 'weapon' && effect.parent.system.secondary;
+                const parentIsNotPrimaryWeapon = effectParent?.type !== 'weapon' || effectParent?.system.secondary;
+                if (isSecondaryWeaponEffect && parentIsNotPrimaryWeapon)
+                    return false;
 
                 return !effect.isSuppressed;
             }


CarlosFdez (dismissed):
Can't test but it conceptually looks good to me, merge if you're feeling lucky.

CarlosFdez:
Though actually, reading later, it does say this in a later sentence:
> A weapon’s features only apply to that weapon; you can’t apply its features to a different weapon.

So bare minimum, secondary weapon should be blocked from benefiting from primary weapon effects as it is another weapon. Perhaps one could state that spells are an exception, but there's some problematic interactions, like with this weapon and any attack spell. Double Fireball sounds like a fun one.
image

Perhaps this should be handled with a toggle or some kind of applies to? To allow specifics overrides general? Or some form of conditionals for AE applications in general, with weapon effects getting a different default.

WBHarry:
It is just really strange. Taken in totality it is saying:
image

1) Features apply only to the weapon.
2) Examples are changing the stats of the character (uh, see # 1 ??)
3) They cannot be applied to other weapons

It feels to me like they are for some reason really weirdly saying that the rules written on the weapon only effect that weapon - which should be blatantly obvious that it doesn't effect some random domain card or whatnot.

---
I'm really not sure. But playing with the system, it does feel like weapon effects should apply to spells. Weapons are more or less completly useless to spellcasters without it. And it certainly feels correct that the mage is channeling magic through their staff, sword or whatnot.

The example of the Hammer doesn't feel bad to me at all, actually. You are after all nuking yourself for 100% of the damage aswell (attack against melee target, so you're in the blast of the fireball and the hammer effect) 🤷

WBHarry:
We do have toggle functionality as far as bonuses to rolls and damage goes.
Reliable and other such things show up as toggles in the Roll Dialogs

cptn-cosmo:
I don't think weapons are useless for Spellcasters, they are basically like cantrips, you can always deal damage with a weapon if you are out of resources. And RAW any effect on a weapon only applies to that weapon, not other spells or abilities?

WBHarry:
I dunno. Two different spellcasters over 2 campaigns. Both used their weapon once, then never again after realising they had spells that are better and do not meaningfully (or at all) cost resources.

WBHarry:
> And RAW any effect on a weapon only applies to that weapon, not other spells or abilities?

It's so weird, because this is blatantly not true. A greatsword gives you -1 to evasion. That's not exactly something that's applied to the weapon. Likewise with the Saint armor or what it's called giving you -1 to all traits.

If we were to go strictly by that definition, then wielding a spear (-1 to finesse) wouldn't actually penalise you at all, because that penality would only be applied to the weapon (which doesn't use finesse) and not to making random trait rolls or abilities I guess? 🤔

cptn-cosmo:
See for me Hammer of Exota would apply only to attacks made with it, not to spells :thinking:

WBHarry:
(As a sidenote, spells -are- attacks by the rules 👅)

Yeah, idk. Ultimately it's just unclear. 
After all, domain cards that have you making attacks do not specify that they use the weapon. It's just logical. They just say to make an attack, and everyone is choosing to do so with their weapon.
Likewise, spells -are- attacks. I'm not sure there's a reason you coudn't choose to do those attacks with a weapon.

WBHarry:
Or I guess domain cards that relate to attacks don't ever actually say to make an attack. They always "trigger" off a successfull attack and such, actually 🤷

cptn-cosmo:
I mean casting 2 Fireballs in melee range does not sound like a good idea as well....

WBHarry:
Whether something is a good idea or not isn't really relevant 😛 

But I might be starting to come back around to bonuses only applying to weapon. Makes it really annoying if we want to enable a weapon to be able to give you a + to spellcasting though

cptn-cosmo:
with the stats both would make sense to me, as +1 to evasion but only for attacks made with the weapon does not make sense... I wish they included some examples...

CarlosFdez:
The fireball example was a simple one. Replace it for aoe spell attacks that use targets in range language instead and double up on that. This sort of gimmick would let you break past the threshold system, you could mark up to 6 hp since you're dealing damage two separate times with one action roll.

I see too many reasons why spells shouldn't work:
1) It says it only works for the weapon
2) Even if you redefine 1, the intent is clear since it doesn't work on other weapons. "But it didn't say spells" is a stretch
3) Weapons are more of a martial toy, but as cosmo said they're already useful for casters (except knowledge ones)
4) It promotes a degenerate "meta". If you're a caster with powerful once per rest/longrest abilities, you're encouraged to get a rapier regardless of what your stat is in order to gain double usage out of your domain card. Siphon Strike from the void for example deals d20 damage using proficiency and heals what you mark. Now my witch pretty much really wants a rapier despite not using presence in order to double up on that card.
5) The fiction doesn't make much sense, especially for non-magical weapons.

The evasion penalty stuff is weird, but it is still weird even if we take what they say to be "works on spells but not secondary weapons". I think we have to assume that features that apply to attacks are the ones that can't work on another weapon. Something like evasion affects you when you don't have the spotlight, so it can't apply to any weapon.

WBHarry:
Mm, my experience definately isn't the same with weapons for primary spellcasters. But yeah, going to have to make a difficult solution instead to specifically allow spellcast bonuses.

CarlosFdez:
Personally my witch (which was allowed to use instinct due to everyone crowding knowledge) is using hand runes + a shield quite effectively. But I don't have very good blast things spells. I got chunky armor and am kinda playing a dark knight.

If its possible to get some sort of applies to flag on the effect, which works kinda like a selector, that could work as well. It could default to all, default to primary only for primary weapons, and default to primary only for secondary weapons (out of convention). But passive stuff like evasion/armor would need to bypass that somehow, so maybe it needs to be based on what the property is? Or we'd have multiple groups of changes sharing conditionals? Would be nice to fully figure out the conditional system.

WBHarry:
Mm. Currently all effects on a Weapon are being applied to the character aslong as the weapon is equipped - so Evasion etc are correctly applied.

It's specifically modifiers that relate to the DiceRoll Dialogs that are being scrutinized when getting modifiers for the roll, and only applied according to the weapon rules.
Spellcast modifiers get caught in the crossfire there since they are `bonuses.roll.spellcast`.

WBHarry:
Worth noting of course that there are no weapons in the SRD that actually give a bonus to that. But it's one of the most obvious homebrews people would do.

CarlosFdez:
hmm yeah spellcraft should be omitted. Anything that is specific should likely not be swept up. 

We don't really even need homebrew to determine that. Secondary weapon features by the general rules don't apply to primary weapons,, but we have things like `Paired: +2 to primary weapon damage to targets within Melee range`. I think specific overrides general, and spells are specific, so it qualifies.

For avoiding a conditionals system, having primary/secondary/spellcast bonuses separate from each other as specific enough things to modify to is probably enough for the weirdness rules of weapon features to have enough of an escape hatch. Very specific "it affects this" mods could be allowed to go past the weapon feature restriction stuff. Maybe they can all be linked getters and setters that store up mutations for specific roll types using some internal modifier system? Would allow things that halve attack/damage to work. Not sure. Blighting Strike is a thing.

PR #2010, opened 2026-06-16
[Feature] Improved Countdown Animations

WBHarry:
Changed to using animate instead of class-based css

### DIFF
--- a/module/applications/ui/countdowns.mjs
+++ b/module/applications/ui/countdowns.mjs
@@ -13,7 +13,6 @@ const { HandlebarsApplicationMixin, ApplicationV2 } = foundry.applications.api;
 export default class DhCountdowns extends HandlebarsApplicationMixin(ApplicationV2) {
     previousCountdownData = null;
     changedCountdownsForAnimation = new Set();
-    countdownChangeAnimationTimeout = null;
 
     constructor(options = {}) {
         super(options);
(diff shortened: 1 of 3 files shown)


module/applications/ui/countdowns.mjs
```diff
@@ -95,35 +94,41 @@ export default class DhCountdowns extends HandlebarsApplicationMixin(Application
...
+            const shimmerTiming = {
+                duration: 1000,
+                iterations: 1
+            };
 
-                /* If the countdown element is not rendered the user doesn't have permissions to it. No animation needed on the elment itself */
-                const countdownElement = this.element.querySelector(`.countdown-container[data-countdown="${countdownKey}"]`);
-                if (!countdownElement) continue;
+            const element = this.element.querySelector(`.countdown-container[data-countdown="${countdownKey}"]`);
+            element.animate(shimmerAnimation, shimmerTiming);
```
CarlosFdez:
```suggestion
            element?.animate(shimmerAnimation, shimmerTiming);
```

module/applications/ui/countdowns.mjs
```diff
@@ -95,35 +94,41 @@ export default class DhCountdowns extends HandlebarsApplicationMixin(Application
...
+                { boxShadow: '0 0 1px 1px var(--golden)' },
+                { boxShadow: '0 0 2px 2px var(--golden)' }
+            ];
+            const pulseTiming = {
+                duration: 1000,
+                iterations: 3
+            };
+
+            const element = this.element.querySelector(`.header-type-toggles .header-type[data-type="${type}"]`);
+            element.animate(pulseAnimation, pulseTiming);
```
CarlosFdez:
```suggestion
            element?.animate(pulseAnimation, pulseTiming);
```

WBHarry:
I was about to say that I don't think the `element` to animate can ever be undefined - but it is a good thing ultimately all the same. Maybe prevents some crash if a module changes the UI elements in some weird way or whatnot :+1:

CarlosFdez:
Yeah that was my primary thought. It's ridiculously unlikely but who knows. Especially with modules.

CarlosFdez (approved):
Tested, works better!

The pulse for a disabled header could be more obvious but separate thing I think.

WBHarry:
We could achieve that by adding a parent container that doesn't get the `opacity: 0.4` and then animated on that. I think the main culprit is the opacity making things dim.

CarlosFdez:
instead of opacity, it could maybe use var(--color-text-subtle)

PR #2016, opened 2026-06-19
[Fix] Resorting features to other feature types

CarlosFdez:
Leads to more predictable behavior when drag dropping adversary and environment features. Its a bit tricky to understand, so just in case I'll break it down.

Right now features are grouped together into passives -> actions -> reactions. This takes priority over sort order, so it is possible for a passive to internally have a higher sort order than an action even if it is actually displayed first.

image

In this situation, if overgrown battlefield was actually created last (and had the highest sort order), dragging "You are not welcome here" to "Overgrown Battlefield" would actually cause "You are not welcome here" to move towards the end (after defiler). Now, it goes to the front, before "Barbed Vines". 

Resorting items is not implemented in setting sheets, so this only affects the non-edit sheets at this time. Later on though it will also support the edit sheets once those allow resorting.

### DIFF
--- a/module/applications/sheets-configs/adversary-settings.mjs
+++ b/module/applications/sheets-configs/adversary-settings.mjs
@@ -54,7 +54,7 @@ export default class DHAdversarySettings extends DHBaseActorSettings {
     async _prepareContext(options) {
         const context = await super._prepareContext(options);
 
-        const featureForms = ['passive', 'action', 'reaction'];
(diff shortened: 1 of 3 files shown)


CarlosFdez:
I would rename CONFIG.DH.ITEM.featureForm to be plural but that would cause a conflict with later work I plan to do.

WBHarry (approved):
:+1: ⭐

PR #2018, opened 2026-06-19
[Fix] old armorscore AEs clobbering data

CarlosFdez:
This causes the AE to run through field changes rather than unguided changes, making the invalid AEs usually no-op rather than corrupt data.

### DIFF
--- a/module/data/actor/character.mjs
+++ b/module/data/actor/character.mjs
@@ -315,7 +315,12 @@ export default class DhCharacter extends DhCreature {
                         label: 'DAGGERHEART.ACTORS.Character.defaultDisadvantageDice'
                     })
                 })
-            })
+            }),
+            /** Accumulated armor score from all sources */
+            armorScore: new fields.SchemaField({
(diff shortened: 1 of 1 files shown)


CarlosFdez:
This does not actually fix the old AEs though, so a better PR title / changelog entry might be ideal.

WBHarry:
Ah, haven't messed with `persisted: false` before. So the idea is to avoid breaking by defining the property on the model, but not propagating it to the DB?

CarlosFdez:
Kinda mostly. When a change effect is applied, it can either go via field changes, or via unguided changes. Field changes are a bit smarter and do some casting. Before armor was going through unguided, and this lets it go through field changes.

I had asked in the foundry channels about the possibility of explicitly erroring in the console and cancelling the change if an AE tries to change an object, but the response I got suggested I do this instead. Its unfortunately silent, but I suppose its good if the changes work a bit smarter.

WBHarry (approved):

PR #2022, opened 2026-06-19
[Fix] TagTag/GroupRoll Resource Handling

### LINKED ISSUE #2019: [Bug] Group/Tag Team roll automation
**Describe the bug**
I have the Hope/Fear automation disabled so they are manually added. Using group and tag team rolls automatically adds Hope and Fear even though the automation is disabled.

**To Reproduce**
1. Go to Daggerheart automation settings
2. Disable Hope and Fear
3. Create a group or tag team roll
4. After finishing the roll see Hope/Fear is added automatically

**Expected behavior**
If disabled, the group/tag team roll should not add Hope/Fear automatically

**Setup Information:**
 - OS: Windows
 - Browser Firefox
 - Foundry Version v14 b364
- System Version v.2.3.4

WBHarry:
Fixes #2019 
Fixed so that GroupRolls and TagTeams respect the HopeFear automation setting

### DIFF
--- a/module/applications/dialogs/groupRollDialog.mjs
+++ b/module/applications/dialogs/groupRollDialog.mjs
@@ -481,21 +481,22 @@ export default class GroupRollDialog extends HandlebarsApplicationMixin(Applicat
 
         await cls.create(msgData);
 
-        if (!shouldUseHopeFearAutomation()) return this.cancelRoll({ confirm: false });
+        /* Handle resource updates for the finished GroupRoll */
+        if (shouldUseHopeFearAutomation({ gmAsPlayer: true })) {
+            const resourceMap = new ResourceUpdateMap(actor);
+            if (totalRoll.isCritical) {
+                resourceMap.addResources([
+                    { key: 'stress', value: -1, total: 1 },
+                    { key: 'hope', value: 1, total: 1 }
+                ]);
+            } else if (totalRoll.withHope) {
+                resourceMap.addResources([{ key: 'hope', value: 1, total: 1 }]);
+            } else {
+                resourceMap.addResources([{ key: 'fear', value: 1, total: 1 }]);
+            }
 
-        const resourceMap = new ResourceUpdateMap(actor);
-        if (totalRoll.isCritical) {
-            resourceMap.addResources([
-                { key: 'stress', value: -1, total: 1 },
-                { key: 'hope', value: 1, total: 1 }
-            ]);
-        } else if (totalRoll.withHope) {
-            resourceMap.addResources([{ key: 'hope', value: 1, total: 1 }]);
-        } else {
-            resourceMap.addResources([{ key: 'fear', value: 1, total: 1 }]);
+            resourceMap.updateResources();
         }
-
-        resourceMap.updateResources();
 
         /* Fin */
         this.cancelRoll({ confirm: false });
--- a/module/applications/dialogs/groupRollDialog.mjs
+++ b/module/applications/dialogs/groupRollDialog.mjs
@@ -1,4 +1,5 @@
 import { ResourceUpdateMap } from '../../data/action/baseAction.mjs';
+import { shouldUseHopeFearAutomation } from '../../helpers/utils.mjs';
 import { emitGMUpdate, GMUpdateEvent, RefreshType, socketEvent } from '../../systemRegistration/socket.mjs';
 import Party from '../sheets/actors/party.mjs';
 
@@ -479,6 +480,8 @@ export default class GroupRollDialog extends HandlebarsApplicationMixin(Applicat
             };
 
         await cls.create(msgData);
+
+        if (!shouldUseHopeFearAutomation()) return this.cancelRoll({ confirm: false });
 
         const resourceMap = new ResourceUpdateMap(actor);
         if (totalRoll.isCritical) {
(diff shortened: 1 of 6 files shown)


CarlosFdez:
hmm, if the GM finishes a group/tag team roll, should it use the GM setting or the player setting? In your solution I think it uses the setting of whoever finalizes it. But I actually don't know for sure.

WBHarry:
Hm..I guess that's true. In this case in particular, the GM would just be helping the players along if they interacted with the dialog assumably. 

How about adding an option to the function akin to `options = { gmAsPlayer: false }` or some such, so we can still keep it localized to a single place?

CarlosFdez:
shore, works for me. Or a user: "gm" | "player" option. Either or.

module/applications/dialogs/groupRollDialog.mjs
```diff
@@ -480,6 +481,8 @@ export default class GroupRollDialog extends HandlebarsApplicationMixin(Applicat
 
         await cls.create(msgData);
 
+        if (!shouldUseHopeFearAutomation()) return this.cancelRoll({ confirm: false });
```
CarlosFdez: This might read clearer if the stuff below here up to but not including fin was in an if instead wrapping up the resource map stuff. Though for now this is fine.

module/applications/dialogs/tagTeamDialog.mjs
```diff
@@ -750,6 +750,8 @@ export default class TagTeamDialog extends HandlebarsApplicationMixin(Applicatio
 
         await cls.create(msgData);
 
+        if (!shouldUseHopeFearAutomation()) return this.cancelRoll({ confirm: false });
```
CarlosFdez: I'm looking below and I'm seeing that the resource updates below also handle the tag team cost?

WBHarry: It does. I think that makes sense. The cost of the TagTeam is `Hope` after all. It seems sensible to me that if you don't want hope automation, then you want to manually apply the cost aswell.

WBHarry:
> shore, works for me. Or a user: "gm" | "player" option. Either or.

I went with `gmAsPlayer` override. More logical for me :+1:

module/applications/dialogs/groupRollDialog.mjs
WBHarry: Haha, I was thinking of that first, but I was under the impression you generally preferred an early if break return rather than wrapping.

I went ahead and changed it. I do think it's better with an if wrap here.

CarlosFdez: its definitely case by case. I wish I knew how to put it into words.

I prefer them more when they're at the start of a function, or if its an "error" case. If there is a chance of more getting added that have nothing to do with why we're early returning, I prefer not early returning. I think its that in this case, its not an error, its just that the rest of the function is resource consumption. But more stuff could be added there, and we could forget that we're early returning in a specific case?

My best guess anyways.

module/applications/dialogs/tagTeamDialog.mjs
CarlosFdez: Discussed separately, In essence I agree, but hope and stress costs are automated elsewhere for things like experiences even when the settings are disabled. For consistency they should match.

WBHarry: Pushed up a change where the hope cost for TagTeamRoll is always taxed regardless of the setting,
But I added an Enable/Disable checkbox to the Hope Cost in the dialog to make it easy to turn off as desired.

module/applications/dialogs/tagTeamDialog.mjs
```diff
@@ -753,32 +768,40 @@ export default class TagTeamDialog extends HandlebarsApplicationMixin(Applicatio
...
-            }
-            if (finalRoll.isCritical) resourceUpdates.push({ key: 'stress', value: -1, total: 1, enabled: true });
-            if (finalRoll.withFear) {
-                fearUpdate.value = fearUpdate.value === null ? 1 : fearUpdate.value + 1;
-                fearUpdate.total = fearUpdate.total === null ? -1 : fearUpdate.total - 1;
-            }
 
-            game.actors.get(memberId).modifyResource(resourceUpdates);
+        if (shouldUseHopeFearAutomation({ gmAsPlayer: true })) {
+            for (let memberId in tagTeamData.members) {
```
CarlosFdez:
```suggestion
            for (const memberId in tagTeamData.members) {
```

module/applications/dialogs/tagTeamDialog.mjs
```diff
@@ -753,32 +768,40 @@ export default class TagTeamDialog extends HandlebarsApplicationMixin(Applicatio
...
+                const resourceUpdates = [];
+                const rollGivesHope = finalRoll.isCritical || finalRoll.withHope;
+                if (memberId === tagTeamData.initiator.memberId) {
+                    const value = tagTeamData.initiator.cost
+                        ? rollGivesHope
+                            ? 1 - tagTeamData.initiator.cost
+                            : -tagTeamData.initiator.cost
+                        : 1;
```
CarlosFdez:
```suggestion
                    const cost = tagTeamData.initiator.cost;
                    const gained = rollGivesHope ? 1 : 0;
                    const value = gained - (cost ?? 0);
```

Looking carefully, it looked like no cost always gave a hope regardless of hope result. This should handle it, but I haven't tested it.

CarlosFdez: I suppose you could also do `const value = (rollGivesHope ? 1 : 0) - (tagTeamData.initiator.cost ?? 0);` if that doesn't trigger line length limits and looks clear.

CarlosFdez: Hmm, looking deeper, you could probably replace `rollGivesHope` with `hopeGained` and you won't even have to do the ternary. 

You could consider `const cost = memberId === tagTeamData.initiator.memberId ? tagTeamData.initiator.cost ?? 0 : 0;` before the if, doing `hopeGained - cost` and then checking if the value is not zero, but I'm not 100% sure. This would skip the if/else.

CarlosFdez (approved):
From some initial poking around, works well, tank u

PR #2023, opened 2026-06-19
Support drag resort on features in actor setting sheets

CarlosFdez:
This should make the features sections in the sheet one step more usable. Its still not quite there yet, as feature forms barriors are not visible yet, and dragging across feature forms leads to weird behaviors (separately solved by https://github.com/Foundryborne/daggerheart/pull/2016).

This fixes an issue where dragging adversaries in the environment sheet caused errors, however it does not support drag/drop resorting there. You can drag/drop an adversary from there into a different category though.

### DIFF
--- a/module/applications/sheets-configs/adversary-settings.mjs
+++ b/module/applications/sheets-configs/adversary-settings.mjs
@@ -97,32 +97,4 @@ export default class DHAdversarySettings extends DHBaseActorSettings {
 
         await this.actor.update({ [`system.experiences.${target.dataset.experience}`]: _del });
     }
-
-    async _onDragStart(event) {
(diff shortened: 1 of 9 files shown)


WBHarry (approved):
:+1: Works great.

While drag resorting in the settings sheet, you really notice how we need some identification of the different action types. If you don't already know which feature is what type it might confuse the user why the feature doesn't end up where you expected (#2016 fixed so that the feature is at least moved to the correct position :+1:)

CarlosFdez:
Yeah. Basically right now I suspect most active homebrewing is done in the view sheet, not the settings sheet, for reasons such as that. There is a feature gap the edit sheets have to catch up on before we can even think of further restricting the view.

PR #2026, opened 2026-06-21
[Feature] In setting sheets, split by feature type and include descriptions

CarlosFdez:
Splits them up into these groups, and gives separate create buttons that makes a feature for that type. Added to adversary, environment, and npcs.

Also does a few extra things:
- show feature type on npc view sheet
- fix tags showing in companion limited sheet
- fixes scroll restoration for setting sheet features and potential adversaries

### DIFF
--- a/module/applications/sheets-configs/adversary-settings.mjs
+++ b/module/applications/sheets-configs/adversary-settings.mjs
@@ -10,11 +10,7 @@ export default class DHAdversarySettings extends DHBaseActorSettings {
         actions: {
             addExperience: DHAdversarySettings.#addExperience,
             removeExperience: DHAdversarySettings.#removeExperience
-        },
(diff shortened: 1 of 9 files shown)


WBHarry:
Works great.

The one thing I'm considering which I'm not sure about,
What do you think about always having the descriptions initially collapsed in the Edit Sheet, regardless of what you have the setting on? 
image
I'm not 100% on it, but I do somewhat feel that getting the big picture overview of the actiontype split has value here, instead of maybe just seeing 2 actions in the initial scroll because they have really long descriptions.

CarlosFdez:
That makes sense to me. When I wake up I'll try to see if there's a way to take care of that.

WBHarry (approved):
:+1:

PR #2027, opened 2026-06-21
[Fix] size of default token image in firefox

CarlosFdez:
Fixes a weird sizing quirk when viewing actors using the default svgs on the canvas

This is what it was like before on firefox
image


CarlosFdez:
The actual change is clearer if you look at the source diff, but basically I changed

`style="height: 512px; width: 512px;"`

to

`width="512" height="512"`

WBHarry (approved):
:+1:

PR #2029, opened 2026-06-21
[Fix] Action Uses Not Spent

### LINKED ISSUE #2025: [Bug] "Uses" are not being spent (1 per longrest etc)
Occurs on actions everywhere from what I can tell, tested actions on items and features.

`action.uses.enabled` used to exist, now it doesn't and it causes the uses to never be spent.

WBHarry:
Fixes #2025 
Fixed so that action uses are automatically spent again. 

In `v2.2.3` I added `event.preventDefault` on submit in d20RollDialog, I did this to avoid the config.title being altered on Companion rolls. 
This wasn't a good fix as it had the side effect of not updating the config with uses and other things.
I removed it again so that the config update still happens (`updateRollConfiguration`) and fixed the changing Companion Roll message title directly.

### DIFF
--- a/module/applications/dialogs/d20RollDialog.mjs
+++ b/module/applications/dialogs/d20RollDialog.mjs
@@ -226,6 +226,7 @@ export default class D20RollDialog extends HandlebarsApplicationMixin(Applicatio
 
     static async submitRoll(event) {
         event.preventDefault();
+        await this.submit();
         await this.close({ submitted: true });
     }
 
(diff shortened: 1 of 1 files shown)


CarlosFdez:
This bit looks like its made of tissue paper and bubblegum, definitely something that wants our attention eventually, but nice find.

CarlosFdez (approved):

WBHarry:
Yah, I agree. It's not great that we do the `prepareContext` from the config object, and then mutate that object in every formUpdate to make the prepareContext work again. And then making the finish work by formUpdate running again at that point.

But, this restores it to work again. Changing this takes some more thonking later.

PR #2030, opened 2026-06-21
[Fix] Trigger Error Info

### LINKED ISSUE #2014: [Bug] Wizard's Strange Patterns Duality Error Message
**This probably doesn't need a coding fix but I wanted to post here in case anyone else runs into the issues I had moving from v13 to v14.**

After updating to v14 of FoundryVTT from v13 two of my players (and the GM) started receiving error messages on every duality roll they would make from their sheet. Both characters are wizards and we were able to narrow it down to the `Strange Patterns` feature as part of their class. I needed to remove the feature from their sheets and add it back in to fix this issue. Simply updating the feature didn't work, we needed to fully remove and add it back.

**To Reproduce**
1. Have a wizard character in v13 with the `Strange Patterns` feature set up
2. Update from FoundryVTT v13 to V14 
3. Receive the following error on every duality roll
```
Duality Roll trigger failed for [WIZARD'S NAME]. It's probably configured wrong. 
```
This didn't cause any errors other than the dialog popping up.

**Setup Information:**
 - OS: mac/ipadOS, windows 11
 - Browser chrome, safari, firefox
 - Foundry Version v14

WBHarry:
Fixes #2014

- Made trigger errors more informative by including the name of the Item the trigger is on (in this case `Strange Patterns`)
image

- I changed so that the error is only printed to console, rather than showing via `ui.notifications`. This part I'm not sure on. It'll make it less annoying if you don't want to fix it, but it'll also mean most users will probably never notice a trigger is broken since they won't check the console 🤷‍♂️

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1451,7 +1451,7 @@
                 },
                 "triggerType": "Trigger Type",
                 "triggeringActorType": "Triggering Actor Type",
-                "triggerError": "{trigger} trigger on {item} failed for {actor}. It's probably configured wrong."
+                "triggerError": "{trigger} trigger on {item}({itemUuid}) failed for {actor}. It's probably configured wrong."
             },
             "WeaponFeature": {
                 "barrier": {
--- a/lang/en.json
+++ b/lang/en.json
@@ -1451,7 +1451,7 @@
                 },
                 "triggerType": "Trigger Type",
                 "triggeringActorType": "Triggering Actor Type",
-                "triggerError": "{trigger} trigger failed for {actor}. It's probably configured wrong."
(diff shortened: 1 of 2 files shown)


CarlosFdez:
Adding the uuid somewhere might also be helpful its a console message. Something in parenthesis. Might only need the one, like item?.uuid ?? currentActor?.uuid. Beyond that, this works for me.

EDIT: Its fine as is, but I didn't merge it in case you think the uuid is a good idea.

CarlosFdez (approved):

WBHarry:
I tossed on `item?.uuid`. Can definately be nice when doing some debugging at some point to copy pasta. :+1:

PR #2031, opened 2026-06-21
[Fix] Reroll Countdown Automation

WBHarry:
Fixed so that automated countdowns reverse progress from Fear when rerolling dice if the duality result changes.

IE: Rolled with fear --> Fear automated countdown progressed. Reroll the fear chat message to `with hope` --> Fear automated countdown backtracks 1

### DIFF
--- a/module/applications/ui/countdowns.mjs
+++ b/module/applications/ui/countdowns.mjs
@@ -342,9 +342,10 @@ export default class DhCountdowns extends HandlebarsApplicationMixin(Application
      * Sends updates of the countdowns to the GM player. Since this is asynchronous, be sure to
      * update all the countdowns at the same time.
      *
-     * @param  {...{ type: <key>, reverseDirection: boolean }} progressTypes Countdowns to be updated
+     * @param  {...{ type: <key>, undo: boolean }} progressTypes Countdowns to be updated
      */
     static async updateCountdowns(...progressTypes) {
+        progressTypes = progressTypes.map(p => typeof p === 'string' ? { type: p } : p);
         const { countdownAutomation } = game.settings.get(CONFIG.DH.id, CONFIG.DH.SETTINGS.gameSettings.Automation);
         if (!countdownAutomation) return;
 
@@ -353,7 +354,7 @@ export default class DhCountdowns extends HandlebarsApplicationMixin(Application
             const countdown = countdownSetting.countdowns[key];
             const progressData = progressTypes.find(x => x.type === countdown.progress.type);
             if (progressData && countdown.progress.current > 0) {
-                acc[key] = { value: progressData.reverseDirection ? 1 : -1 };
+                acc[key] = { value: progressData.undo ? 1 : -1 };
             }
 
             return acc;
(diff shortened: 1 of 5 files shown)


CarlosFdez:
This is optional, but if you like the old style of passing params, and want to keep compatibility, you could probably do something like this at the very start of the function:

`progressTypes = progressTypes.map((p) => typeof p === "string" ? { type: p } : p);`

That would allow either numerical or object based options

The type for the jsdoc would be `...(string | {type: string; reverseDirection?: boolean })`.

An alternative name for reverseDirection could be undo?

WBHarry:
`undo` is definately a better prop name. Changed it :+1:

Went ahead and restored the old string param passing with the `progressType` remapping suggested above. I do like not having to write objects with type all the time because of the one case in rerolls

module/dice/helpers.mjs
```diff
@@ -1,19 +1,33 @@
...
+                type: CONFIG.DH.GENERAL.countdownProgressionTypes.fear.id, 
+                undo: fear === 1 ? false : true 
+            });
+        }
 
-    const resourceUpdates = new ResourceUpdateMap(actor);
-    resourceUpdates.addResources(updates);
-    resourceUpdates.updateResources();
+        game.system.api.applications.ui.DhCountdowns.updateCountdowns(...updates);
+    }
```
CarlosFdez:
```suggestion
    if (countdownAutomation && fear !== 0) {
        game.system.api.applications.ui.DhCountdowns.updateCountdowns({ 
            type: CONFIG.DH.GENERAL.countdownProgressionTypes.fear.id, 
            undo: fear === 1 ? false : true 
        });
    }
```

Untestested but I don't know why the updates if there's only gonna be one thing there and only in a specific case

CarlosFdez: uh, github suggestion thing bugging out. Those are not the lines I selected.

WBHarry: Yah, it doesn't have to be in the format with `updates`. I felt like putting it like that so that it has the same format as the ResourceUpdate, and is already set up to slot in another countdown update if we end up adding a `progressOnHope` progressType or some such, which isn't altogether unlikely.

CarlosFdez: I suspect that would end up being an else branch with a single updateCountdowns call anyways, updates would still only have one item.

WBHarry: Overprep removed 🙃 🧹

CarlosFdez (approved):
Looks fine to me, though I haven't tested, so merge if you're feeling lucky.

You could probably rename updatedCountdowns to countdownDeltas and make the value a number instead of an object with a value prop, but that's me being a picky pete. You don't have to worry about that.

module/applications/ui/countdowns.mjs
```diff
@@ -342,29 +342,31 @@ export default class DhCountdowns extends HandlebarsApplicationMixin(Application
      * Sends updates of the countdowns to the GM player. Since this is asynchronous, be sure to
      * update all the countdowns at the same time.
      *
-     * @param  {...any} progressTypes Countdowns to be updated
+     * @param  {...{ type: <key>, undo: boolean }} progressTypes Countdowns to be updated
```
CarlosFdez:
```suggestion
     * @param  {...(string | { type: string; undo?: boolean })} progressTypes Countdowns to be updated
```

WBHarry:
Feeling pretty confident about this one (v2.4.1 patch here we goooo 🙃)

PR #2032, opened 2026-06-21
[Fix] Weapon Spellcasting Active Effects

WBHarry:
I made changes to `baseAction.getEffects` that is used to grab the toggleable effects shown in the d20RollDialog (such as +1 to attack rolls from `reliable`).

----
Weapons used to not return their effects at all for this unless the action being performed was from the weapon itself.
It will now return the effects, but all `changes` in the effects will be removed unless they're specifically allowed to be applied to things that isn't the weapon.
To do this, I changed so that we return a simple object version of the effect instead of the effect itself. I think this should be fine as we are essentially(maybe literally?) just using the changes array anyway.

### DIFF
--- a/module/data/action/baseAction.mjs
+++ b/module/data/action/baseAction.mjs
@@ -228,7 +228,8 @@ export default class DHBaseAction extends ActionMixin(foundry.abstract.DataModel
         let config = this.prepareConfig(event, configOptions);
         if (!config) return;
 
-        config.effects = await game.system.api.data.actions.actionsTypes.base.getEffects(this.actor, this.item);
+        config.effects = 
+            await game.system.api.data.actions.actionsTypes.base.getActionRelevantEffects(this.actor, this.item);
 
         if (Hooks.call(`${CONFIG.DH.id}.preUseAction`, this, config) === false) return;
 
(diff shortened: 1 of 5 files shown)


CarlosFdez:
There's a typo, but I think perhaps it might be better to do the inverse. Passive effects that do things like alter armor are allowed (see all the shields), its really just stuff that apply to actions that only apply to that weapon (unless its a secondary weapon that applies its effect to the weapon).

WBHarry:
The thing is that this function is specifically used to bring up the Effects (Changes) that are relevant for the d20RollDialog. IE, rolls and damage bonuses pretty much. The effects are checked for these later in the flow.

Maybe the function should be renamed

WBHarry:
So bonuses to armor, penalties to evasion etc always happen aslong as the weapon is equipped. This is specifically to determine what to show as toggle:able effects in the roll dialog.
That's why I figured it'd be better to have that tiny list that's currently just spellcasting bonus - since that's the only part of ActiveEffects on Weapons that should be made available to the roll dialog if it's an action not belonging to the weapon.

CarlosFdez:
Rename might avoid that mixup yeah.

CarlosFdez:
then beyond the typo (its transferred) I guess please include the curlies. Its kinda hard to read without.

module/data/action/baseAction.mjs
```diff
@@ -341,19 +341,31 @@ export default class DHBaseAction extends ActionMixin(foundry.abstract.DataModel
...
+            (acc, effect) => {
+                const effectData = effect.toObject();
                 /* Effects on weapons only ever apply for the weapon itself */
                 if (effect.parent.type === 'weapon') {
                     /* Unless they're secondary - then they apply only to other primary weapons */
                     if (effect.parent.system.secondary) {
-                        if (effectParent?.type !== 'weapon' || effectParent?.system.secondary) return false;
-                    } else if (effectParent?.id !== effect.parent.id) return false;
+                        if (effectParent?.type !== 'weapon' || effectParent?.system.secondary) 
+                            effectData.system.changes.filter(x => weaponTrasnferredEffectKeys.includes(x.key));
```
CarlosFdez: How does this work? filter makes a new list, and its not assigning it to anything.

WBHarry: Oh uh..huh. Yeah that doesn't work. I guess I got tricked because my Effect cases weren't very complex.

The idea is to set `effectdata.system.changes = effectData.system.changes.filter`.
Non destructive since it's now effectData from a `toObject`

WBHarry:
- Fixed the typo and added some curly fries.
- Renamed `getEffects` to `getActionRelevantEffects`
- Corrected so that the filter changes are actually set and used 🤡

module/data/action/baseAction.mjs
```diff
@@ -333,27 +334,44 @@ export default class DHBaseAction extends ActionMixin(foundry.abstract.DataModel
...
                     /* Unless they're secondary - then they apply only to other primary weapons */
                     if (effect.parent.system.secondary) {
-                        if (effectParent?.type !== 'weapon' || effectParent?.system.secondary) return false;
-                    } else if (effectParent?.id !== effect.parent.id) return false;
+                        if (effectParent?.type !== 'weapon' || effectParent?.system.secondary) {
+                            effectData.system.changes = 
+                                effectData.system.changes.filter(x => weaponTransferredEffectKeys.includes(x.key));
+                        } 
+                    }
+                    else if (effectParent?.id !== effect.parent.id) {
```
CarlosFdez:
```suggestion
                    } else if (effectParent?.id !== effect.parent.id) {
```

CarlosFdez (approved):
Code looks fine to me but I haven't tested it, so if you're feeling lucky merge it, otherwise I'll poke into it later.

PR #2033, opened 2026-06-23
[Fix] quirks involving expanding items and hovering over them

CarlosFdez:
Changes are easier to see if you disable whitespace in the diff.

This does a few things:
- Indent more properly in the summary file (refactor)
- Fixes items without a description still expanding despite not showing the "can expand" icon
- Fixes collapsed items not having a bottom border radius (by wrapping it in an item-main div)
- Adds an internal padding to the hover. Its more obvious around the item image:

image

### DIFF
--- a/module/applications/sheets/api/application-mixin.mjs
+++ b/module/applications/sheets/api/application-mixin.mjs
@@ -603,7 +603,7 @@ export default function DHApplicationMixin(Base) {
                 const doc = await fromUuid(itemUuid);
 
                 //get inventory-item description element
-                const descriptionElement = el.querySelector('.invetory-description');
+                const descriptionElement = el.querySelector('.inventory-description');
                 if (!doc || !descriptionElement) continue;
 
                 // localize the description (idk if it's still necessary)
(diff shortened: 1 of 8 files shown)


CarlosFdez:
Ran into a few issues, I'll try to clear them up at a later date.

WBHarry:
Hm, yah. At the moment it seems like items always expand their descriptions and it cannot be collapsed.

CarlosFdez:
I ended up fixing an issue where active effects and actions weren't be detected as "having a description" for the purposes of showing the little expando box.

image

It should be good now, though preferably I'd like it to sit in main for at least a few days before ending up in in a release.

WBHarry:
Looks great. Feels a lot neater with the changes.

One issue I realised the reason for now,
image

Weapons and Armor with features append text title+description via `item/base.getEnrichedDescription()`
That method is async which isn't ideal for the `hasDescription` getter you added, but since we just want to know if there is content in the description it should be possible to make a check for that without async.

----
Alternatively if that becomes tricky you could always add a closer `hasDescription` getter on `Weapon/Armor` specifically that checks both description and if there are weapon/armor features.

CarlosFdez:
Huh, forgot about that. That means that the previous hover icons weren't working for them either. I was initially concerned about adding a getter but was even more apt if a solution than I thought.

I'll try to wrap up those use cases soon.

WBHarry (approved):
:+1:

PR #2037, opened 2026-06-23
Simplify ActiveEffect sheet tags

CarlosFdez:
Removes the Passive/Temporary tag that is currently unused. Also, when the effect parent is the sheet's document, it will display the document name (Actor/Item) instead of `Adversary: X`

image

Granted effects don't show the origin in the tags. This was the previous behavior as well.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -2010,7 +2010,8 @@
             "Attachments": {
                 "attachHint": "Drop items here to attach them",
                 "transferHint": "If checked, this effect will be applied to any actor that owns this Effect's parent Item. The effect is always applied if this Item is attached to another one."
-            }
+            },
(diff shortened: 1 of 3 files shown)


WBHarry:
Hm, I'm not sure the `Actor`|`Item` really adds any information. I imagine it'll more likely confuse users as to what that is supposed to mean.
image

CarlosFdez:
I am down for an alternative label, but I can't think of what fits "this document". Here it is without the tag at all. This better?

image

CarlosFdez:
Went tagless for top level on actor specifically, but added origin for effects that come from other actors.

WBHarry (approved):
:+1:

PR #2039, opened 2026-06-23
[Fix] Cleanup RangeDependant

WBHarry:
Removed the RangeDependent setting for ActiveEffects.
I don't think it's needed anymore since `Paried` and `DoubleDuty` can now toggle the damage bonus in the roll dialog:
image

The Activation/Inactivation when moving in and out of range appears quite buggy to users. I think it's nicer without 🪄 

----
RangeDependant was a required attribute with an enabled prop - so it was present in every single ActiveEffect. The PR touches many files since I removed it from every single one in the SRD.


CarlosFdez:
I actually don't know about outright removal. I agree with removing the auto activation/inactivation on move, but isn't there a way to auto-enable the paired AE based on range and targets when constructing the damage formula?

WBHarry:
We could look at doing that to make the Toggle in the RollDialog smarter yeah.

WBHarry:
Don't think that'd be hard at all actually 🤔 
Can give that a try.

WBHarry:
I've restored the data model, html in the settings etc for RangeDependent on my local branch.
Changed so that it's nullable now so that it's not defined on every single baseEffect anymore, which is nice.

Now regarding the new functionality:
Previously it triggered on move, which meant it could be used for anything. You could gain a +1 to evasion when someone was within melee range.
That won't be the case now, then. It would strictly be used for modifiers in the RollDialog I guess :thinking:

CarlosFdez:
I suppose that's fine. We could think of those cases separately when the time comes. It could be something like effects on the target clone if its only against attackers in melee range. Or not at all I guess.

We can think of the "+1 evasion all the time when someone is melee" case when it actually matters maybe.

WBHarry:
The tricky thing then becomes to communicate what it does within the ActiveEffect sheet. Currently we have:
image

To avoid users trying to do things it currently won't support I guess we change the title?
`Range Dependence (Rolls Only)` or some such

CarlosFdez:
That works for me. It doesn't really stop it from expanding either to more scenarios as needed in the future, but it avoids confusion for now.

module/config/generalConfig.mjs
```diff
@@ -130,7 +130,7 @@ export const otherTargetTypes = {
         id: 'any',
         label: 'DAGGERHEART.CONFIG.TargetTypes.any'
     }
-};
+}
```
CarlosFdez: ?

WBHarry: It's a really important change
image

Nah, just my bringing stuff back with copy-pasta after the initial commit, ended up changed there 👅

WBHarry:
I'm not sure we have all the pieces in place needed to easily replace the current `OnMove` logic.

With that in mind, I went ahead and restored it, making this PR mostly a cleanup:
- Removed all uses of RangeDependence currently shipped in the system as those cases are unncessary now.
- Made RangeDependence a nullable field instead of having an enabled property.
- Changed the ActiveEffect sheet display for RangeDependence to an optional fieldset in the same was as Stacking and Armor.
- Removed all the dummy RangeDepedence data that was on every single ActiveEffect in the SRD because the field wasn't nullable.
- Added an `.active` styling to optional fieldsets so that they have a `row-gap` when not set to null, giving some space between elements.

I couldn't come up with a good way to fix existing effects at the moment, so I left the `enabled` property on the RangeDependence DataModel. If I just remove it, every single effect on existing characters will trigger since every effect had the `RangeDependence` field set (but with enabled false).

src/packs/adversaries/adversary_Archer_Guard_JRhrrEg5UroURiAD.json
```diff
@@ -325,21 +325,15 @@
           "_id": "wGuxOLokMqdxVSOo",
           "type": "base",
           "system": {
-            "rangeDependence": {
-              "enabled": false,
-              "type": "withinRange",
-              "target": "hostile",
-              "range": "melee"
-            }
+            "changes": [
```
CarlosFdez, line 328: It looks like the changes were outside of system data before. Is this a change in v14? Was foundry migrating these?

src/packs/domains/domainCard_Body_Basher_aQz8jKkCd8M9aKMA.json
```diff
@@ -26,27 +26,21 @@
       "_id": "8uOMyEczlY2OrKau",
       "img": "icons/skills/melee/hand-grip-staff-teal.webp",
       "system": {
-        "rangeDependence": {
-          "enabled": true,
```
CarlosFdez, line 30: looks like this one got removed despite being enabled.

daggerheart.mjs
```diff
@@ -390,7 +390,7 @@ const updateActorsRangeDependentEffects = async token => {
     ).rangeMeasurement;
 
     for (let effect of token.actor?.allApplicableEffects() ?? []) {
-        if (!effect.system.rangeDependence?.enabled) continue;
+        if (!effect.system.rangeDependence || effect.system.rangeDependence.enabled === false) continue;
```
CarlosFdez: I imagine the previous logic would be inclusive of this, unless enabled has some value other than true that would allow this. But perhaps I missed something in the conversion.

CarlosFdez:
> I'm not sure we have all the pieces in place needed to easily replace the current `OnMove` logic.
> 
> With that in mind, I went ahead and restored it, making this PR mostly a cleanup:
> 
>     * Removed all uses of RangeDependence currently shipped in the system as those cases are unncessary now.
> 
>     * Made RangeDependence a nullable field instead of having an enabled property.
> 
>     * Changed the ActiveEffect sheet display for RangeDependence to an optional fieldset in the same was as Stacking and Armor.
> 
>     * Removed all the dummy RangeDepedence data that was on every single ActiveEffect in the SRD because the field wasn't nullable.
> 
>     * Added an `.active` styling to optional fieldsets so that they have a `row-gap` when not set to null, giving some space between elements.
> 
> 
> I couldn't come up with a good way to fix existing effects at the moment, so I left the `enabled` property on the RangeDependence DataModel. If I just remove it, every single effect on existing characters will trigger since every effect had the `RangeDependence` field set (but with enabled false).

I might not be understanding actually. I thought you were keeping it around but making it so it was nullable so that old values will still exist but it no longer wastes so much space. But then you said existing cases are unnecessary now, but I don't see any replacement for those cases in this PR.

src/packs/adversaries/adversary_Archer_Guard_JRhrrEg5UroURiAD.json
WBHarry: Yeah, previously this was outside of system data. There is a migration, so it was never an issue with the old data existing, but I went ahead and corrected it.

src/packs/domains/domainCard_Body_Basher_aQz8jKkCd8M9aKMA.json
WBHarry: Intended. Even the cases where it was previously enabled are just annoying. 
I think it's better to just let the player enable/disable in the roll dialog rather than having the effect turn on/off constantly as they move.

CarlosFdez: I thought the goal was to replace with range detection when rolling instead of when moving, but if you remove the data from all cases, that cannot ever be implemented. So I'm not sure why the data still exists at all.

You said this:
> Removed all uses of RangeDependence currently shipped in the system as those cases are unncessary now.

But it looks like instead of unnecessary, you mean abandoned/removed? Unnecessary implies an alternative approach, but I don't see one in the PR.

WBHarry:
Basically, I've come around to that the current toggling on/off with activeeffects is actively bad. I don't want being used on any SRD compendium items.

If anyone wants to use the functionality on items they make, they're free to, so the functionality stays in.

Maybe later we can make it do something in a better way. I couldn't easily do it the way we discussed because it demands a lot of information that isn't there atm.

CarlosFdez:
I suppose sure, though the data loss will mean that doing it in a better way won't really be possible. I think people will start asking us why its not in use for existing items either (edit: I think they'll start reporting it as a bug). I'm leaning on outright removal if we're removing it from all compendium items and we're not gonna try checking during the attack instead.

CarlosFdez (approved):
Very good cleaner data and forms!

PR #2041, opened 2026-06-23
[Fix] Combat Modifier Toggle Fix

### LINKED ISSUE #2040: [Bug] BP Damage Modifier fails to be applied
**Describe the bug**
When toggling the  "-2 BP: if you add +1d4 (or a static +2) to all adversaries' damage rolls (to increase the challenge without lengthening the battle)" BP Modifier, it fails and only adds the "Increase Damage" effect to one adversary. The following error message is produced:

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'includes')
    at daggerheart.js:35969:61
    at Array.filter (<anonymous>)
    at DhpCombat.toggleModifierEffects (daggerheart.js:35969:29)
    at async DhTooltipManager.toggleModifier (daggerheart.js:37814:9)
```

The same issue happens when trying to add a token to a battle with this modifier activated.

**To Reproduce**
1. Add 3 different adversary tokens to a scene
2. Create Combat
3. Add adversaries to combat
4. Hover over the BP calculation and click on the "-2 BP: if you add +1d4 (or a static +2) to all adversaries' damage rolls (to increase the challenge without lengthening the battle)" toggle to activate it
5. See the Increase Damage effect is only added to one of the 3 adversaries.
6. See error message on console

Alternatively:

1. Add 3 different adversary tokens to a scene
2. Create Combat
3. Hover over the BP calculation and click on the "-2 BP: if you add +1d4 (or a static +2) to all adversaries' damage rolls (to increase the challenge without lengthening the battle)" toggle to activate it
4. Select all adversaries.
5. Add them to combat.
6. See the Increase Damage effect is only added to one of the 3 adversaries.
7. See error message on console

**Expected behavior**
"Increase Damage" effect should be added to all adversaries and no error should be produced.

**Screenshots**

Option to be toggled:
Image

**Setup Information:**
 - OS: Windows
 - Browser: Windows Portable Foundry Instance and Chrome
 - Foundry Version: v14 b364
- System Version: v2.4.0

**Additional details**
The error seems to be in [this](https://github.com/Foundryborne/daggerheart/blob/958eaa310c2b33414527b68afbe6b4ecc64f02c3/module/documents/combat.mjs#L49) line. From what I've seen, the `effectTargetTypes` attribute is getting lost somewhere in that reduce from line 23.

WBHarry:
Fixes #2040
Appearantly foundy's `createEmbeddedDocuments` method is mutating.

Toggling the combat modifier in the BP menu was using the same effectData to create an effect on each adversary, and it failed if there were many adversaries in the loop because our `effectTargetTypes` property was stripped out during `createEmbeddedDocuments` because it's not part of the data saved to the DB.

### DIFF
--- a/module/documents/combat.mjs
+++ b/module/documents/combat.mjs
@@ -46,7 +46,9 @@ export default class DhpCombat extends Combat {
             for (let actor of actors) {
                 await actor.createEmbeddedDocuments(
                     'ActiveEffect',
-                    effects.filter(x => x.effectTargetTypes.includes(actor.type))
+                    effects
+                        .filter(x => x.effectTargetTypes.includes(actor.type))
+                        .map(x => foundry.utils.deepClone(x))
                 );
             }
         } else {


CarlosFdez (approved):
I am wondering if the effect is a document and should probably be using toObject() somewhere but otherwise it seems fine.

WBHarry:
We could do that higher up where it's currently just being spread

PR #2044, opened 2026-06-27
[Fix] V13 Migration Fixes

WBHarry:
- Fixed so that damageParts without an applyTo field is assumed to be hitPoints in migration.
This occurs on Character/Adversary basic attacks because the `applyTo` field wasn't set. Just in case there would be some freak case where there were -multiple- parts with the `undefined` key I made any instances beyond the first be tossed out during migrateData.
- Made a tentative fix to the migration script where one user noted that it error:ed for them with `entity` = null. I haven't been able to replicate it. This assumably fixes it, but it's a shot in the dark unless we get some more info and can actually test it.
- Added some extra safeties to the old V13 migrations. An issue happened because someone was updating from v1.2.0 from back in november all the way to latest V14. A few settings that were expected back in those V13 migrations no longer exist in V14 which caused errors.

### DIFF
--- a/module/systemRegistration/migrations.mjs
+++ b/module/systemRegistration/migrations.mjs
@@ -1,5 +1,4 @@
 import { defaultRestOptions } from '../config/generalConfig.mjs';
-import { RefreshType, socketEvent } from './socket.mjs';
 
 export async function runMigrations() {
     let lastMigrationVersion = game.settings.get(CONFIG.DH.id, CONFIG.DH.SETTINGS.gameSettings.LastMigrationVersion);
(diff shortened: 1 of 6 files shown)


module/systemRegistration/migrations.mjs
```diff
@@ -155,61 +155,65 @@ export async function runMigrations() {
...
-                    progress: {
-                        ...countdown.progress,
-                        type: countdown.progress.type.value
-                    }
-                };
-
-                return acc;
-            }, {});
-        };
+        if (!countdownSettings.narrative || !countdownSettings.encounter) {
```
CarlosFdez: Wouldn't this fully skip migrations if one of these two don't exist? Would it be && instead of ||? Or is it that if one doesn't exist, its too forgone?

WBHarry: Both were removed in v2.0, so it's not really something to consider that one would exist and the other wouldn't.

Ultimately, we -could- also solve this by returning the narrative and encounter props to the countdown data - that's the only way it'll work if some crazy person downloaded v1.0 and then updated to v14.
But I don't really care much about servicing that sort of usecase atm. I was really suprised someone was updating from 1.2 today 🙃

CarlosFdez: I guess my thought is that if a person made one narrative countdown, then perhaps encounter countdowns might not have data. But not sure.

WBHarry: Changed it to be a `migrateData` instead. Makes it stable and grannular

CarlosFdez: That is better yeah, these types of migrations are last resort.

module/systemRegistration/migrations.mjs
```diff
@@ -153,63 +152,28 @@ export async function runMigrations() {
...
+                return acc;
+            }, {});
 
-        lastMigrationVersion = '1.2.7';
+            await tagTeam.updateSource({
+                initiator: initatorMissing ? null : tagTeam.initiator,
+                members: missingMembers
+            });
+            await game.settings.set(CONFIG.DH.id, 'TagTeamRoll', tagTeam);
+        } finally {
```
CarlosFdez: Wouldn't a try finally still raise the exception, preventing it from reaching the very end? If so, then the last migration version thing here wouldn't really be doing anything.

CarlosFdez:
Other than the try finally not really doing anything, seems fine to me.

module/systemRegistration/migrations.mjs
WBHarry: Hm, is that so?

What I want to achieve is that if the error occurs, then don't run the code in there and just update the migration number (what's in the `finally`).
I was tired of the lint:er complaining that I can't have `catch(_error)` because the variable is unused - and at the same time it doesn't allow me to just have an empty function either 😆

WBHarry: Pretty sure `finally` is run regardless of if there is an error or not 🤔

CarlosFdez: Finally is run, but not the rest of the function, so there's nowhere to use the variable you set.

```
try {
} catch {
}
```

Should work with the linter?

WBHarry: Huh, waddya know. I had internalised that a catch statement needed to be set up as a function. 
Changed it to that :+1:

PR #2045, opened 2026-06-28
[Fix] SummonAction Actor Choice

WBHarry:
Fixed logic for picking which actor to summon. The previous logic didn't work correctly if the actors in the compendium had artwork - which they might have with certain modules that map art onto them.

The new logic is as such:
**1)** The actor to be summoned is stored in a compendium
    **a)** There exists a an actor in the world with the same name as the one to be summoned
        a1) Use the in-world actor if: They both have the same artwork, or the compendium actor has the default artwork.
        a2) Goto `b`
    **b)** No in-world actor, so we create a new actor.
   
----
**2)** Not a compendium actor, so just use the actor straight up.

### DIFF
--- a/module/data/fields/action/summonField.mjs
+++ b/module/data/fields/action/summonField.mjs
@@ -65,9 +65,16 @@ export default class DHSummonField extends fields.ArrayField {
     /* Check for any available instances of the actor present in the world if we're missing artwork in the compendium. If none exists, create one. */
     static async getWorldActor(baseActor) {
         const dataType = game.system.api.data.actors[`Dh${baseActor.type.capitalize()}`];
-        if (baseActor.inCompendium && dataType && baseActor.img === dataType.DEFAULT_ICON) {
+        if (baseActor.inCompendium) {
             const worldActorCopy = game.actors.find(x => x.name === baseActor.name);
-            if (worldActorCopy) return worldActorCopy;
+            if (
+                worldActorCopy && (
+                    baseActor.img === worldActorCopy.img || 
+                    (dataType && baseActor.img === dataType.DEFAULT_ICON)
+                )
+            ) {
+                return worldActorCopy;
+            } 
 
             return await game.system.api.documents.DhpActor.create(baseActor.toObject());
         }
--- a/module/data/fields/action/summonField.mjs
+++ b/module/data/fields/action/summonField.mjs
@@ -44,7 +44,7 @@ export default class DHSummonField extends fields.ArrayField {
 
             const actor = await getWorldActor(await foundry.utils.fromUuid(summon.actorUUID));
             /* Extending summon data in memory so it's available in actionField.toChat. Think it's harmless, but ugly. Could maybe find a better way. */
-            summon.actor = actor;
+            summon.actor = actor.toObject();
 
             const countNumber = Number.parseInt(count);
             for (let i = 0; i < countNumber; i++) {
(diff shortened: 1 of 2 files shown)


CarlosFdez:
Thoughts on using a compendiumSource based solution? Could be that the compendium actor had art from a module, but someone changed it locally. Could be some sort of priority where art match = higher priority.

WBHarry:
What do you mean by art match in this case?

Atm my thinking is that there are two reasons to search for an in-world actor atm. The reason we do it at all is because summon actions and transformations in the SRD are obviously set up against the UUID of other actors in the compendium, which by default have no images.
1) The compendium actors have no images. A lot/most users will just drag in the actors they want into their world and set up the images they want on them there. Likely, they will not be editing the summon/transform actions to change the UUID.
In these cases it makes sense to check if there's a name matched actor in the world that has different artwork than what's in the compendium, and in that case use that one instead.
2) Regardless of the image status, it's good to not just create a new actor copy in-world every single time a summon/transform action is used. Good to name check for an in-world copy and use that one instead.

With compendiumSource, you're just thinking of checking `game.actors.find(x => x.uuid === summonActor.source.compendiumSource)` kinda thing instead of checking on name, right?

WBHarry:
Hm..I guess one potential pitfall matching on `compendiumSource` would be that someone can have grabbed an Adversary from the compendium and made two copies, changing one of them to be named Steve, while the original is still named Bandit Ambusher or whatnot.
The compendiumSource check wouldn't care and just grab either one.

WBHarry:
Maybe a check that both the compendiumSource -and- the name matches is the right way. Better safe than sorry with whatever weird practises people get up to.

WBHarry:
Maybe this as a general function for how we grab in-world actors works. Don't think we actually need to check what the artwork on them is.
```js
export async function getWorldActor(baseActor) {
    if (baseActor.inCompendium) {
        const worldActorCopy = game.actors.find(x => x._stats.compendiumSource === baseActor.uuid && x.name === baseActor.name);
        if (worldActorCopy)
            return worldActorCopy;

        const baseActorData = baseActor.toObject();
        return await game.system.api.documents.DhpActor.create({ ...baseActorData, _stats: { ...baseActorData._stats, compendiumSource: baseActor.uuid } });
    }

    return baseActor.toObject();
}
```

Dunno if there's another way to make sure the _stats.compendiumSource gets populated when creating a new actor, but this works at least 🤷

CarlosFdez:
> What do you mean by art match in this case?

I mean images matching. But perhaps that is not the right thing to compare. For example, in your thought experiment about the Bandit Ambusher and Steve, my thought is that perhaps it can attempt to pull the Bandit Ambusher first. If there is no Bandit Ambusher, it pulls Steve. Unless Steve is a linked actor, then maybe it shouldn't pull Steve.

Your thought on using a method might be smart. For a world actor heirarchy, I'd maybe do something like this (fixing any errors of course):
```js
const allWorldActors = game.actors.filter(a => a._stats.compendiumSource === baseActor.uuid && !SOMELINKEDCHECK)
const worldActor = allWorldActors.find(a => a.name === baseActor.name) ?? allWorldActors[0];
if (worldActor) return worldActor;
```

WBHarry:
Hm..going back and forth on things, I don't think our proverbial Steve should ever be used in these situations.

If someone explicitly renames an Adversary, I don't think it's reasonable to think that they'd want that one to be used for summons/transforms of the stock Adversary.

WBHarry:
Although eh...nah, I dunno. Either way we do it some users would think it's wrong.

If people just decide that their `Bandit Ambushers` should be called `Chromium Thugs` instead, then they would probably expect them to still be picked up for the actor fetch.
So you're probably right that the best we can do is to check for LinkedActor and exclude those.

PR #2046, opened 2026-06-29
[Feature] Transform Linked Actors

WBHarry:
Added ability to transform linked actors. Doing this occurs a very noticeable laggspike however, so we might have to reconsider how we do the transformation.

### DIFF
--- a/daggerheart.d.ts
+++ b/daggerheart.d.ts
@@ -4,6 +4,15 @@ import '@common/primitives/global.mjs';
 import Canvas from '@client/canvas/board.mjs';
 import { ResourceUpdateMap } from './module/data/action/baseAction.mjs';
 
+import * as applications from './module/applications/_module.mjs';
+import * as data from './module/data/_module.mjs';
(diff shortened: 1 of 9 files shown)


WBHarry:
I went ahead and changed the implementation so that we now delete the previous token and make a new one. This enables working with tokens that are linked without issue.
The benefit the update previously had of the sheet change neatly changing via animation seems to have disappeared anyway at some point, so I don't think there's any real negatives.

----
I kept it so that if you transform a token of ActorA into a token of ActorB where ActorB is linked, then the token does become linked. I think this is corrected as you're essentially transforming some random mook interpretation into the hidden BBEG or the like. 🥷

CarlosFdez (approved):
Other than the effects thing, which could maybe be filed as not a new bug maybe, it looks fine to me. However I made some changes myself, so you may have to look over those. Merge when you're comfortable with it.

CarlosFdez:
This is also a bugfix for transforms not animating, so it can also get that mention in patch notes I think.

PR #2048, opened 2026-06-30
[UI] Remove fieldset from top level notes

CarlosFdez:
Removes the fieldsets from all top level notes. Character biography is unaffected.

Another extraction from the great fieldset purge, this time just for notes.


WBHarry (approved):
:+1:

PR #2049, opened 2026-06-30
[Feature] ActiveEffect Auras

WBHarry:
- Added Aura type to action areas. Auras uses the core feature of `attachToToken` to make it follow along with a token it's placed on.
- Added `hasHole` property to areas. This toggles on the core `hole` feature of regions, creating a hole in the region where it's placed down. If this is done with an aura, the hole is the size of the token it's attached to.

### DIFF
--- a/src/packs/subclasses/feature_Elemental_Aura_2JH9NaOh69yN80Gw.json
+++ b/src/packs/subclasses/feature_Elemental_Aura_2JH9NaOh69yN80Gw.json
@@ -168,12 +168,7 @@
           "type": "self",
           "amount": null
         },
-        "effects": [
-          {
-            "_id": "mJBA2QTyM9SM5NVS",
-            "onSave": false
-          }
-        ],
+        "effects": [],
         "roll": {
           "type": "diceSet",
           "trait": null,
@@ -296,7 +291,7 @@
       "name": "Elemental Aura: Earth",
       "disabled": false,
       "img": "icons/magic/control/buff-strength-muscle-damage-red.webp",
-      "description": "<p><span style=\"color:rgb(239, 230, 216);font-family:Montserrat, sans-serif;font-size:14px;font-style:normal;font-variant-ligatures:normal;font-variant-caps:normal;font-weight:400;letter-spacing:normal;orphans:2;text-align:start;text-indent:0px;text-transform:none;widows:2;word-spacing:0px;-webkit-text-stroke-width:0px;white-space:normal;background-color:rgba(24, 22, 46, 0.753);text-decoration-thickness:initial;text-decoration-style:initial;text-decoration-color:initial;display:inline !important;float:none\">You gain a +1 bonus to Strength.</span></p>",
+      "description": "<p>Your allies gain a +1 bonus to Strength.</p>",
       "transfer": false,
       "statuses": [],
       "system": {
--- a/src/packs/subclasses/feature_Elemental_Aura_2JH9NaOh69yN80Gw.json
+++ b/src/packs/subclasses/feature_Elemental_Aura_2JH9NaOh69yN80Gw.json
@@ -52,7 +52,7 @@
           "includeBase": false
         },
         "target": {
-          "type": "hostile",
+          "type": "",
           "amount": 1
         },
         "effects": [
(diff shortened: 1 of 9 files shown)


CarlosFdez:
From testing it seems to work well, and it still handles placing emanations that stay behind. I did run into one issue though, there is a blank option next to placed template and aura, and selecting that blank option breaks the sheet.

However, I'm not sure about calling it "Aura". Every use of the word aura I can find in the core book (Confusing Aura/Shield Aura/Overwhelming Aura) are single target effects, not a floating range that follows the user. Furthermore, what is a cone shaped aura?

What effects is this trying to model? That might help us come up with a more fitting name.

WBHarry:
It's not a Daggerheart term at all. I don't think there are any uses of the classical "aura" ability in Daggerheart.
It's just something that Foundry does support, and that feels like a lot of homebrew would like to use. That is to say, I have an RPG style aura that travels with me and affects people.

WBHarry:
image

WBHarry:
As far as different aura shapes are concerned, I dunno what they'd be used for. I just didn't see an overly compelling reason to limit which shapes it could take.

CarlosFdez:
I think the feature is fine (other than the bug I mentioned), I just don't think the name fits. Both because aura is used for a different meaning in the book where that's not what aura means and because calling this an aura seems strange.

https://github.com/user-attachments/assets/0c4255c6-54e7-4c6b-8f72-35887f42c74f

My immediate thoughts are "Attached" and "Follows Target". I would like more opinions on this.

WBHarry:
Mm, yeah as far Foundry is concerned it's `Attached To Token`

WBHarry:
Removed the blank options that were causing issues (This was actually the case in all 3 selects)

I think `Attached` could be a good name unless we come up with another.

src/packs/subclasses/feature_Elemental_Aura_2JH9NaOh69yN80Gw.json
```diff
@@ -52,7 +52,7 @@
           "includeBase": false
         },
         "target": {
-          "type": "hostile",
+          "type": "",
```
CarlosFdez, line 55: This will have to be fixed at some point, but it seems like this isn't a new issue.

src/packs/subclasses/feature_Elemental_Aura_2JH9NaOh69yN80Gw.json
```diff
@@ -258,79 +287,120 @@
...
+      "_key": "!items.effects!2JH9NaOh69yN80Gw.WX5AMEpmUAutB9Hm"
     },
     {
-      "name": "Elemental Aura (Fire)",
-      "img": "icons/magic/fire/barrier-wall-explosion-orange.webp",
-      "origin": "Compendium.daggerheart.subclasses.Item.2JH9NaOh69yN80Gw",
+      "name": "Elemental Aura: Earth",
+      "disabled": false,
+      "img": "icons/magic/control/buff-strength-muscle-damage-red.webp",
+      "description": "<p><span style=\"color:rgb(239, 230, 216);font-family:Montserrat, sans-serif;font-size:14px;font-style:normal;font-variant-ligatures:normal;font-variant-caps:normal;font-weight:400;letter-spacing:normal;orphans:2;text-align:start;text-indent:0px;text-transform:none;widows:2;word-spacing:0px;-webkit-text-stroke-width:0px;white-space:normal;background-color:rgba(24, 22, 46, 0.753);text-decoration-thickness:initial;text-decoration-style:initial;text-decoration-color:initial;display:inline !important;float:none\">You gain a +1 bonus to Strength.</span></p>",
```
CarlosFdez: This HTML looks strange. Did you copy paste without holding shift? You need to use shift to get rid of styling copied from the source material.

CarlosFdez:
Outside of the HTML, looks good to me.

src/packs/subclasses/feature_Elemental_Aura_2JH9NaOh69yN80Gw.json
WBHarry: Ah, yeah that one did not spark joy. Copy pasta killed the radio star.

Fixed :+1:

PR #2050, opened 2026-07-01
[Refactor] Sidebar resources styling

CarlosFdez:
Reduces a lot of extraneous styling and reduces reliance on position absolute for spacing, allowing things to be a bit more responsive and easier to resize and tweak. For example, if I were to hypothetically reduce the width of the adversary sidebar. I also made the pips styling slightly more general than before and removed an unnecessary wrapping element.

There are likely more improvements that could be done, including renaming some of the class names (slot-bar isn't really informative, and slot-value for slot + value is strange) but perhaps that might be better saved for an actual pips rework.

Most things should look the same, however there are a few visual tweaks. Most of them involve borders, but for example this changed:

Before
image
After
image

Files that changed completely are mostly because of carriage returns getting removed automatically by gitattributes (which were not in place when the project was started). If you disable whitespace changes when viewing the diff, they should be easier to read.

This is highly likely to break some UI modules, so it might be worth making a mention in patch notes?

### DIFF
--- a/module/applications/sheets/actors/adversary.mjs
+++ b/module/applications/sheets/actors/adversary.mjs
@@ -58,12 +58,13 @@ export default class AdversarySheet extends DHBaseActorSheet {
             template: 'systems/daggerheart/templates/sheets/actors/adversary/features.hbs',
             scrollable: ['.feature-section']
         },
-        notes: {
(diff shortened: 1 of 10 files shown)


WBHarry (approved):
:+1:

PR #2052, opened 2026-07-01
[Fix] Tweak retrieving world actor

CarlosFdez:
Now only retrieves those where "linked" matches, and prioritizes name matches over non name matches.


WBHarry (approved):
:+1:

PR #2054, opened 2026-07-01
[Fix] ChatMessage Alias Use

### LINKED ISSUE #2043: [Bug] Chat Messages show Actor name instead of Token name
In these situations:
- A token is selected and the `Public as character` chat mode is selected
- A roll is made/feature is used from a token

When the chat message is made the title will state the Actor's name. This isn't ideal as the token might have set a different name for the purposes of hiding the real name, or just to make many copies with different names for RP purposes.

WBHarry:
Fixes #2043 
Changed so that chat messages never show the actor name, instead using alias

### DIFF
--- a/templates/ui/chat/chat-message.hbs
+++ b/templates/ui/chat/chat-message.hbs
@@ -6,18 +6,13 @@
             <div class="message-sub-header-container">
                 {{#if message.title}}
                     <h4>{{message.title}}</h4>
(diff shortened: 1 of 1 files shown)


CarlosFdez:
Lets give it a shot. If its finicky we can always refine it.

CarlosFdez (approved):

PR #2055, opened 2026-07-01
[UI/UX] Reduce width of adversary sidebar

CarlosFdez:
Reduces width of the initial sheet size and the adversary sidebar by 15px.

image


WBHarry (approved):
:+1:

PR #2056, opened 2026-07-01
[Fix] item expand state not being retained across rerenders

CarlosFdez:
This occured last release when I wraped the header and summary in an item main for styling reasons.


WBHarry (approved):
:+1:

PR #2057, opened 2026-07-02
More preliminary sheet work and refactors

CarlosFdez:
Moves things more to where they are used, and does some style fixes on inventory item buttons (noticeable on the right border of actions that don't have uses counts).

item-button is a multi-button and can be a standard element, but the grid layout for multiple side by side actions is specifically an inventory thing.

There is a minor visual change to glassy fieldsets due to the removal of the invisible border. If it feels weird let me know. The inline padding only makes the normal value explicit, by default on chrome fieldsets have a 10.5px inline padding (which scales on higher font sizes).

### DIFF
--- a/styles/less/global/elements.less
+++ b/styles/less/global/elements.less
@@ -264,10 +264,11 @@
         margin-top: 5px;
         border-radius: 6px;
         border-color: @color-fieldset-border;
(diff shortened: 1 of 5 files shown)


WBHarry (approved):
Nothing really looks off to me :+1:

PR #2059, opened 2026-07-03
[Fix] adjusting tiers of higher level creatures and hordes

CarlosFdez:
Everything was presumed as tier 1. calculateAdjustedDamage() returned just value in a specific branch, causing errors in early returns (such as reducing attack damage that is already low, for example tier 2 hordes reduced to tier 1) so rather than recalculate the formula at every early return, I instead moved formula generation out of that function.


WBHarry (approved):
:+1:

PR #2061, opened 2026-07-03
[Feature] ComboDice Modifier

WBHarry:
Solved ComboDice functionality by adding a new Modifer. I made a `BaseDie` class that overrides Foundry's basic dice so that it can be used on any roll.

I set it up so that the basic `2d4c` will run the standard logic of continuing to roll d4s aslong as the latest roll was higher than the one before it
image

----
I also added so that `2d4c1` will run the upgraded behavior from `Compounding`
image
Additional info on the modifier has to be a number, so I just went with 1. Maybe something smarter could be chosen. 🤷 

image
It's possible the above could be improved. All the shown dice are the d4 dice, eventhough the third dice in this example was a d6 that rolled a 5. This happens because the `Die` class has a single denomination used for all the dice inside of the expression. Will have to check on this more.


CarlosFdez:
I guess two things:

1) Is there a use for 2d4c? Could the functionality of 2d4c1 become 2d4c?
2) Let me know when you fix the die icon issue. That seems like it might be pretty important.

WBHarry:
`2d4c` is the base rule for combo dice. `2d4c1` is the modified rule you get if your character uses the Martial Artist stance `Compounding`, so both need to exist.

WBHarry:
> I guess two things:
> 
> 1. Is there a use for 2d4c? Could the functionality of 2d4c1 become 2d4c?
> 2. Let me know when you fix the die icon issue. That seems like it might be pretty important.

I managed to solve 2. The dices are excluded from the normal DiceSoNice flow and instead rolled in place in the comboDice handling.

WBHarry:
I changed so that we now have two registered modifiers:
`c` - ComboDice. The normal logic. `2d4c`
`cc` - Compound ComboDice. The adjusted compound logic. Ex: `2d4cc`

Both go into the same handler and just pass a different bool. I think this is a lot more logical than the previous `2d4c1`

WBHarry:
> I guess two things:
> 
> 1. Is there a use for 2d4c? Could the functionality of 2d4c1 become 2d4c?

Ah, yeah now I get what you meant by 1. 
Yeah we could either omit it entirely and just have people use foundry's exploding dice functionality here. 
Or we could keep the `2d4c` as a shortcut that rewrites itself to the exploding dice formula, potentially.

WBHarry:
Although the commonly suggested roll formula doesn't work correctly:
`/r 1d4x1x2x3x4`
image

The above shouldn't have gone beyond 2 dice since the second dice wasn't higher than the first 🤷 
It also makes the dice get rolled one by one and waiting on each one. Currently I didn't set up our implementation to do that as I found it annoying.

Psitacus:
The formula gave the right result though

IMG_2132

“…if the value of the second roll is EQUAL TO OR GREATER THAN your first roll…”

WBHarry:
Ah, equal or greater, right-ho.

WBHarry:
I really dislike Brawler. Silly one-note severe damage machine 😆

module/dice/die/baseDie.mjs
```diff
@@ -8,5 +17,199 @@ export default class BaseDie extends foundry.dice.terms.Die {
...
+
+    /** @override */
+    getResultCSS(result) {
+        const hasSuccess = result.success !== undefined;
+        const hasFailure = result.failure !== undefined;
+        const isMax = result.result === this.faces;
+        const isMin = result.result === 1;
+        return [
+            this.constructor.name.toLowerCase(),
+            result.denomination ?? this.denomination, // Accomodating ComboDie as a result can have a different denomination than the die as a whole
```
CarlosFdez: It might be good to write this comment at the top, and add more details. I was actually uncertain about this method, and was about to suggest refactors for what is effectively with a copy paste with one replaced property.

You could maybe do this though, if it seems more clear. This may not need as much of an explanation. If this feels fragile to do we can keep what you have here, though I think the comment needs a bit more help.

```js
// ComboDie can have a different denomination than the die as a whole
const results = super.getResultCSS(result);
const termIdx = results.indexOf(`d${this.faces}`);
if (termIdx >= 0) results[termIdx] = result.denomination ?? this.denomination;
return results;
```

module/dice/die/baseDie.mjs
```diff
@@ -8,5 +17,199 @@ export default class BaseDie extends foundry.dice.terms.Die {
         const rerolledResult = this.results[this.results.length - 1];
         this.results.splice(this.results.length - 1, 1);
         this.results.splice(resultIndex, 0, rerolledResult);
+
+        if (['c', 'cc'].some(x => this.modifiers.includes(x))) {
+            await this.handleComboDiceReroll(resultIndex, result);
+        } 
+    }
+
+    /** @override */
```
CarlosFdez: Prefer `@inheritDoc`. inheritDoc implies override and is used when there are no other changes.

module/dice/die/baseDie.mjs
```diff
@@ -8,5 +17,199 @@ export default class BaseDie extends foundry.dice.terms.Die {
...
+           The actual rolls are done here in place so every dice gets the correct denomination.
+        */
+        if (game.modules.get('dice-so-nice')?.active) {
+            const resultsToRoll = this.results.filter((x, index) => { 
+                if (!x.active) return false;
+                if (rerollStartIndex) 
+                    return index === rerollStartIndex || index > initialResultsLength - 1;
+
+                return true;
+            });
```
CarlosFdez:
```suggestion
            const resultsToRoll = this.results.filter((x, index) =>
                x.active && (!rerollStartIndex || index === rerollStartIndex || index > initialResultsLength - 1)
            );
```

I looked at this more carefully since I was gonna comment on the no curly for rerollStartIndex but I think this might be the same truth table as the original?

module/dice/die/baseDie.mjs
CarlosFdez: I went ahead and fixed it myself!

CarlosFdez:
I fixed an error here, and I also did some suggestions here https://github.com/Foundryborne/daggerheart/pull/2101

module/dice/die/baseDie.mjs
CarlosFdez: I made it a separate PR.

CarlosFdez (approved):

PR #2063, opened 2026-07-05
[Fix] SRD Towering Stalk Fix

WBHarry:
Removed a faulty setup for the domain card Towering Stalk.

It's now set up so that it has a resource with a 1 max that refreshes every rest. That resource is needed for either action.

### DIFF
--- a/src/packs/domains/domainCard_Towering_Stalk_n0P3VS1WfxvmXbB6.json
+++ b/src/packs/domains/domainCard_Towering_Stalk_n0P3VS1WfxvmXbB6.json
@@ -25,14 +25,6 @@
             "scalable": false,
             "step": null,
             "consumeOnSuccess": false
-          },
-          {
-            "key": "resource",
-            "itemId": "n0P3VS1WfxvmXbB6",
-            "value": 1,
-            "scalable": false,
-            "step": null,
-            "consumeOnSuccess": false
           }
         ],
         "uses": {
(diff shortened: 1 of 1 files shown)


src/packs/domains/domainCard_Towering_Stalk_n0P3VS1WfxvmXbB6.json
```diff
@@ -140,7 +141,15 @@
...
       }
     },
-    "resource": null,
+    "resource": {
+      "type": "simple",
+      "value": 1,
+      "max": "1",
+      "recovery": "shortRest",
+      "progression": "decreasing",
+      "dieFaces": "d4",
```
CarlosFdez, line 150: Is this something just added by the data model?

WBHarry: No, it's the core of the change.

The Feature has 2 actions, and both are limited to 1 use per rest.
So I gave it a resource with a maximum of 1 that recharges on shortrest. Both actions cost 1 item resource.

CarlosFdez: The resource has a die size?

WBHarry: Oh, die size hm. Yeah it's not nullable as is, so it'll get added like that.

CarlosFdez (approved):

PR #2064, opened 2026-07-05
[Fix] DamageReductionDialog Conditions

WBHarry:
Fixed so that a character having ReduceSeverity or ThresholdImmunity rules are considered for popping the DamageReductionDialog.

Those rules are only used within the dialog, so because it was only opened if armor or stress was available for use, you could be out of armor and not get the benefit of those rules.
We -could- just apply them automatically in that case without a dialog since no choices are to be made - but I think it makes it a lot more clear what's happening to still pop the dialog so they can see that the severity is reduced etc.

### DIFF
--- a/module/documents/actor.mjs
+++ b/module/documents/actor.mjs
@@ -622,22 +622,30 @@ export default class DhpActor extends Actor {
         return rollData;
     }
 
-    #canReduceDamage(hpDamage, type) {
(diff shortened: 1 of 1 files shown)


CarlosFdez (approved):

CarlosFdez:
not tested but conceptually seems fine to me.

PR #2067, opened 2026-07-06
Fix some issues involving party in observer view

CarlosFdez:
Fixes tag team and group roll buttons not being accessible in party sheet in observer view despite the dialogs still working otherwise. Also hides the party removal button if the user doesn't have ownership.

There are still missing permission edge cases, such as cursors and hover highlight showing even when the user lacks permission over that actor. I think these should be handled separately and are low priority.

Also some jsdoc as well as tweaks in order to follow conventions.


WBHarry (approved):
Noice :+1:

PR #2068, opened 2026-07-07
Fix missing group roll hint message

CarlosFdez:
Caught this one while I was making a new dialog and referencing some old stuff

image


WBHarry (approved):
Ah, nice catch :+1:

PR #2069, opened 2026-07-07
[UI/UX] Minor tweaks to tooltip card styling

CarlosFdez:
While I moved sheet.less to index.less since sheet.less doesn't really make much sense there, the only change I did to the actual styling was reduce 16px padding bottom to 4px on the tags, and also remove the margin bottom from the help text and instead made it apply to the tooltip itself.

## Before

image
image

## After

image
image


WBHarry (approved):
:+1:

PR #2070, opened 2026-07-07
Support vaulting and unvaulting domain cards via dragdrop

CarlosFdez:
Hiding whitespace when reviewing makes the crln -> ln auto converted file (loadout.hbs) more readable.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -3233,7 +3233,7 @@
                 "beastformToManyAdvantages": "You cannot select any more advantages.",
                 "beastformToManyFeatures": "You cannot select any more features.",
                 "beastformEquipWeapon": "You cannot use weapons while in a Beastform.",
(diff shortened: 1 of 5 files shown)


CarlosFdez:
Eventually we can try to handle drag over to replace (only if loadout is full), but we may want a swap confirmation dialog with a recall cost maybe? And perhaps some drag over effects to highlight the card in red.

What I did in this PR was low hanging fruit.

WBHarry (approved):
Feels really nice. Good improvement :+1:

It does beg to question if we actually need to have the bespoke arrow buttons for `Send To Vault` and `Send To Loadout` anymore.
I guess it does have some value in that it makes it obvious to new users maybe? 

Either way, bring it in as you see fit 🙃

CarlosFdez:
If we need the space for something, I suppose we can remove the buttons, but otherwise yeah I guess it does make it obvious.

PR #2072, opened 2026-07-08
[Fix] Beastform TagTeam Options

WBHarry:
Fixed beastform basic attack being available. More specifically, unnarmed strike is available if it would be on the CharacterSheet sidebar - so also a fix for a character that is using their unnarmed strike in general.


CarlosFdez:
More accurate name for the changelog entry might be "Include unarmed attack in tag team options" but the actual code looks fine and dandy to me.

CarlosFdez (approved):

PR #2073, opened 2026-07-10
[Refactor] remove postEvaluate static method

CarlosFdez:
Needs some heavy testing, so far it seems fine tho,. but I might have missed a spot.

Attempts to remove one of the protected methods of the build pipeline to simplify matters, letting that task be performed by buildEvaluate(). There is one bit of duplication as a result (in damage), but I think in this case the duplication is easier to follow than what was there before.

It would be nice if there is a way to remove buildPost() (leaving the overrides to build()) and/or buildConfigure() but I don't quite know how to do those yet.

### DIFF
--- a/module/dice/d20Roll.mjs
+++ b/module/dice/d20Roll.mjs
@@ -185,8 +185,10 @@ export default class D20Roll extends DHRoll {
         return changeKeys;
     }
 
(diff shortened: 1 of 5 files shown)


WBHarry (approved):
:+1:

PR #2074, opened 2026-07-11
[Fix] TemplateEnricher Shape Data

WBHarry:
- Fixed a typo in getTemplateShape that was causing TemplateEnricher buttons to error. 
- Added stopPropagaton when clicking enriched buttons. This stops the description in chat messages from closing when clicking on one.


CarlosFdez:
Rather than click wrappers, perhaps we should add stop propagation to each event?

CarlosFdez (approved):
Actually nvm, its quite minor. Looks good to me.

PR #2077, opened 2026-07-11
Start supporting flat damage without custom formulas

CarlosFdez:
This allows a flat modifier of 0 and a bonus of 5 to print `5` instead of `0d6 + 5` in something like an adversary attack's damage formula.

This may be insufficiently exhaustive to be the new way of doing things for minions, but its a start.


WBHarry (approved):
Think it should work for minions, but can be tried in some SRD updates after the damage rework is in :+1:

PR #2079, opened 2026-07-11
[Rework] ChatMessage Damage

WBHarry:
Changed so that `ActorRoll.Damage` is hard defined in `ActorRoll.mjs` instead of just being a nebulous ObjectField.
It's intended to hold keys for the various damage types IE `hitPoints`, `stress` etc.
The keyed objects hold the unaugmented JSON for the damage roll, aswell as auxiliary data like damageTypes used for hitPoint damage. Previously we were saving a -lot- of information that could easily go out of data and was very hard to keep track of. Now we can reroll the unagumented roll straight up and not have issues.


module/applications/ui/chatLog.mjs
```diff
@@ -183,13 +183,8 @@ export default class DhpChatLog extends foundry.applications.sidebar.tabs.ChatLo
...
                 hitPoints: {
-                    parts: [
-                        {
-                            applyTo: 'hitPoints',
-                            damageTypes: [],
-                            total
-                        }
-                    ]
+                    damageTypes: [],
+                    roll: { total }
```
CarlosFdez: Personally I dislike comma delimited variable declares, and prefer if each line here started with const. Much easier to read on a skim. The actual change here is great though.

module/applications/ui/chatLog.mjs
```diff
@@ -259,23 +254,14 @@ export default class DhpChatLog extends foundry.applications.sidebar.tabs.ChatLo
...
-                };
-            });
-            const updateMessage = game.messages.get(message._id);
-            await updateMessage.update({
-                [`system.damage.${damageType}`]: {
-                    total: parsedRoll.total,
-                    parts: damageParts
-                }
+            const { damageType, dice, result } = target.dataset;
+            await message.system.damage.rerollDamageDice(damageType, dice, result);
```
CarlosFdez: Is this to reroll a single die? If so, I think rerollDamageDie might be more clear?

Inspecting the actual output of a damage roll though, it looks like each die when using a dagger with proficiency has the same index, so I don't know how well this is actually working:
Image

module/data/chat-message/actorRoll.mjs
```diff
@@ -29,6 +29,60 @@ export const originItemField = () =>
         actionIndex: new fields.StringField()
     });
 
+class ChatMessageRollDamage extends foundry.abstract.DataModel {
```
CarlosFdez: Thoughts on renaming to ChatDamageData and putting this in a new file (such as `chat-message/chatDamageData.mjs` or `chat-message/chat-damage-data.mjs`, not sure what convention we use, we have a mix) that we import here?

module/applications/ui/chatLog.mjs
WBHarry: That's expected. The way Foundry handles something like `3d6 + 4` is like this:
A roll instance is made with Terms [DieTerm, Operator, Number] (3d6, +, 4)
The DieTerm gets set up like `_faces=6`, `number=3`.

So in other words each grouping of dice are a single DieTerm. The individual rolls are the `die.results`
So the `data-dice` just points out the DieTerm. The `data-result` points out the individual die in that dice grouping.

module/data/chat-message/actorRoll.mjs
WBHarry: Yah, that's good. Moved it out :+1:

module/applications/ui/chatLog.mjs
WBHarry: Oh, I do too. Section not written by me originally and didn't change it.

Moved it out to individual lines 😎

WBHarry:
I realised now that there are needed changes to TagTeamDialog aswell since it handles and saves damage data.
Going to have to go over that and migrate that data aswell

module/applications/ui/chatLog.mjs
CarlosFdez: hmm, individual die is likely gonna be an issue visually if there are matches I guess, but perhaps that's unavoidable. That makes sense though.

WBHarry: Yeah, going to have to make our own reroll solution to handle that. Foundry's solution just doesn't care about preserving dice order.

WBHarry: I added a `rerollResult` method to a `BaseDie` class that all our dies extend now.
It rerolls a specific `die.result` and manipulates the results array so that the order is preserved. 😎

module/data/chat-message/chatDamageData.mjs
```diff
@@ -0,0 +1,56 @@
...
+    get active() {
+        return Boolean(Object.keys(this.types).length);
+    }
+
+    static #validateRoll(rollJSON) {
+        const roll = JSON.parse(rollJSON);
+        if (!roll.evaluated) throw new Error('Roll objects added to ChatMessage documents must be evaluated');
+    }
+
+    prepareRolls() {
```
CarlosFdez: I'm actually unsure about this function. What is its purpose. It doesn't return anything, so what is it actually doing? Is there any way we can keep this internal to this class rather than exposed somehow?

If may have a purpose, but its not commented so I can't tell :P

WBHarry: The purpose is to transform the JSONField `roll` field to an actual Roll instance in-memory, the same way done for `ChatMessage.rolls`. `ChatDamageData` is a `DataModel`, so I set up a function that handles it's data prep, but then it needs to be called during `prepareDerivedData` on any `TypeDataModel` that makes use of `ChatDamageData`
image

CarlosFdez: fromData has side effects? That might need a comment, that's strange behavior.

You could do type.roll = Roll.fromData(type.roll) maybe to make it more intuitive? I don't know what its doing internally.

WBHarry: Don't think fromData has any side effects. It's just a method to generate a roll from an object/JSON. 
I moved it into just directly assigning on `type.roll`

module/systemRegistration/migration-handlers/2_6_0.mjs
```diff
@@ -0,0 +1,20 @@
+import { MigrationHandlerBase } from './base.mjs';
+
+export class Migration_2_6_0 extends MigrationHandlerBase {
+    version = '2.6.0';
+
+    /** @inheritdoc */
+    async updateActorSource(actor) {
+        if (actor.type === 'party') {
```
CarlosFdez: Would it be possible to check that a tagTeam exists, to avoid the update if there isn't one?

module/data/chat-message/chatDamageData.mjs
CarlosFdez: I misread lmao, my bad. Looked like you were deserializing then doing nothing with it.

This can't be done during construction?

module/systemRegistration/migration-handlers/2_6_0.mjs
WBHarry: Good call. Fixed :+1:

module/data/chat-message/chatDamageData.mjs
WBHarry: Huh, waddya know. Moved it to there, and seems good :+1:

PR #2080, opened 2026-07-12
Adjust styling of secret blocks

CarlosFdez:
Attempts to clean up the styling of secret blocks. This isn't my preference long term for environment feature notes, but it is a start. Generally pretty similar to the 5e one, though I also made the button only render when hovering.

image

Note that toggling reveal/hide doesn't work from the actor sheet. Given this also happens in 5e, I suspect its a foundry bug.


WBHarry (approved):
Awesum :+1:

PR #2081, opened 2026-07-12
[Fix] chat messages with list items or weapon/armor features

CarlosFdez:
Fixes item chat descriptions for items that have list items (such as from some environments) or have addends or prepends (such as weapon/armor abilities). There's also a minor beneath notice styling change where descriptions now have a nicer gap from the header (margins overlap, paddings don't).

This is something I caught while testing gm notes (ongoing feature im working on)


WBHarry (approved):
:+1:

PR #2082, opened 2026-07-13
[Feature] Add support for GM Notes

CarlosFdez:
image


CarlosFdez:
Once https://github.com/Foundryborne/daggerheart/pull/2083 is in I'll convert the environments to use gm notes instead of secret blocks for the italic hint text.

CarlosFdez:
I am mostly missing handling for GM notes in actions. We could omit them, but perhaps we may want it in the case where a description is inherited from an item?

module/applications/sheets/api/base-item.mjs
```diff
@@ -76,10 +77,16 @@ export default class DHBaseItemSheet extends DHApplicationMixin(ItemSheetV2) {
     /**@inheritdoc */
     async _preparePartContext(partId, context, options) {
         await super._preparePartContext(partId, context, options);
+        const TextEditor = foundry.applications.ux.TextEditor.implementation;
 
         switch (partId) {
             case 'description':
-                context.enrichedDescription = await this.document.system.getEnrichedDescription();
+                context.enrichedDescription = await this.document.system.getEnrichedDescription({ gmNotes: false });
+                context.enrichedGMNotes = await TextEditor.implementation.enrichHTML(this.item.system.gmNotes, {
```
WBHarry, line 85: `TextEditor.implementation` === `TextEditor` in this case, right?

WBHarry:
Looks great. Handles smooth. ⭐ 

I do think that since you did go ahead with adding the GM notes into the chat message, it would make sense to go ahead and inherit it into actions somehow probably, but that could be seen as a future improvement.

WBHarry (approved):

module/applications/sheets/api/base-item.mjs
CarlosFdez: yeah, I did shadow it. If its confusing I could perhaps rename it. Or let the line go for long and force a wrap. What do you think?

CarlosFdez:
I suppose let me know if you plan to do a small patch release first. If so we probably shouldn't bring it in yet.

We can see how it feels for actions later.

WBHarry:
Released a patch, so we should be in the clear now for letting things build up in main for a bit :+1:

PR #2083, opened 2026-07-13
Remove duplicate action descriptions in environments

CarlosFdez:
I believe these weren't caught in a previous sweep because of the secret blocks. I made a script in foundry to compare ignoring the secret id.


WBHarry (approved):
:+1:

PR #2086, opened 2026-07-13
[Fix] Summon Wildcard Handling

WBHarry:
- Fixed to using Foundry's inbuilt tokenDocument data creator (`actor.getTokenDocument()`) so that wildcard token image paths get handled as expected. It was previously erroring out on wildcard paths since it was just trying to resolve an image on that exact path, which doesn't exist.
- Fixed ChatMessage data for summon actions. The summary text in the chat message for how many summons were created wasn't being passed along.
image
I fixed this by passing along the data in the config. Unsure if we could do it some other way.

### DIFF
--- a/module/documents/tokenManager.mjs
+++ b/module/documents/tokenManager.mjs
@@ -22,7 +22,7 @@ export default class DhTokenManager {
         return await canvas.tokens.placeTokens(
             [
                 {
-                    ...actor.prototypeToken.toObject(),
+                    ...(await actor.getTokenDocument()).toObject(),
                     actorId: actor.id,
                     displayName: 50,
                     ...tokenData
--- a/module/documents/tokenManager.mjs
+++ b/module/documents/tokenManager.mjs
@@ -19,7 +19,7 @@ export default class DhTokenManager {
             }
         }
 
-        return await canvas.tokens.placeTokens(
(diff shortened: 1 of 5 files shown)


module/documents/tokenManager.mjs
```diff
@@ -19,17 +19,21 @@ export default class DhTokenManager {
...
                     displayName: 50,
                     ...tokenData
                 }
             ],
             { create: false }
         );
+
+        if (!placedData.length) return null;
+
+        return placedData[0];
```
CarlosFdez:
```suggestion
        return placedData[0] ?? null;
```
If placedData can be null, use placedData?.[0] ?? null instead.

CarlosFdez (approved):

CarlosFdez:
Seems to work well. Just tested with some wildcard tokens.

module/documents/tokenManager.mjs
WBHarry: image

Yah it's possible for null to be sent back. But that syntax handles it :+1:

CarlosFdez: is placedData itself null, or is it a list containing null/empty list? Neither of what we did would handle the first, but it would handle the latter.

PR #2087, opened 2026-07-13
Add guard for null placedData

CarlosFdez:


WBHarry (approved):
Oh, huh. I thought I saw a commit pushed that did this. Oh wells. Good catch.

PR #2088, opened 2026-07-13
[Fix] issues with secret block styling

CarlosFdez:
Not to be confused with secret fixes, just secret block fixes

### DIFF
--- a/styles/less/global/elements.less
+++ b/styles/less/global/elements.less
@@ -599,27 +599,23 @@
     }
 
     secret-block {
-        position: relative;
-        section.secret {
-            background-color: @red-10;
-            padding: 0;
-            margin-top: 0.375rem;
-            &.revealed {
-                background-color: @green-10;
-            }
-            p {
-                margin: 0.5rem 0;
-            }
-        }
+        display: block;
+
+        /** A buffer to make the hover behavior work a bit better. The bottom in the button needs to compensate */
+        @buffer: 8px;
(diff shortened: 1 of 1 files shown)


WBHarry:
Hm. Revealing text in the chat messages have an issue atm as the event propagation isn't stopped - so the description they're in gets minimized

CarlosFdez:
Huh that's weird. That doesn't happen to me, funny enough. Any thoughts on how to fix it given its a foundry thing, and its only happening to you?

WBHarry:
Kinda cray
image

CarlosFdez:
https://github.com/user-attachments/assets/c3ae8ea2-78bf-41fd-a55a-b2bae7e60452

WBHarry:
Daheck

https://github.com/user-attachments/assets/7ccf593d-812a-43d1-a44d-11c18d3614c1

CarlosFdez:
I am testing in chrome if that helps. But even then clicking the body of the chat message doesn't close it for me normally. Does it for you?

WBHarry:
Nah, there's no general close functionality in the section, you have to click on the header part.
Got no clue what's making the chat message rerender. I'm also doing it on chrome 🙃

WBHarry:
The chatmessage timestamp gets updated, so there is definately a rerender happening

WBHarry (approved):
Unrelated bug that happens when auto-expand for chat message descriptions isn't on

PR #2089, opened 2026-07-14
[Fix] Preserve description expand state on re-render

CarlosFdez:
any `<details>` elements are not preserved by foundry, and in general it was previously rejected: https://github.com/foundryvtt/foundryvtt/issues/12063

So we gotta do it ourselves.


WBHarry (approved):
Awesome digging on this one :+1:

PR #2090, opened 2026-07-15
[Feature] Allow deleting countdowns via right click context menu on the main panel

CarlosFdez:
image

It also refactors the countdown rendering flow to occur based on the data updating, not a hook. There may be a double render somewhere I forgot to remove, but so far it seems to behave fine. There's also a change where the animation won't play on start updates, so that the initial configuration won't spam the animation.


CarlosFdez:
On the actual code itself, I don't know the purpose of `this.data` on the editor. Removing it, or having it fetch it again on re-render, might simplify things if we don't need to decouple it? Are there some temporary settings that don't lead to an update right away?

WBHarry (approved):
Seems all good. Good improvement :+1:

PR #2092, opened 2026-07-16
[Feature] Show item tooltips for item use macros

CarlosFdez:
Also splits off existing tooltip behaviors into other functions, so please test to make sure I didn't breaky anything.


WBHarry (approved):
Everything seems to be working well.

Kudos on the switch case and function split of the code. Tech debt that had built up here from a simple beginning 🙃

PR #2094, opened 2026-07-17
[Rework] Damage and Damage Resource split

WBHarry:
Initial data model update.

This branch is intended to take in separate work on making it happen before being merked into the initial rework branch.


CarlosFdez (approved):

module/dice/dhRoll.mjs
```diff
@@ -41,6 +41,10 @@ export default class DHRoll extends BaseRoll {
         return config;
     }
 
+    static createRollInstance(config) {
```
CarlosFdez, line 44: What's the purpose of adding this object? If needed, should we call it something like createInstance instead?

module/documents/actor.mjs
```diff
@@ -656,62 +656,66 @@ export default class DhpActor extends Actor {
...
-            if (key === CONFIG.DH.GENERAL.healingTypes.hitPoints.id)
-                damage.total = this.calculateDamage(damage.total, damage.damageTypes);
-            const update = updates.find(u => u.key === key);
-            if (update) {
-                update.value += damage.total;
-                update.damageTypes.add(...new Set(damage.damageTypes));
-            } else updates.push({ value: damage.total, key, damageTypes: new Set(damage.damageTypes) });
-        });
+        if (damages.main) {
+            damages.main.total = this.calculateDamage(damages.main.total, damages.main.damageTypes);
```
CarlosFdez, line 660: Personally I don't like that we're doing this side-effect, but thiis is what it did before and I suspect it was because of postCalculateDamage. I'd like to know what's using the hook in the wild so we can give it with something more useful like the updates array.

module/documents/actor.mjs
CarlosFdez, line 651: Personally given this is all changing anyways, I am considering changing this to an object but I'm not certain yet.

PR #2095, opened 2026-07-18
[Fix] chat being jumpy as timestamps get longer

CarlosFdez:
Fixes this funky issue
https://github.com/user-attachments/assets/4c0a2550-a62b-4710-8664-611992854d75

Looks like this now when the timestamp is long
image

Doesn't seem to break the UI modules I messed with, at least not in any obvious way. Because of the way chat is rendered, this doesn't break any existing chat messages.


WBHarry (approved):
Neato :+1:

PR #2096, opened 2026-07-19
[Fix] 2094 - Critical Damage

### LINKED ISSUE #2093: [Bug] Critical Damage gets hardbaked after initial roll
To be resolved after the [Damage rework](https://github.com/Foundryborne/daggerheart/pull/2079) PR is in

----
Currently when a roll is added to a chatMessage the modifiers for the initial roll result get baked into the formula and used as is.
This means that if you have a damage formula of "1d6+2" and the attack roll was a critical, then the stored formula will instead be "1d6+8". If you later reroll the main roll so that it's no longer a critical, the damage will not be corrected.

TagTeamDialog is currently in a similar state. The total damage displayed in the Results section aswell as used once it's finalised has to be checked depending on whether it's a critical or not.

----

My thinking is that we should always store the original formula and not modify it based on crit state. Instead the correct formula and modifiers to display in various places should be calculated in some easy to access fashion so we can show it in chat messages etc.

WBHarry:
Fixes #2093 

- Fixed so that DamageRolls are actualy with class='Damage Roll' in the chatmessage. Previously they stored 'BaseRoll', and therefore got instantiated as that, giving us no access to DamageRoll methods.
- Fixed so that DamageRolls do not get saved with a formula including the crit damage. I opted to just redo the formulas when the DamageDialog submits. I think this is the most stable so that the formula shown while in the dialog will be showing crit damage.
image
- DamageRoll now overrides the `total` and `modifierTotal` getters to retrieve critical damage when appropriate. The formula and terms are never changed. 
- Changed the utility method GetCritDamageBonus to not be async. Led to some `await` removals.

### DIFF
--- a/module/applications/dialogs/damageDialog.mjs
+++ b/module/applications/dialogs/damageDialog.mjs
@@ -121,7 +121,14 @@ export default class DamageDialog extends HandlebarsApplicationMixin(Application
     }
 
     static async submitRoll() {
+        const isCritical = this.config.isCritical;
         const { damageFormula, resourceFormulas } = this.roll.constructFormulas({ ...this.config, isCritical: false }); 
+        /* Sideeffect occuring in constructFormulas that sets this.config.isCritical to the false value. Can remove the below if it can be prevented */
+        this.config.isCritical = isCritical;
+        damageFormula.roll.options.isCritical = isCritical;
+        for (const formula of resourceFormulas)
+            formula.roll.options.isCritical = isCritical;
+
         this.config.damageFormula = damageFormula;
         this.config.resourceFormulas = resourceFormulas;
         await this.close({ submitted: true });
--- a/module/applications/dialogs/damageDialog.mjs
+++ b/module/applications/dialogs/damageDialog.mjs
@@ -6,6 +6,7 @@ export default class DamageDialog extends HandlebarsApplicationMixin(Application
 
         this.roll = roll;
         this.config = config;
+        this.originalIsCritical = config.isCritical;
         this.selectedEffects = this.config.bonusEffects;
     }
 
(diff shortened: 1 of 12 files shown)


module/applications/dialogs/damageDialog.mjs
```diff
@@ -6,6 +6,7 @@ export default class DamageDialog extends HandlebarsApplicationMixin(Application
 
         this.roll = roll;
         this.config = config;
+        this.originalIsCritical = config.isCritical;
```
CarlosFdez, line 13: Might be worth adding a `/** blah blah */` doc comment above this line. JSDoc is smart enough to add the comment to intellisense used elsewhere.

CarlosFdez: Coming in I don't know why this variable exists. I know what it does, just not why.

module/applications/dialogs/tagTeamDialog.mjs
```diff
@@ -696,6 +677,12 @@ export default class TagTeamDialog extends HandlebarsApplicationMixin(Applicatio
                 }
             }
 
+            if (mainRoll.damageRollData.main) {
+                const isCritical = overrideIsCritical ?? mainRoll.roll.isCritical;
+                mainRoll.damageRollData.main.options.isCritical = isCritical;
+            }
+
+
```
CarlosFdez:
```suggestion
```

module/applications/dialogs/damageDialog.mjs
WBHarry: The logic was actually wrong. Must've mixed the streams at the end. I pushed a change.🙃 

- I tested all the cases of flipping the `isCritical` switch in the Damage Dialog with this, and it turned out right.
- Added a comment in the constructor for the `originalIsCrit`.

module/data/chat-message/chatDamageData.mjs
```diff
@@ -28,7 +41,15 @@ export class ChatDamageData extends foundry.abstract.DataModel {
...
-        this.main &&= Roll.fromData(this.main);
+        this.main &&= Roll.fromData({
+            ...this.main, 
+            options: { 
+                ...this.main.options, 
+                isCritical: 
+                    this.main.options.isCritical === false ? false : (this.main.options.isCritical || this.isCritical) 
+            } 
+        });
+
```
CarlosFdez, line 52: Might be less lines if you use foundry.utils.mergeObject().

CarlosFdez: so maybe like, uh, this? Double check I didn't screw up the truth table of the value
```js
this.main = Roll.fromData(
  foundry.utils.mergeData(this.main, {
      "options.isCritical": Boolean(this.main.options.isCritical ?? this.isCritical)
  }, { inplace: false }
);
```

CarlosFdez: Well, actually, you're out of time, and this is minor at best.

CarlosFdez (approved):

PR #2097, opened 2026-07-21
[Rework] Fear Tracker

moliloo:
## Description

- rework fear tracker
- enhance token and bar display
- create custom dragger and resize methods

## Type of Change

- [x] New feature

## How Has This Been Tested?

- [x] Manual testing

image
image

## Checklist

---

### DIFF
--- a/module/applications/ui/fearTracker.mjs
+++ b/module/applications/ui/fearTracker.mjs
@@ -13,6 +13,14 @@ const { HandlebarsApplicationMixin, ApplicationV2 } = foundry.applications.api;
 export default class FearTracker extends HandlebarsApplicationMixin(ApplicationV2) {
     constructor(options = {}) {
         super(options);
+
+        this._dragData = {
+            isDragging: false,
+            startX: 0,
+            startY: 0,
+            startLeft: 0,
+            startTop: 0
+        }
     }
 
     /** @inheritDoc */
@@ -21,19 +29,20 @@ export default class FearTracker extends HandlebarsApplicationMixin(ApplicationV
         classes: [],
         tag: 'div',
         window: {
-            frame: true,
+            frame: false,
             title: 'DAGGERHEART.GENERAL.fear',
             positioned: true,
             resizable: true,
             minimizable: false
         },
+        classes: ['daggerheart', 'dh-style', 'fear-tracker'],
         actions: {
             setFear: FearTracker.setFear,
             increaseFear: FearTracker.increaseFear
         },
         position: {
-            width: 222,
-            height: 222
+            width: 540,
+            height: 'auto'
         }
     };
 
@@ -53,6 +62,10 @@ export default class FearTracker extends HandlebarsApplicationMixin(ApplicationV
         return game.settings.get(CONFIG.DH.id, CONFIG.DH.SETTINGS.gameSettings.Homebrew).maxFear;
     }
 
+    get fearPosition() {
+        return game.settings.get(CONFIG.DH.id, CONFIG.DH.SETTINGS.gameSettings.appearance).fearPosition;
+    }
+
     /* -------------------------------------------- */
     /*  Rendering                                   */
     /* -------------------------------------------- */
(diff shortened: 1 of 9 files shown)


WBHarry:
Going to check the code, but some general thoughts/findings after trying it:
**Positioning**
Bottom + Left: Would it be more logical to slot the fear application above the player list, rather than below?
image

**Minor Bug**
I found a funny behavior, the container "becomes afraid" if you hover in the bottom right corner. Hover behavior probably not on the outmost container or some such.

https://github.com/user-attachments/assets/96323b9b-f85d-48dd-b813-c7fa6e3ef64b

**Possible Further Refinement After PR Or Some Such**
I think it might be nicer for the bar render to not show the background at all. Maybe have a dedicated **Fear** title above it or some such?
Could also be neater to either remove the `+ -` buttons, or to place them as indicators on the bar itself - while the behavior becomes left click to increase, rightclick to decrease. 
image

moliloo:
**Positioning**
we can let it above players sure, i was undecided if it was better above or bellow it

**Minor Bug**
Thats curious, i think that was happening because the icon as rotated 45deg

**Possible Further Refinement After PR Or Some Such**
i can remove the background if you think that was better, i just not remove thinking would be good have some 'container' around the bar. About the buttons yeah, could be good move inside the bar, but im unsure about the left click to add and right one to decrease, buttons are more intuitive.

This just was a mock i do changing the elements styles in browser to remove the background 
image

module/applications/ui/fearTracker.mjs
```diff
@@ -63,15 +76,51 @@ export default class FearTracker extends HandlebarsApplicationMixin(ApplicationV
...
+                break;
+                
+            default:
+                this.handleStyleElement(fearPosition);
+                document.body?.appendChild(this.element);
+                const position =
+                    game.user.getFlag(CONFIG.DH.id, 'app.resources.position') ?? FearTracker.DEFAULT_OPTIONS.position;
+                this.setPosition(position);
+                break;
+        }
```
WBHarry, line 120:
```suggestion
        this.handleStyleElement(fearPosition);
        switch (fearPosition) {
            case 'topCenter':
                document.getElementById('ui-top')?.appendChild(this.element);
                break;
            case 'bottomCenter':
                document.getElementById('ui-bottom')?.prepend(this.element);
                break;
            case 'rightTop':
                document.getElementById('ui-right-column-1')?.appendChild(this.element);
                break;
            case 'leftBottom':
                document.getElementById('ui-left-column-1')?.appendChild(this.element);
                break;
            default:
                document.body?.appendChild(this.element);
                const position =
                    game.user.getFlag(CONFIG.DH.id, 'app.resources.position') ?? FearTracker.DEFAULT_OPTIONS.position;
                this.setPosition(position);
                break;
        }
```

module/applications/ui/fearTracker.mjs
```diff
@@ -80,13 +129,39 @@ export default class FearTracker extends HandlebarsApplicationMixin(ApplicationV
...
+
+        const containsStyle = position => this.element.classList.contains(position);
+
+        Object.entries(positions).forEach(([key, value]) => {
+            if (containsStyle(value)) this.element.classList.remove(value);
+        })
+
+        this.element.classList.add(positions[fearPosition]);
     }
```
WBHarry, line 135:
```suggestion
    handleStyleElement(fearPosition) {        
        for (const position of Object.values(CONFIG.DH.GENERAL.fearPosition)) {
            this.element.classList.remove(position.value);
        }

        this.element.classList.add(fearPosition);
    }
```

WBHarry (changes requested):

WBHarry:
It seems very exessive to have to redo all of Foundry's reposition and resize logic.
I gave it a brief try to manage and plug into the existing logic, but I didn't quite manage it just now.

I think it's fine to bring in as is, and it can be seen as a todo to remove a lot of code.

WBHarry:
> **Positioning** we can let it above players sure, i was undecided if it was better above or bellow it
> 
I -think- it's better at least. I think it's generally best to try and not move Foundry's base elements when possible.

> **Possible Further Refinement After PR Or Some Such** i can remove the background if you think that was better, i just not remove thinking would be good have some 'container' around the bar. About the buttons yeah, could be good move inside the bar, but im unsure about the left click to add and right one to decrease, buttons are more intuitive.
>
I'm not a 100% on this one, but it's a thought I had. It could definately be left as a future improvement/change we try to align on.

module/data/settings/Appearance.mjs
```diff
@@ -41,6 +41,11 @@ export default class DhAppearance extends foundry.abstract.DataModel {
                 choices: CONFIG.DH.GENERAL.fearDisplay,
                 initial: CONFIG.DH.GENERAL.fearDisplay.token.value
             }),
+            fearPosition: new StringField({
+                required: true,
+                choices: CONFIG.DH.GENERAL.fearPosition,
+                initial: CONFIG.DH.GENERAL.fearPosition.free.value
```
WBHarry: Suggesting to set the default to topCenter

```suggestion
                initial: CONFIG.DH.GENERAL.fearPosition.topCenter.value
```

WBHarry (approved):
:+1:

PR #2098, opened 2026-07-21
[Fix] Damage Related Getters

WBHarry:
Fixed so that non attack/damage actions that don't have `this.damage` don't error out when used.

We could also opt to change the `get hasDamage` to check specifically that the type is in `['damage', 'attack']` or some such instead if preferable.


CarlosFdez:
No, I think this is perfect.

CarlosFdez (approved):

PR #2100, opened 2026-07-21
[Feature] SRD Mark Hit Points Update

WBHarry:
Updated SRD entries that should use the new mark hit points possibility of the damage rework.

### DIFF
--- a/src/packs/domains/domainCard_Onslaught_I7pNsQ9Yx6mRJX4V.json
+++ b/src/packs/domains/domainCard_Onslaught_I7pNsQ9Yx6mRJX4V.json
@@ -14,7 +14,7 @@
         "type": "damage",
         "_id": "MxaqNvY9IfWnFe5P",
         "systemPath": "actions",
-        "description": "<p class=\"Body-Foundation\">When a creature within your weapon’s range deals damage to an ally with an attack that doesn’t include you, you can <strong>mark a Stress</strong> to force them to make a Reaction Roll (15). On a failure, the target must mark a Hit Point.</p>",
+        "description": "<p>When a creature within your weapon’s range deals damage to an ally with an attack that doesn’t include you, you can <strong>mark a Stress</strong> to force them to make a Reaction Roll (15). On a failure, the target must mark a Hit Point.</p>",
         "chatDisplay": true,
         "actionType": "action",
         "cost": [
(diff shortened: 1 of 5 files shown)


CarlosFdez:
Hard for me to review, the valueAlt keeps throwing me for a loop. What decides if valueAlt gets used again?

WBHarry:
Good question, I don't think it ever -is- used in any general situation. 
It's used to store the alternative Horde Damage, but not sure if there is any other case.

WBHarry:
Suffice to say, it's not an intentional addition on my part, updating the data just added that on it's own due to the data model.

CarlosFdez:
We'll have to look at how to clean that up in a future date I suppose.

src/packs/domains/domainCard_Onslaught_I7pNsQ9Yx6mRJX4V.json
```diff
@@ -14,7 +14,7 @@
         "type": "damage",
         "_id": "MxaqNvY9IfWnFe5P",
         "systemPath": "actions",
-        "description": "<p class=\"Body-Foundation\">When you successfully make an attack with your weapon, you never deal damage beneath a target’s Major damage threshold (the target always marks a minimum of 2 Hit Points).</p><p class=\"Body-Foundation\">Additionally, when a creature within your weapon’s range deals damage to an ally with an attack that doesn’t include you, you can <strong>mark a Stress</strong> to force them to make a Reaction Roll (15). On a failure, the target must mark a Hit Point.</p>",
+        "description": "<p class=\"Body-Foundation\">When a creature within your weapon’s range deals damage to an ally with an attack that doesn’t include you, you can <strong>mark a Stress</strong> to force them to make a Reaction Roll (15). On a failure, the target must mark a Hit Point.</p>",
```
CarlosFdez:
```suggestion
        "description": "<p>When a creature within your weapon’s range deals damage to an ally with an attack that doesn’t include you, you can <strong>mark a Stress</strong> to force them to make a Reaction Roll (15). On a failure, the target must mark a Hit Point.</p>",
```

Might as well clean up css modified by copy pasta from rich text while we're at it.

WBHarry: 🍝 👨‍🍳

CarlosFdez (approved):
I'm not really equipped to evaluate this much json especially with all that valueAlt noise in there, so uh, looks good to me!

WBHarry:
I think that's fair enough. It's not manually edited JSON to begin with 🙃

PR #2101, opened 2026-07-22
Suggestions for ComboDice

CarlosFdez:


templates/ui/chat/parts/damage-part.hbs
```diff
@@ -57,13 +58,13 @@
...
-                    {{#each results}}
+                {{#each roll.dice as |dice diceKey|}}
+                    {{#each results as |dieResult resultKey|}}
                         {{#if active}}
                             <div class="roll-die{{#unless @../first}} has-plus{{/unless}}">
                                 <div 
-                                    class="dice reroll-button {{ifThen denomination denomination ../denomination}}" 
-                                    data-type="damage" data-damage-type="{{@../../key}}" data-dice="{{@../key}}" data-result="{{@key}}" {{#if ../../isResource}}data-is-resource="true"{{/if}}
+                                    class="dice reroll-button {{coalesce dieResult.denomination dice.denomination}}" 
+                                    data-type="damage" data-dice="{{diceKey}}" data-result="{{resultKey}}" {{#if ../../isResource}}data-is-resource="true"{{/if}}
```
CarlosFdez: data-damage-type used to point to "hitPoints" or "stress". That wasn't really a damage type, so I replaced it with a top level data-resource on the fieldset.

WBHarry: The code in `chatlog.mjs` hasn't been changed to go along with this. Currently you can't reroll a single dice from `damage.resources` in this branch because it's trying to use `data-damage-type` to know which resource the reroll is under.
```js
if (target.dataset.type === 'damage') {
            const { isResource, damageType, dice, result } = target.dataset;
            await message.system.damage.rerollDamageDie(isResource, damageType, dice, result);
```

WBHarry (changes requested):
Otherwise great stuff. A lot of simplifications, and coalesce is a neat helper 🧑‍🍳

CarlosFdez:
so data-damage-type is blank for main damage, so I think there might actually be a bug there then.

WBHarry:
Not a bug as such, but it could probably be simplified down to just one `data-resource` property like you did.
Atm it's like this:
image
(`ChatDamageData.rerollDamageDie`)

CarlosFdez:
Ah I see, not a resource is handled in another place, so rerolling main isn't broken. I think we can make it more clear, give me a second.

CarlosFdez:
Actually because of the tag team damage parts, I kept it to data-damage-type. But hopefully being in the if should make it more intuitive to the reader.

WBHarry (approved):
:+1:

PR #2102, opened 2026-07-22
[Refactor] Simplify header links in the char sheet

CarlosFdez:


CarlosFdez:
I was considering the possibility of making these have tooltips, but that doesn't work yet, so only the inline partial for now.

WBHarry (approved):
Beautiful :+1:

PR #2103, opened 2026-07-22
[Fix] Reroll Fixes

WBHarry:
Fixed so rerollDamage grabs the right dice for DiceSoNice. Fixed so that non-damage rerolls work again

- baseDie.rerollResult shuffles around the results, and the rerollDamageDie method wasn't accounting for that. Fixed.
- the `rerollEvent` method wasn't using a universal enough selector for it's target, so non-damage rerolls errored out. 
- Fixed so the reroll marking visualisations show up again

### DIFF
--- a/module/applications/ui/chatLog.mjs
+++ b/module/applications/ui/chatLog.mjs
@@ -247,7 +247,7 @@ export default class DhpChatLog extends foundry.applications.sidebar.tabs.ChatLo
         }
 
         const message = game.messages.get(messageData._id);
-        const target = event.target.closest('[data-result]');
+        const target = event.target.closest('[data-type]');
 
         if (target.dataset.type === 'damage') {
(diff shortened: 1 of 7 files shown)


module/applications/ui/chatLog.mjs
```diff
@@ -268,6 +268,8 @@ export default class DhpChatLog extends foundry.applications.sidebar.tabs.ChatLo
                     isReaction: message.system.roll.options.actionType === 'reaction'
                 }
             });
+            rerollDice.results[rerollDice.results.length - 1].rerolled = true;
```
WBHarry, line 271: This could probably be handled in a better place than here. Could be looked at some more. But at least it'll make any click to individiually reroll a hope or fear dice show up as rerolled 🤷‍♂️

CarlosFdez (approved):

PR #2105, opened 2026-07-22
[Fix] BaseAction Getter Fixes

WBHarry:
Fixed hasDamage and hasHealing getters to work correctly with empty resources.

Before the fix, no damage.main and empty resources still led to `hasDamage = true`. This made the `damageField.execute` continue for an AttackAction that only had a save without any damage, having the action chain error out.


CarlosFdez (approved):
👍

PR #2106, opened 2026-07-23
[Fix] Combat Turn Sounds

WBHarry:
If you had an active encounter and combat sounds turned on you get the current issue;
- Spotlight something -> Remove the spotlight -> `nextUp` sounds plays. 
 
 ----
 **Fixes**
- NextCombatant is not a concept in daggerheart. Set the getter to null.


CarlosFdez:
Can you give more details on the _onUpdate() change? That's the part that concerns me. For point two, did you check if the sound is made from the player side? Looking at the original code, the sound plays for whomever has the turn. So the player should hear the sound if its a player getting spotlit, and the GM should hear the sound if the GM is getting spotlight.

WBHarry:
Removed the second check. Correctly identified as not an issue. 
GM gets a sound if a not player-owned actor gets the spotlight - Player gets a sound if it's their player owned actor being given the spotlight.

CarlosFdez (approved):
Looks like some extract stuff slipped in but no problem

PR #2108, opened 2026-07-23
[Fix] fear tracker top center/right positioning in some situations

CarlosFdez:
https://github.com/user-attachments/assets/168e97d3-099f-4579-a13b-7e7d01fa4904

image


WBHarry (approved):
Awesome :+1:

PR #2109, opened 2026-07-23
[Fix] fear tracker offset on appearance change

CarlosFdez:
Fixes a particular case when switching from top right position (2 rows) to top center (1 row) on a 1080p monitor while the sidebar is open and a creature with effects is selected. While it would rectify itself on any change, giving the user the impression its broken is not ideal.


WBHarry (approved):
Sneaky sneaky countdown 🥷

PR #2110, opened 2026-07-23
[Fix] Improve HP Damage Multiplier Active Effect

WBHarry:
- Fixed so that HpDamageMultiplier ActiveEffect actually modifiers the formula so doubled, halve etc damage is actually displayed instead of being invisible
- This active effect now shows up as a toggleable alternatie in the Damage Roll Dialog

image

### DIFF
--- a/module/dice/damageRoll.mjs
+++ b/module/dice/damageRoll.mjs
@@ -119,7 +119,9 @@ export default class DamageRoll extends DHRoll {
 
     getActionChangeKeys() {
         const type = this.options.messageType ?? (this.options.hasHealing ? 'healing' : 'damage');
-        const changeKeys = [];
+        const changeKeys = [
+            'system.rules.attack.damage.hpDamageMultiplier'
+        ];
 
         for (const damageType of this.options.damageFormula?.damageTypes?.values?.() ?? []) {
             changeKeys.push(`system.bonuses.${type}.${damageType}`);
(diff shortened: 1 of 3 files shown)


module/dice/dhRoll.mjs
```diff
@@ -248,6 +248,22 @@ export default class DHRoll extends BaseRoll {
         });
     }
 
+    getTotalBonus(path) {
```
CarlosFdez, line 257: Can you add a comment to both this and getBonus()? I can't tell the difference at a glance.

CarlosFdez:
Conflicts!

module/dice/dhRoll.mjs
WBHarry: Done!

Honestly, the names could be changed to be better. They're pretty complex things to try and give a name to though.
I figured `getTotalBonus` could be useful for other things soon enough, so might aswell make it a general function.

module/dice/damageRoll.mjs
```diff
@@ -210,6 +212,22 @@ export default class DamageRoll extends DHRoll {
                     formulaData.roll.terms.push(...this.formatModifier(total));
                 }
             }
+
+            const damageTakenMultiplier = this.getTotalBonus('system.rules.attack.damage.hpDamageMultiplier');
+            if (damageTakenMultiplier && damageTakenMultiplier !== 1) {
```
CarlosFdez, line 217: why is -1 specifically the not allowed one? If its that it must be positive, maybe `damageTakenMultiplier > 0`?

module/dice/dhRoll.mjs
```diff
@@ -248,6 +248,35 @@ export default class DHRoll extends BaseRoll {
...
+            }, 0);
+        }, 0);
+    }
+
+    /**
+     * Grabs all instances of ActiveEffect.change values from the toggleable bonusEffects of the roll 
+     * that match a given change.key partial path.
+     * @param {string} path The full or partial effect.change key 
+     * @param {string} label The label to give the modifiers
+     * @returns {[label: string, value: any]}
```
CarlosFdez:
```suggestion
     * @returns {[{ label: string, value: any} ]}
```

CarlosFdez:
ah, I see. Maybe should be getBonuses() and getBonusValue() but this is fine.

CarlosFdez (approved):

PR #2111, opened 2026-07-24
[Feature] Targeting Rework

WBHarry:
Added in a D&D style `Targets | Selected` split in targeting. Players do not see any tabs.

**Targets Tab**: Tokens the user of the ability had `targeted` before using the ability.
**Selected Tab**: Tokens that are currently selected on the canvas.

The GM has access to the two tabs. If they want to target something different than the initial targeting for any reason, they can just swap to the `Selected` tab and select the tokens they want.

**Neat Things**
If nothing is targeted initially, the `Selected` tab will be selected right away.
In a players case, they are also put into select mode. This means that Player A can use a healing potion to heal Player B -> Player B has their own token selected and uses the heal button.

I removed `self` targeting on all typical potions consumables in the SRD to work according to the flow above.

### DIFF
--- a/module/data/chat-message/actorRoll.mjs
+++ b/module/data/chat-message/actorRoll.mjs
@@ -129,17 +129,22 @@ export default class DHActorRoll extends foundry.abstract.TypeDataModel {
      */
     _getCurrentTargets() {
         const getCommonData = data => {
+            const actor = data.actorId ? foundry.utils.fromUuidSync(data.actorId) : null;
             const toHitNumber = data.difficulty || data.evasion;
             const hitSuccessfull = (toHitNumber === null || !this.roll) ? false : this.roll.total >= toHitNumber;
 
             const saveValue = this.targetSaves[data.id];
             const saveSuccessfull = saveValue === undefined ? false : 
                 saveValue >= (this.action.save.difficulty ?? this.action.actor?.baseSaveDifficulty);
-            
+            const hasResistData = this.hasDamage && this.damage.active && actor;
+            const resistData = hasResistData ? actor.getResistanceStatus(this.damage.main.options.damageTypes) : null;
+
             return {
                 ...data,
                 hitResult: this.hasRoll ? { success: hitSuccessfull } : null,
-                saveResult: saveValue ? { success: saveSuccessfull } : null
+                saveResult: saveValue ? { success: saveSuccessfull } : null,
+                resistant: Boolean(resistData?.resistant),
+                immune: Boolean(resistData?.immune)
             }
         };
 
@@ -198,6 +203,8 @@ export default class DHActorRoll extends foundry.abstract.TypeDataModel {
                 selectedTargetsData: this._getSelectedTargetsData(),
                 hasSave: this.hasSave,
                 hasRoll: this.hasRoll,
+                hasDamage: this.hasDamage,
+                damage: this.damage,
                 isGM: game.user.isGM
             }
         );
--- a/module/data/chat-message/actorRoll.mjs
+++ b/module/data/chat-message/actorRoll.mjs
@@ -3,23 +3,6 @@ import { ChatDamageData } from './chatDamageData.mjs';
 
 const fields = foundry.data.fields;
 
-const targetsField = () =>
-    new fields.ArrayField(
-        new fields.SchemaField({
-            id: new fields.StringField({}),
-            actorId: new fields.StringField({}),
-            name: new fields.StringField({}),
-            img: new fields.StringField({}),
-            difficulty: new fields.NumberField({ integer: true, nullable: true }),
-            evasion: new fields.NumberField({ integer: true }),
-            hit: new fields.BooleanField({ initial: false }),
-            saved: new fields.SchemaField({
-                result: new fields.NumberField(),
-                success: new fields.BooleanField({ nullable: true, initial: null })
-            })
-        })
-    );
-
 export const originItemField = () =>
     new fields.SchemaField({
         type: new fields.StringField({
@@ -31,11 +14,16 @@ export const originItemField = () =>
     });
 
 export default class DHActorRoll extends foundry.abstract.TypeDataModel {
+    constructor(data, options) {
+        super(data, options);
+
+        this.targeting = { usingSelect: !this.targets.length };
+    }
+
     static defineSchema() {
         return {
             title: new fields.StringField(),
             actionDescription: new fields.HTMLField(),
-            targets: targetsField(),
             hasRoll: new fields.BooleanField({ initial: false }),
             hasDamage: new fields.BooleanField({ initial: false }),
             hasHealing: new fields.BooleanField({ initial: false }),
@@ -45,6 +33,17 @@ export default class DHActorRoll extends foundry.abstract.TypeDataModel {
             reloadCheckValue: new fields.NumberField({ integer: true, nullable: true, initial: null }),
             isDirect: new fields.BooleanField({ initial: false }),
             onSave: new fields.StringField(),
+            targets: new fields.ArrayField(
+                new fields.SchemaField({
+                    id: new fields.StringField({}),
+                    actorId: new fields.StringField({}),
+                    name: new fields.StringField({}),
+                    img: new fields.StringField({}),
+                    difficulty: new fields.NumberField({ integer: true, nullable: true }),
+                    evasion: new fields.NumberField({ integer: true })
+                })
+            ),
+            targetSaves: new fields.TypedObjectField(new fields.NumberField({ integer: true })),
             source: new fields.SchemaField({
                 actor: new fields.StringField(),
                 item: new fields.StringField(),
@@ -111,38 +110,36 @@ export default class DHActorRoll extends foundry.abstract.TypeDataModel {
         return null;
     }
 
-    get targetMode() {
-        return this.parent.targetSelection;
-    }
+    get currentTargets() {
+        const getCommonData = data => {
+            const toHitNumber = data.difficulty || data.evasion;
+            const hitSuccessfull = (toHitNumber === null || !this.roll) ? false : this.roll.total >= toHitNumber;
 
-    set targetMode(mode) {
-        if (!this.parent.isAuthor) return;
-        this.parent.targetSelection = mode;
-        this.registerTargetHook();
-        this.updateTargets();
-    }
-
-    get hitTargets() {
-        return this.currentTargets.filter(t => t.hit || !this.hasRoll || !this.targetMode);
-    }
-
-    async updateTargets() {
-        if (!ui.chat.collection.get(this.parent.id)) return;
-        let targets;
-        if (this.targetMode) targets = this.targets;
-        else
-            targets = Array.from(game.user.targets).map(t =>
-                game.system.api.fields.ActionFields.TargetField.formatTarget(t)
-            );
-
-        await this.parent.update({
-            flags: {
-                [game.system.id]: {
-                    targets: targets,
-                    targetMode: this.targetMode
-                }
+            const saveValue = this.targetSaves[data.id];
+            const saveSuccessfull = saveValue === undefined ? false : 
+                saveValue >= (this.action.save.difficulty ?? this.action.actor?.baseSaveDifficulty);
+            
+            return {
+                ...data,
+                hitResult: this.hasRoll ? { success: hitSuccessfull } : null,
+                saveResult: saveValue ? { success: saveSuccessfull } : null
             }
-        });
+        };
+
+        if (!this.targeting.usingSelect) return this.targets.map(getCommonData);
+
+        return (canvas.tokens?.controlled ?? []).map(token => ({
+            id: token.id,
+            actorId: token.document.actor?.uuid,
+            name: token.document.prototype?.name ?? token.document.name,
+            img: token.document.texture.src,
+            difficulty: token.document.actor?.system.difficulty,
+            evasion: token.document.actor?.system.evasion
+        })).map(getCommonData);
+    }
+
+    get hasUnfinishedSaves() {
+        return this.hasSaves && this.currentTargets.some(x => !x.saveResult);
     }
 
     async getRerolledDamage() {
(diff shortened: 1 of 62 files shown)


module/data/chat-message/actorRoll.mjs
```diff
@@ -111,38 +110,63 @@ export default class DHActorRoll extends foundry.abstract.TypeDataModel {
...
-                [game.system.id]: {
-                    targets: targets,
-                    targetMode: this.targetMode
-                }
+    syncSelectedTokens = foundry.utils.debounce(async () => {
+        if (!this.targeting.usingSelect) return;
+        
+        this.updateTargetHTML();
+    }, 50);
+
```
CarlosFdez:
```suggestion
        if (this.targeting.usingSelect) this.updateTargetHTML();
    }, 50);

```

module/data/fields/action/effectsField.mjs
```diff
@@ -48,7 +48,9 @@ export default class EffectsField extends fields.ArrayField {
         let effects = this.effects;
         const messageTargets = [];
         targets.forEach(async baseToken => {
```
CarlosFdez: You should convert this to a for loop. All of the awaits in the function are not actually awaited like they would be in a for loop. If you want each loop entry to be resolve asynchronously with each other, convert the forEach() to a map() fed into Promise.allSettled().

Same with the forEach() that's nested inside.

module/data/chat-message/actorRoll.mjs
```diff
@@ -111,38 +110,89 @@ export default class DHActorRoll extends foundry.abstract.TypeDataModel {
...
-            );
-
-        await this.parent.update({
-            flags: {
-                [game.system.id]: {
-                    targets: targets,
-                    targetMode: this.targetMode
-                }
+    get hasUnfinishedSaves() {
+        return this.hasSaves && this.currentHitTargets.some(x => !x.saveResult);
```
Skarlso: `hasSaves` this does not appear to exists 🤔

module/data/fields/action/effectsField.mjs
```diff
@@ -47,8 +47,10 @@ export default class EffectsField extends fields.ArrayField {
         const conditions = CONFIG.DH.GENERAL.conditions();
         let effects = this.effects;
         const messageTargets = [];
-        targets.forEach(async baseToken => {
-            if (this.hasSave && baseToken.saved.success === true) effects = this.effects.filter(e => e.onSave === true);
+        for (const baseToken of targets) {
+            if (this.hasSave && baseToken.saveResult.success === true)
```
Skarlso: I think this will crash if there are no saves, right? Don't you want `baseToken.saveResult?.success === true` here?

module/data/fields/action/effectsField.mjs
```diff
@@ -47,8 +47,10 @@ export default class EffectsField extends fields.ArrayField {
         const conditions = CONFIG.DH.GENERAL.conditions();
         let effects = this.effects;
         const messageTargets = [];
-        targets.forEach(async baseToken => {
-            if (this.hasSave && baseToken.saved.success === true) effects = this.effects.filter(e => e.onSave === true);
```
Skarlso, line 51: As far as I understand callbacks and the for... your return now returns the entire function. Previously, it only exited the forEach. Are you sure that's a good thing? :D

module/documents/chatMessage.mjs
```diff
@@ -358,12 +367,8 @@ export default class DhpChatMessage extends foundry.documents.ChatMessage {
         }
     }
 
-    filterPermTargets(targets) {
-        return targets.filter(t => fromUuidSync(t.actorId)?.canUserModify(game.user, 'update'));
-    }
-
     consumeOnSuccess() {
-        if (!this.system.successConsumed && !this.targetSelection) this.system.action?.consume(this.system, true);
+        if (!this.system.successConsumed && !this.system.targets) this.system.action?.consume(this.system, true);
```
Skarlso: I think this needs to be `!this.system.targets.length` right? Since targets will always be true since it's an array? And so this will always be true. I think. :D I'm not a js expert, but I know we shouldn't use them as booleans. :D

module/documents/chatMessage.mjs
```diff
@@ -68,6 +59,12 @@ export default class DhpChatMessage extends foundry.documents.ChatMessage {
         }
     }
 
+    _onDelete(options, userId) {
+        super._onDelete(options, userId);
+
+        Hooks.off('targetToken', this.onSelectToken);
```
Skarlso: Errr... isn't this the wrong event? @cptn-cosmo halp :D

templates/ui/chat/parts/target-tokens-part.hbs
```diff
@@ -0,0 +1,60 @@
...
+                {{/if}}
+            </div>
+        {{else}}
+            <i>{{localize "DAGGERHEART.GENERAL.noTarget"}}</i>
+        {{/each}}
+    {{else}}
+        <div class="selected-targets-container">
+            {{#with selectedTargetsData}}
+                {{#each tokens as |target|}}
+                    <div class="roll-target" data-token="{{target.token.id}}">
```
Skarlso: selectedTargetsData is currentargets, right? This appears to be a flat map I don't see how target.token.id exists in there? 🤔

templates/ui/chat/parts/target-part.hbs
```diff
@@ -13,43 +14,7 @@
...
-                                    {{#if hit}}
-                                        {{localize "DAGGERHEART.GENERAL.hit.single"}}
-                                    {{else}}
-                                        {{localize "DAGGERHEART.GENERAL.miss.single"}}
-                                    {{/if}}
-                                </div>
-                            {{/if}}
-                        </div>
-                        {{#if (and ../hasSave (or hit (not @root.hasRoll) (not @root.targetMode)))}}
-                            <div class="target-save{{#if saved.result includeZero=true}} is-rolled{{/if}}" data-perm-id="{{actorId}}">
```
Skarlso, line 44: There is UX loss without using `includeZero=true` since now the 0 is truhy in your checks down below.

module/data/chat-message/actorRoll.mjs
```diff
@@ -111,38 +110,89 @@ export default class DHActorRoll extends foundry.abstract.TypeDataModel {
...
-        });
+        );
+        const chatMessageHTML = ui.chat.element.querySelector(`.chat-message[data-message-id="${this.parent.id}"]`);
+        const element = chatMessageHTML.querySelector('.chat-roll .target-section .roll-part-content .wrapper');
+        element.outerHTML = targetTokensHTML;
+        this.parent.addTargetSectionListeners(chatMessageHTML);
+    }
+    
+    setSelectedAsTargets() {
+        this.parent.update({ 'system.targets': this.currentTargets });
```
Skarlso: Hm, this is not being a nice citizen since it's passing in the entire thing even fields that aren't needed right? Where as previously, it was rather clean.

Skarlso (commented):
This is my first review here, but I was trying to be helpful. :) Hopefully these make sense, as I'm still trying to learn the codebase. Forgive me if something is off.

Could you also please include some testing and screenshots? 🙇 ❤️ ✨  Thanks!

module/data/chat-message/actorRoll.mjs
CarlosFdez: Foundry strips out unused fields anyways, but this is being fed a getter, which already only grabs certain fields. What's superfluous?

WBHarry: Technically it is being fed 4 properties that are not part of the `target` datamodel, since the getter calculates these for rendering purposes:
```js
hitResult,
saveResult
resistant,
immune
```

We could strip them out, but they are indeed dropped by Foundry. Only what's in the actual datamodel gets saved to the DB.

Skarlso: Correct. Foundry strips it out. Previously, Foundry didn't have to do anything. :) It adds extra load.

CarlosFdez: I wouldn't worry about that sort of micro optimization, especially for something that isn't a hot loop.

I have some architectural concerns with the getter in the first place that I'll think about later, but more props in the diff is nothing to worry about.

Skarlso: Understood 👍

CarlosFdez: That said, the method itself looks simple enough that we maybe shouldn't make it part of the API (the public methods). This is only used in a single location anyways.

module/data/fields/action/effectsField.mjs
```diff
@@ -47,8 +47,10 @@ export default class EffectsField extends fields.ArrayField {
...
+        for (const baseToken of targets) {
+            if (this.hasSave && baseToken.saveResult.success === true) 
+                effects = this.effects.filter(e => e.onSave === true);
+            
             if (!effects.length) return;
```
CarlosFdez:
```suggestion
            if (!effects.length) continue;
```

At least, I think this is the intent. If it should be break, do that instead.

module/data/fields/action/effectsField.mjs
```diff
@@ -69,12 +71,12 @@ export default class EffectsField extends fields.ArrayField {
                     : null
             });
 
-            effects.forEach(async e => {
+            for (const e of effects) {
                 const effect = (this.item.applyEffects ?? this.item.effects).get(e._id);
                 if (!token.actor || !effect) return;
                 await EffectsField.applyEffect(effect, token.actor);
```
CarlosFdez:
```suggestion
                if (token.actor && effect) {
                    await EffectsField.applyEffect(effect, token.actor);
                }
```

This avoids the return statement breaking the entire function as well.

module/data/chat-message/actorRoll.mjs
WBHarry: Correct! Changed it to `hasSave` 👍

module/data/fields/action/effectsField.mjs
WBHarry: Hm, yeah I -think- that could end up happening here if only some of the saves have been rolled. Changed :+1:

module/data/fields/action/effectsField.mjs
WBHarry: Yeah, it should be `continue`. Changed it :+1:

module/documents/chatMessage.mjs
WBHarry: Yup. Originally it was working of `targetToken`, but now it's changed to `controlToken`. Fixed it :+1:

module/data/chat-message/actorRoll.mjs
CarlosFdez: As in, the method could be removed, and the one line that it does moved to where the function would normally be invoked.

templates/ui/chat/parts/target-part.hbs
WBHarry: Previously it checked `saved.result` which was a number, the rolled value of the save.
Now it checks wether `savedResult` is null or not, which works :+1:

module/data/fields/action/effectsField.mjs
WBHarry: Fixed :+1:

module/data/fields/action/effectsField.mjs
WBHarry: Changed to this :+1:

module/documents/chatMessage.mjs
WBHarry: Actually, I can't find a case where `consumeOnSuccess` is ever ran.

I removed it. If some case I can't figure out right now does end up showing up, then we'll have to bring it back at that point and write a comment about it at that point.

templates/ui/chat/parts/target-tokens-part.hbs
WBHarry: Hm, yeah it was wrong. Changed to `target.id` :+1:

module/data/chat-message/actorRoll.mjs
WBHarry: I went ahead and removed it. I figured it was the concern of the system type, but it was pretty silly. :+1:

module/documents/chatMessage.mjs
CarlosFdez: Isn't consume on success that thing where you set up a cost and it only consumes a hope or stress or what have you on success? You first have to set up a roll, then go back to the previous tab and you can setup consume on success.

WBHarry: Yeah, we have that functionality. But even if I set that up, that works without this function.

WBHarry: I've got to check the exactities of the old `targetSelection`.
I guess it could have an effect if you never had an initial target

WBHarry: Restored it and fixed an existing error that made it not work.

Skarlso: Nice, thanks so much for the investigation! :) 🙇

module/documents/chatMessage.mjs
Skarlso: Ah noice. I wasn't sure here. :D

module/data/chat-message/actorRoll.mjs
Skarlso: Thank you so much again for the investigation. I understand it a bit better now. :)

module/documents/chatMessage.mjs
```diff
@@ -1,15 +1,28 @@
 import { emitGMUpdate, emitGMCreate, GMUpdateEvent } from '../systemRegistration/socket.mjs';
 
 export default class DhpChatMessage extends foundry.documents.ChatMessage {
-    targetHook = null;
-
     static #EXPAND_SECTIONS = [
         { selector: '.roll-section [data-action="expandRoll"]', key: 'roll' },
         { selector: '.damage-section', key: 'damage' },
-        { selector: '.target-section', key: 'target' },
```
CarlosFdez, line 9: I'd imagine this removal would break expand state restoration on re-render. The expandRoll is on the roll-section, not in a child element.

Ideally we'd have expandRoll on the header only, but expandRoll is a core foundry action in the chat log and its not that easy to change that behavior.

EDIT: Given the new default expanded state and the removal of the setting though, perhaps we can remove the collapse behavior for targets altogether? That'll make the click behavior less finicky.

module/documents/chatMessage.mjs
```diff
@@ -387,9 +394,20 @@ export default class DhpChatMessage extends foundry.documents.ChatMessage {
         game.canvas.pan(token);
     }
 
-    onTargetSelection(event) {
+    /** Handle the user clicking the button to convert selected to targets and update the chat message */
+    async #onSelectedToTargets(event) {
+        event.stopPropagation();
+        // Update the targets. Regardless of what happens, swap to the targets tab
```
```suggestion
        // Update the targets and ensure that we swap to the targets tab
```

CarlosFdez (approved):

PR #2114, opened 2026-07-26
[Fix] scroll restoration for several item sheets

CarlosFdez:
A few are missing but I knocked out a few simple ones.


CarlosFdez:
The snuck in PR is because it seems to fix some overlap issues that happen inconsistently, perhaps only in development mode. Its good to have, but can't really be considered part of the patch notes. It was something I caught while doing the rest of this, which was motivated from trying to edit domain cards on my own character sheet.

WBHarry (approved):
Good catch :+1:

PR #2115, opened 2026-07-26
Reduce top center fear tracker shifting on lower resolutions

CarlosFdez:
Same as before, but now no longer shifts position based on selecting tokens with effects or not, it only shifts based on countdown size and sidebar status.


WBHarry (approved):
:+1:

PR #2116, opened 2026-07-26
[Fix] visibility of equipped weapons with transparency in party sheet

CarlosFdez:
The top left is an svg with transparency

image


WBHarry (approved):
:+1:

PR #2117, opened 2026-07-27
Refactor scoped color definitions, add immunity, fix error

CarlosFdez:
I originally split them because we were gonna need to split for some things but now I'm thinking it'll be easier to maintain if we at least merge what we can.


CarlosFdez:
I had drafted cause there was a error with light mode but it looks like its not new.

moliloo (approved):

PR #2107, opened 2026-07-23
[Feature] Support tooltips for loadout card view

CarlosFdez:
Cleans up loadout card view and adds tooltips. Also fixes some css regression in v14 due to font awesome changes. Tooltips only show up when mousing over the name.

This was mostly a request @cptn-cosmo so I'm going to tag them as a reviewer.

PR #2119, opened 2026-07-28
Improve visibility in light mode, fix theme mix errors, and loadout card tweaks

CarlosFdez:
First, a few new descriptive css variables. Yay! While button color css already exists in foundry, its another one of those absurdly scoped ones, forcing us to redefine it.

Certain elements in light mode are more visible, such as resources. While this styling may not be ideal, tweaking it otherwise has bad visibility. It may be worth giving it another look once resources have been further tweaked:

image

This also fixes issues that would occur if foundry was set to light mode but a specific sheet was set to dark mode. Certain elements with specific overrides using the appTheme mixin would continue to use light mode styling even on a dark mode sheet. This caused issues with things like the unstoppable die.

Finally, it carries over some of the non-controversial elements of https://github.com/Foundryborne/daggerheart/pull/2107. Specifically changes in the appearance of the recall cost, handling of multilines, and the image aspect ratio. No tooltip though.

### DIFF
--- a/styles/less/global/elements.less
+++ b/styles/less/global/elements.less
@@ -99,9 +99,10 @@
     }
 
     button:where(:not(.plain, color-picker *, file-picker *)) {
-        background: light-dark(transparent, @golden);
-        border: 1px solid light-dark(@dark-blue, @dark-blue);
-        color: light-dark(@dark-blue, @dark-blue);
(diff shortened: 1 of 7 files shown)


WBHarry (approved):
Beautiful `@button-color` :+1:

PR #2120, opened 2026-07-28
[Fix] ContextMenu Damage Option

WBHarry:
The Damage option in a weapon's context menu wasn't changed with the damage rework. Fixed so that it's available again.
image


CarlosFdez (approved):
Looks good to me!

PR #2121, opened 2026-07-28
[Fix] a few minor fixes here and there and linter update

Skarlso:
## Description

This PR fixes a couple of issues I found while I was going over the codebase. :) I tried to also update the linter settings. :)

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Code cleanup/refactor
- [ ] Documentation update
- [ ] Test coverage
- [ ] Dependency update
- [ ] Configuration change

## How Has This Been Tested?

- [ ] Manual testing
- [ ] Other:

## Checklist

- [ ] My code follows the project style guidelines
- [ ] I have performed a self-review of my code
- [ ] My changes generate no new warnings or errors

## Additional Comments

---

### DIFF
--- a/module/applications/sheets-configs/action-base-config.mjs
+++ b/module/applications/sheets-configs/action-base-config.mjs
@@ -498,11 +498,9 @@ export default class DHActionBaseConfig extends DaggerheartSheet(ApplicationV2) 
 
     async onSummonDrop(data) {
         const actionData = this.action.toObject();
-        let countvalue = 1;
         for (const entry of actionData.summon) {
             if (entry.actorUUID === data.uuid) {
                 entry.count += 1;
-                countvalue = entry.count;
                 await this.constructor.updateForm.bind(this)(null, null, {
                     object: foundry.utils.flattenObject(actionData)
                 });
@@ -510,7 +508,7 @@ export default class DHActionBaseConfig extends DaggerheartSheet(ApplicationV2) 
             }
         }
 
-        actionData.summon.push({ actorUUID: data.uuid, count: countvalue });
+        actionData.summon.push({ actorUUID: data.uuid, count: 1 });
         await this.constructor.updateForm.bind(this)(null, null, { object: foundry.utils.flattenObject(actionData) });
     }
 
--- a/module/applications/sheets-configs/action-base-config.mjs
+++ b/module/applications/sheets-configs/action-base-config.mjs
@@ -498,14 +498,12 @@ export default class DHActionBaseConfig extends DaggerheartSheet(ApplicationV2) 
 
     async onSummonDrop(data) {
         const actionData = this.action.toObject();
-        for (const entry of actionData.summon) {
-            if (entry.actorUUID === data.uuid) {
-                entry.count += 1;
-                await this.constructor.updateForm.bind(this)(null, null, {
-                    object: foundry.utils.flattenObject(actionData)
-                });
-                return;
-            }
+        const existingSummon = actionData.summon.find(x => x.actorUUID === data.uuid);
+        if (existingSummon) {
+            existingSummon.count++;
+            return await this.constructor.updateForm.bind(this)(null, null, {
+                object: foundry.utils.flattenObject(actionData)
+            });
         }
 
         actionData.summon.push({ actorUUID: data.uuid, count: 1 });
--- a/module/applications/sheets-configs/action-config.mjs
+++ b/module/applications/sheets-configs/action-config.mjs
@@ -13,7 +13,7 @@ export default class DHActionConfig extends DHActionBaseConfig {
 
     async _prepareContext(options) {
         const context = await super._prepareContext(options);
-        if (!!this.action.effects) context.effects = this.action.effects.map(e => this.action.item.effects.get(e._id));
+        if (this.action.effects) context.effects = this.action.effects.map(e => this.action.item.effects.get(e._id));
         context.getEffectDetails = this.getEffectDetails.bind(this);
 
         return context;
--- a/module/applications/sheets-configs/action-config.mjs
+++ b/module/applications/sheets-configs/action-config.mjs
@@ -56,7 +56,7 @@ export default class DHActionConfig extends DHActionBaseConfig {
         if (!this.action.effects) return;
 
         const { areaIndex, index } = button.dataset;
-        let effectId = null;
+        let effectId;
         if (areaIndex !== undefined) {
             effectId = this.action.areas[areaIndex].effects[index];
             const data = this.action.toObject();
--- a/module/applications/sheets/actors/adversary.mjs
+++ b/module/applications/sheets/actors/adversary.mjs
@@ -8,7 +8,6 @@ export default class AdversarySheet extends DHBaseActorSheet {
     static DEFAULT_OPTIONS = {
         classes: ['adversary'],
         position: { width: 645, height: 750 },
-        window: { resizable: true },
         actions: {
             toggleHitPoints: AdversarySheet.#toggleHitPoints,
             toggleStress: AdversarySheet.#toggleStress,
--- a/module/applications/sheets/actors/adversary.mjs
+++ b/module/applications/sheets/actors/adversary.mjs
@@ -8,13 +8,6 @@ export default class AdversarySheet extends DHBaseActorSheet {
     static DEFAULT_OPTIONS = {
         classes: ['adversary'],
         position: { width: 645, height: 750 },
-        actions: {
-            toggleHitPoints: AdversarySheet.#toggleHitPoints,
-            toggleStress: AdversarySheet.#toggleStress,
-            reactionRoll: AdversarySheet.#reactionRoll,
-            toggleResourceDice: AdversarySheet.#toggleResourceDice,
-            handleResourceDice: AdversarySheet.#handleResourceDice
-        },
         window: {
             resizable: true,
             controls: [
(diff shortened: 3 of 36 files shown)


module/applications/sheets-configs/action-config.mjs
```diff
@@ -13,7 +13,7 @@ export default class DHActionConfig extends DHActionBaseConfig {
 
     async _prepareContext(options) {
         const context = await super._prepareContext(options);
-        if (!!this.action.effects) context.effects = this.action.effects.map(e => this.action.item.effects.get(e._id));
```
Skarlso, line 16: I'm _fairly_ sure this negation here is incorrect. 🤔

Skarlso: This is the result of `no-extra-boolean-cast`. Will have to investigate it. :D

Skarlso:
Some of these are linter added. I will test them throughout. :)

module/documents/item.mjs
```diff
@@ -252,6 +246,8 @@ export default class DHItem extends foundry.documents.Item {
 
     /** @inheritDoc */
     static migrateData(source) {
+        if (source.system?.attack && !source.system.attack.type) source.system.attack.type = 'attack';
```
CarlosFdez:
```suggestion
        if (source.system?.attack) source.system.attack.type ??= 'attack';
```

Would this be equivalent?

CarlosFdez: That said poking @WBHarry. It looks like this was from a previous migrateData() that got overshadowed, and thus it was never running. Is this code still relevant?

module/documents/actor.mjs
CarlosFdez: Moving these to where you deleted the other ones might be good. Title is "higher up" wrt priority.

module/applications/sheets-configs/action-config.mjs
CarlosFdez: This change looks equivalent. !!x is equivalent to Boolean(x). You don't have to coerce something into a boolean for an if statement so that's probably why the linter is upset?

module/applications/sheets/actors/adversary.mjs
```diff
@@ -8,7 +8,6 @@ export default class AdversarySheet extends DHBaseActorSheet {
     static DEFAULT_OPTIONS = {
         classes: ['adversary'],
         position: { width: 645, height: 750 },
-        window: { resizable: true },
```
CarlosFdez, line 11: window should probably exist above actions for convention reasons but otherwise yeah this is correct.

module/applications/sheets-configs/action-config.mjs
Skarlso: I think so, yes. 🤔

module/documents/item.mjs
Skarlso: With my limited knowledge I think this is not equivalent. 🤔 ! is a negation and ?? only works for null fields, right? And as far as I can see from the code that wouldn't result in the same logic here.

CarlosFdez: There is a second !. Double negation.

Skarlso:
Found a couple more invalid assignments. Adding them..

module/documents/item.mjs
Skarlso: A second `!`... where? Am I blind? :D

CarlosFdez: I was on the phone, I mixed it up for the other one where the !!this.action.effects was removed. Disregard, there is no !!.

??= works for null and undefined. While the original expression also override on false or empty string, attack.type cannot be `false` or empty string regardless. The intent for this expression was to set a default value if it was missing.

However this is something the schema is intended to handle, and it wasn't even executing anyways (you merged the two functions, but the first fn was originally ignored). So it might just be removable outright. Requires @WBHarry since he has more context here.

module/applications/sheets-configs/action-base-config.mjs
```diff
@@ -498,19 +498,17 @@ export default class DHActionBaseConfig extends DaggerheartSheet(ApplicationV2)
 
     async onSummonDrop(data) {
         const actionData = this.action.toObject();
-        let countvalue = 1;
         for (const entry of actionData.summon) {
             if (entry.actorUUID === data.uuid) {
                 entry.count += 1;
-                countvalue = entry.count;
```
CarlosFdez, line 505: Needs @WBHarry eyes. It looks like you removed it because there is a return here anyways, so this count is completely pointless. But now I'm wondering if there was an issue with the original logic in the first place.

module/documents/actor.mjs
```diff
@@ -879,28 +875,26 @@ export default class DhpActor extends Actor {
             }
         });
 
-        Object.keys(updates).forEach(async key => {
+        for (const key of Object.keys(updates)) {
             const u = updates[key];
             if (key === 'items') {
-                Object.values(u).forEach(async item => {
+                for (const item of Object.values(u)) {
```
CarlosFdez: As someone really happy to see the forEach loops go, this is a logic change. Before the awaits were not actually doing anything, since the function's callback was a new scope and the forEach itself wasn't awaited in turn. But now it'll await for each iteration. Removing await will return the old behavior, as well using map() into Promise.all().

Perhaps this change in behavior is actually a bugfix though, but without testing I can't be sure.

module/documents/item.mjs
WBHarry: Hm..yeah I have to imagine it's fine to just remove it since it hasn't been causing any issues for however long it's been non-func like this.

module/documents/actor.mjs
Skarlso: I'll test this :)

Skarlso:
Okay, I'm going to run a full test in here and play some things through to verify all the changes and paths I touched. Especially the loops and the promise changes. I'm not sure about those I basically read the doc and saw how to fix that warning. :D

module/documents/item.mjs
Skarlso: Removed. :)

module/data/fields/action/saveField.mjs
```diff
@@ -58,31 +58,25 @@ export default class SaveField extends fields.SchemaField {
...
+                        : actor.owner.query('reactionRoll', {
+                            actionId: this.uuid,
+                            actorId: actor.uuid,
+                            event,
+                            message
+                        });
+                const result = await rollSave;
+                await SaveField.updateSaveMessage.call(this, result, message, target.id);
+            })
+        );
```
CarlosFdez: Not an issue but at least for this one, the logic appears to be the same. Nice cleanup.

module/applications/sheets-configs/action-base-config.mjs
WBHarry: Yeah, that's some nonsense. I'd replace the whole function as so:
```
async onSummonDrop(data) {
        const actionData = this.action.toObject();
        const existingSummon = actionData.summon.find(x => x.actorUUID === data.uuid);
        if (existingSummon) {
            existingSummon.count++;
            return await this.constructor.updateForm.bind(this)(null, null, {
                object: foundry.utils.flattenObject(actionData)
            });
        }

        actionData.summon.push({ actorUUID: data.uuid, count: 1 });
        await this.constructor.updateForm.bind(this)(null, null, { object: foundry.utils.flattenObject(actionData) });
    }
```

module/data/fields/action/saveField.mjs
Skarlso: Thanks! :) I just followed the docs. :D I still need to properly learn how const result = await rollSave works, and how forEach ( async ) changes... Like, of course I understand that changing things to for here and there will make it sequential instead. But that might be okay. And here, I had to read so much to even begin to understand why the linter is complaining. :D

CarlosFdez: in summary for the async/await, this is an oversimplification so things may not be exact, but here goes. You may need to google more to round out your understanding.

Basically any function that is an async function returns a promise. Anything that is a promises's "then" (as in .then()) gets added to a global event queue in the same way a `window.setTimeout()` can be used to defer execution. When you await inside an async function, what does is make the rest of the function the "then" of the promise being returned. Every single await causes a new "then" to be registered. In the past, before async/await, this used to be done via callback hell. But async await create a linear view.

In the case of a forEach(), the await will work inside each invocation, but NOT globally. The parent function isn't awaiting it, so they all just get added to the event queue all at once.

Skarlso: Ah thanks! That's a good explanation. I'll write this down! :) 🙇

module/documents/actor.mjs
WBHarry: Looking at it, I'm not a 100% sure what event exactly `game.socket.emit` waits for. Assumably it's just that the message has been -sent-, which would make an await harmless.

We could have it as sync (clearly, since it's been up till now 🙃). `updates.items` could theoretically cause an issue in `sync`, since right now all resource updates to items are set on unique keys ``${r.itemId}-${r.key}``. I don't think -is- ever a case, but ideally we should swap to storing all updates to `updates.items` under just the `r.itemId` keys so every update to a single item is gauranteed to be sent in one bunch of data to the GM client.

WBHarry: Tested it with a funny source of damage, and it doesn't seem to be an issue to be async :+1:

It did reveal what I mentioned about Item updates though, as you can see that the 2 armor slot markings are not joined together. But that's a separate issue.

Image

Skarlso: @WBHarry Would you like me to attempt to fix that in this PR? 🤔

CarlosFdez:
Would you be able to merge/rebase soon? I'd like to get this in before any new features if possible. If not I'll look into getting this past the finish line.

module/applications/sheets-configs/action-base-config.mjs
Skarlso: Got it. Thank you!

Skarlso: Oh wait, did I misunderstood this? I replaced this with what you pasted in.. :D

module/documents/actor.mjs
Skarlso: Got an answer for this from Harry that it should be a different one. :)

Skarlso:
@CarlosFdez Should be good to go now. 🤞 ❤️

CarlosFdez (approved):
Looks good to me! Thanks for the PR.

PR #2122, opened 2026-07-29
[Fix] ChatMessage Migration

WBHarry:
Tested migration of an old world. The old way of storing damage rolls was so brokenly bad that it's a real bother to try and stitch it together into a proper roll.

We -could- do it, but I'm not sure it's really worthwhile. Any existing damge messages in chat should sensibly not be relevant next session that's started up. With that in mind I simply set old damage messages to have their rolls evaluated as they were always `evaluated = false` in the old messages, and that causes an error that makes the chat message invalid.


CarlosFdez:
This works for me, though I haven't tried it.

CarlosFdez (approved):

WBHarry:
Essentially, old messages with damage end up non-func, but they do render so that they can be deleted.

PR #2128, opened 2026-07-30
[Fix] ChatMessage Crits Not Considered Hits

WBHarry:
Fixed so that criticals are considered hits in chat message target section


Ikraik (approved):
LGTM 👍 Tested on lower values as well.

PR #2136, opened 2026-08-01
[Fix] Roll Automation

### LINKED ISSUE #2137: [Bug] Automation Apply Effects Doesn't Work
I went over the others and got them working, but was having some trouble here at the moment.
I don't really see these functions are high prio, so I passed it over to an issue.

Image

WBHarry:
Fixes #2137 

Fixed so that Automation Roll behaviors work again.
image

### DIFF
--- a/module/helpers/utils.mjs
+++ b/module/helpers/utils.mjs
@@ -909,3 +909,32 @@ export async function getWorldActor(baseActor) {
 
     return baseActor;
 }
+
+
+/**
+ * Get target data augmented with hitResult, saveResult, resistance and immunity
+ * @param {BaseTargetData} target
+ * @param {object} data - Config or ChatMessage.system data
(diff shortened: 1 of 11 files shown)


module/helpers/utils.mjs
```diff
@@ -908,4 +908,33 @@ export async function getWorldActor(baseActor) {
...
+}
+
+
+/**
+ * Get target data augmented with hitResult, saveResult, resistance and immunity
+ * @param {BaseTargetData} target
+ * @param {object} data - Config or ChatMessage.system data
+ * @returns {TargetData}
+ */
+export function getTargetData(target, data) {
```
WBHarry: I ended up moving this to utils while trying to fix auto apply effects.
Not strictly neccessary, but I figure it'll probably come in handy for fixing that.

PR #2138, opened 2026-08-01
[Feature] Ephemeral Effects

WBHarry:
Not finished. Mainly for me to diff check on things.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -2045,8 +2045,11 @@
         },
         "EPHEMERAL": {
             "FIELDS": {
+                "type": { "label": "Ephemeral Applies On" },
                 "bonusValue": { "label": "Bonus Value" },
-                "effects": { "label": "Effects" }
+                "effectData": {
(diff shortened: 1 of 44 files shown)


WBHarry:
I have added a bunch of logic for letting Ephemeral Effects have ActiveEffects they can add.
I think I will probably strip that out though. After looking at all the cases, I don't think Daggerheart ever actually has an ability that does: 
"BEFORE you make a roll, mark 1 stress to add <condition> if you succeed"
It is always:
"If you succeed, you may mark a stress to make the target <condition>"

And that should be handled by just using another action on the feature.

CarlosFdez:
Don't you pay that resource ahead of time for stuff like know the tide?

EDIT: Wait, I see what you mean, the on success isn't retroactive.

WBHarry:
Talking specifically about having Ephemeral Effects apply ActiveEffects. That probably isn't a case, so that code will likely go bye-bye.

Know The Tide is a good customer. You pay X to get X bonus to roll 👼 
(Mind that I haven't actually implemented `Scalable` fully yet)

PR #2142, opened 2026-08-03
[Feature] Rework countdown inheritance

CarlosFdez:
This allows setting the value of "hidden" in the ownership selection and changes the default permission in the field to be hidden instead. Furthermore, players now only have INHERIT/OWNER/OBSERVER permissions. Players set to observer or owner can see hidden countdowns. The "Hide new countdowns" option is now persisted, and replaces default permission for all intents and purposes.

image

There are no migrations, and I didn't localize a label. I'll do so after this is approved.


WBHarry (dismissed):
Nice. :+1:

Omit is quite handy for these sort of partial `choices` on fields.

CarlosFdez:
Alrighty, this should be done now. The one wart remaining is the label of the null hidden, but otherwise it works and I don't have an idea for that atm.

WBHarry (approved):
:+1:

PR #2144, opened 2026-08-04
[Feature] Evolution

WBHarry:
Edit: I've resolved the points I had thoughts about. Nothing else on my mind to fix atm. 👨‍🍳


WBHarry:
`Usable` didn't quite do the trick, but I just changed so that the ActionButtons in particular are disabled if a feature is `EvolutionLocked`. Right now  you can still click on the icon or the `ToChat` icon, and in either case it'll just post the feature to chat, which I think is fine as a communication tool.

----
I added `//DEVMODE` markings everywhere where I added a conditional to avoid rendering or making functionality available so it can be easily removed by searching on that later.
Right now DevMode is set in the `system.json`, and there's a `deploy` script addition to delete that system flag on build. I don't think it's ideal, but I'm unsure of another way to make it work.

PR #2145, opened 2026-08-04
Countdown Rework - Markers

WBHarry:
There is one situation I think ends up akward for the GM with the current hidden handling.

If a player has hidden a countdown, that countdown will show the `eye-slash` icon to denote that it is hidden. But since players can be set to `Observer` and `Owner`, the countdown is in fact not hidden to everyone.
In the main PR I added a small quality of life where the tooltip tells you how many it is still visible to, but I still think GMs will be confused about which countdowns are -actually- hidden to everyone.

The solution I tried here is to add coloured dots with the user color for the users that can still see a hidden countdown.
image

**Hidden - No Player Access**
image

**Hidden - One Player Access**
image

**Hidden - 7 Players Access**
Completly unrealistic case. I designed the dots to fit 4x2. More than that and it currently doesn't layout well, but I really don't think that's an issue. You'd have to hide a countdown and explicitly give 9 players access anyway.
image

**Hover Shows Player Name**
image

### DIFF
--- a/module/applications/ui/countdowns.mjs
+++ b/module/applications/ui/countdowns.mjs
@@ -161,13 +161,23 @@ export default class DhCountdowns extends HandlebarsApplicationMixin(Application
                 !countdownEditable ||
                 (isLooping && (countdown.progress.current > 0 || countdown.progress.start === '0'));
 
-            const playersCountdownIsVisibleTo = game.users.filter(x => 
+            const usersCountdownIsVisibleTo = game.users.filter(x => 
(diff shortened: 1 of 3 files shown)


WBHarry:
A simpler solution we could try that perhaps also gets the message across would be to use the partial covered eye icon
image

If this has a tooltip stating the numbers of players that can see it, or perhaps really the names, then that might make it cleaner 🤔

CarlosFdez:
How does this handle the shrink mode?

I like either idea, go with whatever you personally like more. Unless it can't handle the collapsed view, then the eye-low-vision is perfect. Honestly, could even do both.

PR #2151, opened 2026-08-04
[Feature] Standard Attack ChangeType

WBHarry:
image

image


WBHarry:
I updated the Armor ChangeType section to the same look I did for StandardAttack.
The previous thought was to stay in line with the `standard-form` look, but I think this is a lot nicer.
**Before**
image
**After**
image

Ikraik (dismissed):
Image
Image

Tested few things. I feel the damage reference formula is wrong and is not linking properly

Image

It is reflected here as well. Possibly a condition needed here?

WBHarry:
Ah, yeah, the damage type isn't getting set correctly. I never opened the damage section to notice :+1:
image

WBHarry:
Pushed a fix.

Ikraik (approved):
All test cases worked correctly. 👍

CarlosFdez:
Since you've poked into how it works, I'll focus more on architecture.

Once we add the third change type, we should update the UI to be a Select + Add button. I have an idea but I'd rather wait till we have a bit more.

module/applications/sheets-configs/activeEffectConfig.mjs
```diff
@@ -203,7 +200,7 @@ export default class DhActiveEffectConfig extends foundry.applications.sheets.Ac
                 }));
                 break;
             case 'changes':
-                const singleTypes = ['armor'];
+                const singleTypes = ['armor', 'standardAttack'];
```
CarlosFdez: We should promote this to the change data itself (not the persisted one) as a property such as `single: true`. Perhaps as a property in `changeEffect` and/or `activeEffectModes`? Then we can fetch it from the actual data and also use it in UI once we allow dynamic addition of change types (future UI feature I'm planning)

I'm having trouble figuring out what to do with `changeEffect` and `activeEffectModes`. Looks like we have a bit of a split there. Any ideas?

module/applications/sheets-configs/activeEffectConfig.mjs
```diff
@@ -242,19 +239,20 @@ export default class DhActiveEffectConfig extends foundry.applications.sheets.Ac
...
+            this.addTypedChange(game.system.api.data.activeEffects.changeTypes[type]);
         } else {
-            this.removeTypedChange(event.target.dataset.index);
+            this.removeTypedChange(index);
         }
     }
 
     /* Could be generalised if needed later */
-    addArmorChange() {
+    addTypedChange(changeType) {
```
CarlosFdez: Perhaps make changeType a string? That might make it more intuitive to understand. Otherwise rename to changeData and add `@param {object} changeData`

module/data/item/beastform.mjs
```diff
@@ -99,27 +117,25 @@ export default class DHBeastform extends BaseDataItem {
...
     get beastformAttackData() {
-        const effect = this.parent.effects.find(x => x.type === 'beastform');
-        return DHBeastform.getBeastformAttackData(effect);
+        const effect = this.parent?.effects.find(x => x.type === 'beastform');
+        
+        return DHBeastform.getBeastformAttackData(effect, this.parent);
     }
 
-    static getBeastformAttackData(effect) {
+    static getBeastformAttackData(effect, actor) {
```
CarlosFdez: Is there a way to make this a non static method on effect?

Also it looks like actor is possible to derive from effect.parent, though one usage does effect.parent.getRollData() instead. Would it be possible to simplify the params to derive actor from effect? If its that its uncertain if effect.parent is an actor, there is a static private helper in DhActiveEffect called resolveParentDocument() that could maybe be promoted to a general importable helper function?

module/applications/sheets-configs/activeEffectConfig.mjs
WBHarry: I changed so that our custom change types set an immutable property `single`. By default it becomes true for them, but it can be set to false if we want that for some future type.
This property is now used for this filter check.

I did some cleanup regarding changeEffect and activeEffectModes. We no longer keep a list containing all the default changeTypes. We just have a list of our custom ones. The logic has been changed to check on foundry's listing and this short one instead. I think it's better 🙃

module/applications/sheets-configs/activeEffectConfig.mjs
WBHarry: I renamed the function to `addCustomChange`, unsure if it's better, but the function is only for our own changeTypes.
The input is now a string, and if it's not a proper key the function just returns.

module/data/item/beastform.mjs
WBHarry: I removed it as a static method. It's now instead a normal method on `ActiveEffect/Beastform.mjs`, which makes a lot more sense. This means it needs no parameters since the effect is `this` and actor is `this.parent.parent`.

I did have to keep the getter in `item/beastform.mjs` due to the way one of the tooltips gets it data, but I think that's quite clean still :+1:

CarlosFdez: I agree yeah, the getter was fine.

module/applications/sheets-configs/activeEffectConfig.mjs
CarlosFdez: Agreed. Significant improvement. Moving the effect based stuff from general was also good.

module/applications/sheets-configs/activeEffectConfig.mjs
CarlosFdez: Its better yeah, its a ui function. The thing I was concerned of was mostly a javascript where if you name something slightly wrong it causes confusion because types are not available.

module/applications/sheets-configs/activeEffectConfig.mjs
```diff
@@ -151,20 +152,16 @@ export default class DhActiveEffectConfig extends foundry.applications.sheets.Ac
...
-        htmlElement
-            .querySelector('.range-dependence-change-checkbox')
+        htmlElement.querySelector('.range-dependence-change-checkbox')
             ?.addEventListener('change', this.rangeDependenceChangeToggle.bind(this));
 
-        htmlElement
-            .querySelector('.armor-change-checkbox')
-            ?.addEventListener('change', this.armorChangeToggle.bind(this));
+        for (const element of htmlElement.querySelectorAll('.typed-change-checkbox'))
+            element.addEventListener('change', this.typedChangeToggle.bind(this));
```
CarlosFdez: Convention for these functions is private `#onX` functions. So this would be `this.#onToggleChangeType`. Search for `.bind(this)` in all files in the foundry source code. Changing the others to follow suit might be appropriate. 

This is a minor issue though, and I'll otherwise ignore it for the review, but I'd be happy at least.

CarlosFdez:
The swap to actor.system.attack from actor.system.usedUnarmed makes me quite happy.

module/applications/sheets-configs/activeEffectConfig.mjs
WBHarry: I tend to convince myself that "It's already wrong here, so I'll follow the standard here". It -is- better to go ahead and swap things though. Strangle method to bit by bit clean it up.

I changed it :+1:

module/data/activeEffect/changeTypes/baseCustomChangeType.mjs
```diff
@@ -0,0 +1,11 @@
+export default class BaseCustomChange extends foundry.abstract.DataModel {
+    constructor(data, { single = true, ...options } = {}) {
+        super(data, options);
+
+        Object.defineProperty(this, 'single', {
```
CarlosFdez: might be better to make single static on the subclass, and perhaps make a getter for this.constructor.single. That way single won't be incorrect on like a clone or something.

CarlosFdez: I'm not seeing anything else that would block a merge, though admittedly I didn't deep dive the migration that looked a bit intimidating.

WBHarry: I'd prefer not to need migrations like these, but it is what it is 😿 

I do feel confident about it though.

WBHarry: I swapped the single setup to:
```
#single = true;

get single() {
    return this.#single;
}
```

CarlosFdez: That's fine. I suppose if we ever get a non single type we can tweak that. If the stuff is in place we're good.

PR #2154, opened 2026-08-05
[Feature] Countdown Third Tab

WBHarry:
Added a third tab to the Countdowns UI `Misc`. It shows countdowns with type `Miscellanous`.
image
image

The idea is to allow a countdown type for projects/personal and "busywork", so they can easily be kept separate and only shown when wanted.


IrkTheImp (approved):
Looks good to me.

PR #2157, opened 2026-08-05
[Feature] Countdown Clearer Language

WBHarry:
I wanted to achieve 2 (3) things:
image
**1)** I think the `inherit` label should call back to whatever title we have for the `default permission` line to make the relationship clear. In this case I tried `Countdowns Status`, so the line became `Follow Status`.
**2)** The GM mainly interacts with `Hidden`/`Revealed`. I think that makes it a lot more clear to use those labels here aswell. `none='Hidden'`, `observer='Revealed'`.

**3)** Optional. I thought it looked a lot nicer to have the lines completly symmetrical instead of a void where the player lines have an image. Tossed in the selected icon for the countdown itself.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -3177,7 +3177,8 @@
                 "hide":"Hide Countdown",
                 "loop": "Looping",
                 "decreasingLoop": "Decreasing Looping",
-                "increasingLoop": "Increasing Looping"
+                "increasingLoop": "Increasing Looping",
+                "inherit": "Use Default"
             },
             "EffectsDisplay": {
                 "removeThing": "[Right Click] Remove {thing}",
--- a/lang/en.json
+++ b/lang/en.json
@@ -719,10 +719,9 @@
                 "explanation": "You are adding {class} as your multiclass",
                 "selectDomainPrompt": "Select your new domain"
             },
-            "OwnershipSelection": {
-                "title": "Ownership Selection - {name}",
+            "CountdownPermissions": {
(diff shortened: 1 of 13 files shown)


WBHarry:
The labels on **1** could definately be changed, but I think the relationship there is a good idea.

CarlosFdez:
So this was actually what I was afraid of. This is a bit clearer (I have some suggestions on the actual text) but these are general components that have become very entangled with countdowns.

You should probably rename ownershipSelection to countdownPermissions as it is no longer a general component. You should probably also rename the stuff in general config maybe to be more specific to countdowns. Personally, because countdown hidden is a boolean, I'm thinking perhaps we shouldn't have a config option for "default ownership" and instead bake them into the dialog (its no longer a general element, so we might as well?)

As for making it more intuitive, how about changing the inherit label based on what the setting is. Something like `Default (Hidden)` vs `Default (Visible)` or something?

WBHarry:
The name of the app is bad yeah, it's only ever been used as a specialized thing for Countdowns :+1:

CarlosFdez:
it was fine before, since in theory it could have been used for any similar sort of thing. But now its definitely 100% countdowns only.

WBHarry:
Mm, it -could- have been used for other things. I more so meant that it's only ever been used for Countdowns.
I do think something specialized is the way to go though. `Inherit` is just way to vague, and default doesn't really rhyme with `hide`/`reveal`, which is what GMs experience.

PR #2159, opened 2026-08-05
[Feature] CountdownAction Hidden Label

WBHarry:
Added a nullLabel context prop to NullableBooleanField. 
If you feel another label better communicates that the hidden property gets -hardset- to the current global default, then feel free to change it. I feel it's important to be clear that it does get a static value based on the current default, and doesn't dynamically change it if you flip the global default later.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -486,7 +486,8 @@
                 "types": {
                     "narrative": "Narrative",
                     "encounter": "Encounter"
-                }
+                },
+                "hiddenNullLabel": "Set To Countdown UI Default"
             },
             "CountdownEdit": {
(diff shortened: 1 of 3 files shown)


WBHarry:
image

CarlosFdez:
If it could be dynamic somehow, `Default (Hidden)` vs `Default (Visible)` could be neat but perhaps a pain in the ass to do without making custom selectOptions, which are annoying to do in these custom actions.

CarlosFdez (approved):

CarlosFdez:
However I think this is sufficient. We probably shouldn't be chasing long after minor ux improvements in an area that few people see.

PR #2160, opened 2026-08-06
[Fix] Healing Damage Summary Token Data

WBHarry:
- Fixed an issue where DamageSummary wasn't shown correctly for healing
- Removed separator line between multiple actors in DamageSummary


CarlosFdez (approved):
I see why this was an issue. I think I commented that lowering the portrait size on damage taken by entries might lead to a nicer result, but that can always be done separately.

WBHarry:
Made a little attempt, but it can always be finetuned later aswell :+1:

PR #2161, opened 2026-08-06
[Fix] Summary Speakers

WBHarry:
Fixed damage and effect summary speakers


CarlosFdez (approved):
Untested, but conceptually great to me. I'd be surprised if this had an issue you didn't catch.

PR #2169, opened 2026-08-07
[Feature] Default Folder Entity Type

### LINKED ISSUE #1826: [Feature] Tagable Folders for pre selection
Hi Team,

my Feature Idea is to make adding hombrew stuff a little bit faster and less annoying.
Let a Folder be tagable and use this tag while creating.
For example:
Make a folder in the actor section, tag it with "Environment" and when creating a new actor by clicking the button on the folder, it automatically switches the default on creating an actor within that folder to Environment instead of Character.

That way adding anything in "bulk" won't need to be picked everytime adding sth. or forgetting to change and delete and recreate.

WBHarry:
Fixes #1826 
Set it up so that a folder can have a set default entity that will be the default selection when making a new entity in the folder. Made sure it works in collections too.
image


WBHarry:
It could be preferable to do the handlebars change via injection in `render` instead of changing the template like I did here.
I just felt the template was tiny enough, and probably unlikely to change considering it's basic role that it'd be fine and less ugly.

module/documents/folder.mjs
```diff
@@ -0,0 +1,24 @@
...
+     * @param {string} collectionName
+     * @param {string} pack - Optional pack id.
+     * @returns {DhFolder|null}
+     */
+    static getFolder(id, collectionName, pack) {
+        const collection = 
+            pack ? game.packs.get('world.hm')?.folders : game[collectionName]?.folders;
+            
+        if (!collection) return null;
+        return collection.get(id);
```
CarlosFdez:
```suggestion
        return collection?.get(id) ?? null;
```

module/documents/folder.mjs
```diff
@@ -0,0 +1,24 @@
...
+
+    /** Convenience method to grab a folder via ID from the world collection or a specific pack
+     * @param {string} id - The folder ID
+     * @param {string} collectionName
+     * @param {string} pack - Optional pack id.
+     * @returns {DhFolder|null}
+     */
+    static getFolder(id, collectionName, pack) {
+        const collection = 
+            pack ? game.packs.get('world.hm')?.folders : game[collectionName]?.folders;
```
CarlosFdez: world.hm? That does something or did you mean to put pack in there?

WBHarry: Oh, dang it. That should be pack yeah. That was my first test with the strict value 😆

module/config/flagsConfig.mjs
CarlosFdez: This should poof given the deletion of attachable item.

CarlosFdez:
for the handlebars, guess we'll find out what any repercussions are if anyone reports a module break.

module/documents/folder.mjs
WBHarry: Fixed 🙃

module/config/flagsConfig.mjs
WBHarry: Poof 🪄

WBHarry:
> for the handlebars, guess we'll find out what any repercussions are if anyone reports a module break.

Mm, the possibility would be that some module is doing destructive rerendering and basing it on Foundry's template instead of the actually registered one. IE, wiping out our changes. I guess we'll see!

CarlosFdez (approved):
Hmm, if DhFolder#getFolder() static helper could be avoided, that might be nice. Public static helpers on non-singletons can be a bit unwieldy sometimes. It is a two liner so maybe doable?

Other than that (and the stuff I'm gonna ignore) it seems fine. I haven't tested it, so if someone wants to test that should be good.

WBHarry:
> Hmm, if DhFolder#getFolder() static helper could be avoided, that might be nice. Public static helpers on non-singletons can be a bit unwieldy sometimes. It is a two liner so maybe doable?
> 
> Other than that (and the stuff I'm gonna ignore) it seems fine. I haven't tested it, so if someone wants to test that should be good.

I went ahead and moved it into utils so it's not static 🧑‍🍳

CarlosFdez:
I figured no helper inline would be fine given the helper only saves one line but helper is fine too.

WBHarry:
Actually, yeah, I changed my mind after a few runs through it. Utility not needed. It's just raw now.
image

PR #2171, opened 2026-08-07
[Feature] CountdownUI Empty Render

WBHarry:
Changed so that the countdowns-container with a lot of padding is not rendered if there are no countdowns or they're not visible.

**Old**
image

**Now**
image


CarlosFdez:
Looks good to me.

CarlosFdez (approved):

PR #2177, opened 2026-08-09
[Feature] Changed System Background

WBHarry:
Changed the system background from the book cover to the full panorama as suggested. It does work better for the login screen as the main art pieces are to the side and not blocked. The book cover has the main art right where the login UI is.
image

image


CarlosFdez:
Non blue one looks much nicer yeah.

CarlosFdez (approved):

PR #2178, opened 2026-08-09
Make countdown scrollbar gutter stable

CarlosFdez:
makes it so that a countdown that overflows looks a bit nicer. Very minor thing otherwise.


WBHarry (approved):
:+1:

PR #2179, opened 2026-08-10
[Fix] Prevent domain cards with attribution double scroll bars when editing

CarlosFdez:
Fixes this nonesense
image


WBHarry (approved):
Mon dieu 🫨 :+1:

PR #2182, opened 2026-08-10
[Fix] Improve responsiveness of drag/drop to item sheets

CarlosFdez:
This fixes a weird quirk I ran into. If you try to drag a feature from the compendium into a subclass's or class's features, the feature will not be successfully dragged. You actually need to drag it a second time for it to stick. I don't know why this happens, but its like the event loses all drag data during the super call.

This PR adds something similar to _onDropItem() from the actor sheet, which allows the drag event data to be parsed only once. This seems to circumvent the whole issue.

I could use some help testing, particularly with ancestry, where I don't understand what it is trying to do.

### DIFF
--- a/module/applications/sheets-configs/environment-settings.mjs
+++ b/module/applications/sheets-configs/environment-settings.mjs
@@ -130,6 +130,11 @@ export default class DHEnvironmentSettings extends DHBaseActorSettings {
         event.stopPropagation();
         const data = foundry.applications.ux.TextEditor.implementation.getDragEventData(event);
         const doc = await fromUuid(data.uuid);
+       
+        if (data.type === 'Item') {
+            return super._onDropItem(event, doc);
+        }
+
         if (doc?.type === 'adversary' && event.target.closest('.category-container')) {
             const target = event.target.closest('.category-container');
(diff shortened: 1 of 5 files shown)


WBHarry:
I tested around on this and got verification from Kim at Foundry after about it after.

The issues we're running into is a base Javascript one. You can only read `event.dataTransfer` syncroniously.
We are doing an async with `fromUuid` - after that call, event.dataTransfer will no longer be available.

That's why it failed specifically for the compendium with an -unloaded- item, since `fromUuid` only actually needs to await in that situation. The second pull from compendium of the same item has it loaded, and therefore it stays in sync time.

WBHarry:
https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/Drag_data_store#protected_mode

WBHarry:
I went over our uses of `_onDrop`. We have four cases where we run something async and then pass the event intp `super._onDrop`
```
actor-setting.mjs
environment-settings.mjs
class.mjs
subclass.mjs
```

`onDropItem` is a valid way to get around this as you did, aslong as we're not ending up with any double logic because of `super.onDrop` previously having been blocked but now getting to run.

PR #2183, opened 2026-08-10
Add overpass font, no uses yet

CarlosFdez:
Will be used for certain projects, but for ease of developers, it is being added on its own early.

We need to eventually make fonts actual files we bundle, for convenience and reliability reasons. Minor task though.


WBHarry (approved):
:+1:

PR #2189, opened 2026-08-11
Prefer item roll data as a base for actions

CarlosFdez:
In an action, get actor() requires retrieving the item's actor, so if there is no item, there is no actor regardless.


CarlosFdez:
I slipped in fear, but let me know if fear should instead be an object with value/max like a regular resource. That might allow us to cap something to max fear maybe.

WBHarry (approved):
I think it's perfectly fine having `rollData.fear` just be the current value. I struggle to see a case where the maximum would be used, especially since the maximum changing is beyond rare.

PR #2190, opened 2026-08-12
[Fix] active effects being created twice when dragged

CarlosFdez:
This does two things, let me know if it outta be split, particularly if you're uncertain about either solution

1) Fixes dragging an active effect to an item creating two of said effect (regression from prev release)
2) Allows drag dropping effects to an action's effects fieldset. This will hopefully make it slightly easier to reuse restrained and vulnerable for Hope/Fear data entry

The changes to facilitate the second might contribute to somewhat cleaner code in the evolution PR, once merged in, though it'll be a bit of an annoyance. It could probably be cleaned further particularly with code in the mixin, but I tried to show restraint and caution right regarding existing functionality.

### DIFF
--- a/daggerheart.d.ts
+++ b/daggerheart.d.ts
@@ -22,12 +22,6 @@ import DhEffectsDisplay from './module/applications/ui/effectsDisplay.mjs';
 // This declare global hopefully fixes that
 // Note: eslint is not aware of these, whatever is added here should go in the eslint's globals list
 declare global {
-    // These are convenience types for common imported things. This allows them to be used in JSDoc directly
-    // For actual use such as instanceof, an import is still required
-    type DHItem<T extends BaseDataItem = BaseDataItem> = InstanceType<typeof documents.DHItem<T>>;
-    type DhpActor<T extends BaseDataItem = BaseDataItem> = InstanceType<typeof documents.DhpActor<T>>;
-    type DhActiveEffect = InstanceType<typeof documents.DhActiveEffect>;
-
     /**
      * A simple event framework used throughout Foundry Virtual Tabletop.
      * When key actions or events occur, a "hook" is defined where user-defined callback functions can execute.
--- a/daggerheart.d.ts
+++ b/daggerheart.d.ts
@@ -22,6 +22,12 @@ import DhEffectsDisplay from './module/applications/ui/effectsDisplay.mjs';
 // This declare global hopefully fixes that
 // Note: eslint is not aware of these, whatever is added here should go in the eslint's globals list
 declare global {
+    // These are convenience types for common imported things. This allows them to be used in JSDoc directly
+    // For actual use such as instanceof, an import is still required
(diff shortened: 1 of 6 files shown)


WBHarry:
Hm. What's the point of dragging AE's to actions?
Actions that apply effects always have a bespoke AE not otherwise in use of the Feature. I don't think we have a single case where it's shared with the future in any way.

WBHarry:
Even if the effect is just a `restraint` or `vulnerable` vehicle, it'll still want to have a correct image and possibly description, so the only thing dragging in an existing one seems to do to me is to swap two clicks to open statuses and click in the one you want, compared to now searching up whatever you want to drag which is probably rougher. 🤔

CarlosFdez:
> Hm. What's the point of dragging AE's to actions? Actions that apply effects always have a bespoke AE not otherwise in use of the Feature. I don't think we have a single case where it's shared with the future in any way.

Reuse and repair primarily, especially if multiple creatures inflict the same bespoke condition. Though I haven't found an example of sharing bespoke conditions in official stuff yet, it could happen for homebrew. 

Temporarily Restrained/Vulnerable were short term examples only, they're first class enough that we should be providing better ways. I'm actually noting right now how often in our compendium an ability that applies temporary vulnerable/restrained is actually doing it via a bespoke status without a good description.

https://github.com/user-attachments/assets/9fd080a2-bd2f-4daa-87bc-4eef652ae0cc

CarlosFdez:
In short, it seemed a basic enough useful thing to support that was also an inefficient solution to a problem we don't have a good idea how to solve yet, rather than the way to solve the problem specifically.

CarlosFdez:
Sorry for the triple, but I noticed Vicious Entangle has Restrained twice in the current compendium due to existing limitations, and it was probably entered by hand twice. It might be possible to detect drag/drop on the same item and instead add to the list of ids rather than creating an active effect in that scenario maybe?

CarlosFdez:
because it looks like extra stuff has to be done for it anyways, and you're not certain about it, I'll split off the dragging bits.

WBHarry:
Hm, alright-y. I'm definately not strongly against it, I just couldn't quite see the usecase for myself at least.

If we add functionality to have `simple`/`basic` AEs that just act as pure vehicles for adding solely `restrained` and the other base conditions, then I would definately see a usecase.

WBHarry (approved):

CarlosFdez:
The primary use case for this feature is strictly for effect reuse between actions, and for 2 actions to point to be able to the point to the same effect, instead of duplicating it in the case of restrained. Any mentioned of restrained/vulnerable was mostly for "this becomes slightly less of a pain in the ass too in the meantime".

How to solve the reuse case was easier than the "how to add pre-registered effects" case.

WBHarry:
Right, I don't recall where right now, but I know there was some usecase where I thought we'd need to select an already existing effect. My thinking then was that we'd eventually add a select where an existing effect could be picked in the ActionDialog to add that one instead of a new one

PR #2192, opened 2026-08-12
[Data] Convert special characters in extraction and fix common errors

### LINKED ISSUE #2186: [Feature] Purge certain unicode characters from the data
Certain unicode characters make searching/matching trickier or contribute little. While we should keep emdash, there are certain others that should be removed:

`’ ” “ −` should be replaced for quote, double quote, double quote, and hyphen. The dash here isn't emdash, but a subtraction symbol.

A good place to do this might be to in the extract script, to do some simple string replacements on description and action descriptions. This would remove them once and for all.

CarlosFdez:
I was gonna wait till after hope/fear but these little quirks actually are causing some awkwardness with finding errors in new data.

Closes https://github.com/Foundryborne/daggerheart/issues/2186

This add some automated to the extraction to remove certain special characters, duplicate action descriptions, and ownership data. I also manually removed certain cases of empty paragraphs (primarily beginning of string, end of string, and final list item). I also got rid of a lot of copy pasta classes on p tags (I didn't touch li tags). For empty paragraphs, its probably a thing we could automate for start of/end. I'll think about it.

I would appreciate as many people as possible looking at this.


WBHarry (approved):
:+1:

PR #2194, opened 2026-08-12
[Fix] Damage OnSave Not Working

WBHarry:
Fixed so that lowered damage on save works again. With the Targeting rework the save info on the chat message was swapped from `target.saved?.success` to `target.saveResult?.success` - this seems to be the last place where that was forgotten 🙃


CarlosFdez (approved):

CarlosFdez:
I'll trust it works!

PR #2196, opened 2026-08-12
Add getter-return linter rule, fix violations

CarlosFdez:
Just internal api stuff.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -100,6 +100,7 @@ export default defineConfig([
                 }
             ],
             'array-callback-return': 'error',
+            'getter-return': 'error',
             'no-async-promise-executor': 'error',
             'no-const-assign': 'error',
             'no-dupe-class-members': 'error',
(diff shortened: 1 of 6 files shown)


WBHarry (approved):
Hm..it's worth noting that this is not a rule that Foundry runs by.
They quite often distinguish between returning undefined via `return;` and `return false;`.

I don't think this will ever become an issue for us in interfacing with core, but if it does we can just remove the rule at that point.

CarlosFdez:
I think we can explicitly return undefined in that case if we need to make such a distinction. Its more the unexplicit nature of it that the rule is for.

CarlosFdez:
it could be that I made the wrong assumption of it, and isCritical for not evaluated should instead be null?

WBHarry:
No, isCritical is definately a `falsy` thing.

I think you're right. Any case where we -do- want a null/undefined result can just explicitly do that.

CarlosFdez:
Wait till next release I think. I can see a case where isCritical can be null for "hasn't rolled yet, so its neither critical nor non-critical".

PR #2199, opened 2026-08-13
[Fix] DamageBonus Subtract Handling

WBHarry:
Added a fix so that add and subtract are handled correctly.

It revealed that we don't really handle any other change types though. I added a `TODO` since it's not a trivial thing, and a fair bit of thinking about cases and what we want is needed.


CarlosFdez (approved):
This looks good. And actually if this is how damage is tallied up, it looks like its actually pretty easy to create a breakdown. May be worth creating a breakdown and then having a simplification pass on the formula. Any successfully resolved upgrade or downgrade would have to clobber the breakdown though.

I suppose multiplies and divides would be pretty useful to handle certain things like first strike and ghost transform's "weakness" to magic damage.

PR #2201, opened 2026-08-14
Shorten heritage item descriptions and create journal registry

CarlosFdez:
This will allow us to eventually create tooltips for these item types. The journal registry was created to support the ability for modules to register replacement journals, and exists for a few reasons:

1) Our journals can only be linked via anchors, and uuid redirects don't seem to support that (or mix/matches). This allows replacement journals to be via anchors or directly to new journal pages.
2) This adds some resiliency. If we ever change up our journals, we don't suddenly have to migrate stuff or create yet-more redirects. We can simply update the registry.

Property names and the actual config can be changed. We can also remove the check for descriptions if you want.


WBHarry (approved):
I think it's good as is 👍 

Some gnarly work there with the migration 😁

CarlosFdez:
I hope I didn't miss any cases with the migration, but there's no way to be sure I suppose rather than see what happens.

PR #2207, opened 2026-08-15
[Fix] using world item copies of classes in character creator

CarlosFdez:
Text by dragging bard to your world items, then going through character creation, and then drag the bard in your world items instead of the compendium. Afterwards complete the character creation. Before this PR, the subclass would fail to be added to the character at the end, as the class newly added would not have preserved compendium source info.


WBHarry (approved):
Verified the issue, and it's a good fix :+1:

PR #2216, opened 2026-08-16
[Fix] missing feature lookup errors in ancestry/subclass sheets

### LINKED ISSUE #2206: [Bug] Ancestries with missing features are breaky
I have encountered the following issues. This will become a problem once voidborne removes certain items.

1) The sheet can't open due to features html choking
2) Even if you fix 1, you can't replace the missing primary/secondary feature in the item. Its because the feature still exists.
3) If you fix 1, then try to delete a feature that does exist while invalid features exists, it explodes. This is due to the map in the final update in base-item's #deleteFeatures.

I don't have time to fix it right now but logging as something that should be fixed before hope/fear releases.

CarlosFdez:
For data preparation, features now have a system data only uuid property that will always have the actual uuid. tab-features.less was also removed and replaced with feature-section.less (and renamed to tab-features). The former was only in use by a single sheet.

Closes https://github.com/Foundryborne/daggerheart/issues/2206


CarlosFdez:
To test, install voidborne, add some subclasses/class/ancestry imports, then disable voidborne.

WBHarry (approved):
Works :+1:

Nice job on the `rankFeatures` inline

templates/sheets/items/subclass/features.hbs
```diff
@@ -32,48 +32,27 @@
...
-            <a><i data-action="addFeature" data-type="mastery" class="fa-solid fa-plus icon-button"></i></a>
+            {{label}} 
+            <a><i data-action="addFeature" data-type="{{type}}" class="fa-solid fa-plus icon-button"></i></a>
         </legend>
 
         <div class="feature-list">
-            {{#each source.system.masteryFeatures as | feature |}}
-                {{> 'systems/daggerheart/templates/sheets/global/partials/feature-section-item.hbs' type='mastery' feature=feature}}
+            {{#each features as | feature |}}
+                {{log feature}}
```
CarlosFdez:
```suggestion
```

CarlosFdez: I slipped something in, whoops

PR #2217, opened 2026-08-17
Avoid line wrap with custom weapon features

CarlosFdez:
Internally, it works by making weapons and armor use the same features template as the other items

Before
image

After
image


WBHarry (approved):
:+1:

PR #2218, opened 2026-08-17
Add support for embedded item tables

CarlosFdez:
image

Supports either `@EmbedTable[UUID1 UUID2 UUID3]` or `@EmbedTable[path=some.path.to.uuids]`. This also adds support to a new `@Resolve` enricher that simplifies to a number (and also creates a tooltip for any fallback labels).

Currently only weapon and armor is supported. Consumables and Loot have a numerical "rarity value" that I don't know yet how to reproduce, and we need to discuss it. I'm also not sure about the enricher name, and am wondering if something like `@Compute` or `@Evaluate` might be better.


WBHarry (approved):
Awesome work here getting the resolve enricher in aswell :+1:

PR #2221, opened 2026-08-17
Fix comment typos and sort rolltable

CarlosFdez:
The sort is a thing I saw core do for an embed, so I'm thinking its useful to have for safety reasons.


WBHarry (approved):
⭐ 👌

PR #2224, opened 2026-08-18
[Houskeeping] Feature Granter Data

WBHarry:
Moved existing fields related to feature origin into granter schema that now includes ID


CarlosFdez:
Last thing I pushed will cause newly created characters to break. The ids of the grants were actually not valid since foundry will create new ids if you do not set keepId. Once fixed, the character creation should be good again.

The preCreate stuff is intended to kill the granter data if dragged to a world item/compendium/other actor. This should hopefully keep our extracted data clean.

EDIT: should now be resolved, though I suppose still test character creation and levelups.

CarlosFdez (approved):

PR #2225, opened 2026-08-18
[Feature] support lookup enricher types and index adjustments

CarlosFdez:
Adds support for `@Lookup[@item.attack.range|type:range|adjustment:2]{some fallback label}` which will be handy for certain H&F features.


WBHarry (approved):
Works great. Awesome work :+1:

PR #2226, opened 2026-08-19
[Fix] error in char creator if items exist with invalid domains

CarlosFdez:
Minor fix that is probably beneath patch note mention, but it occurs if there are compendium items for domains that are unregistered.


WBHarry (approved):
:+1:

PR #2228, opened 2026-08-20
[Feature] Duality Roll Rework

WBHarry:
Swapped to setting up DiceSoNice appearance based on modifiers.

There's a lot of code that can probably be ripped out now - everything relating to HopeDie/FearDie/AdvantageDie/DisadvantageDie and the ugly getter/setters in the Roll classes.
I think that should be done after this however, since that's more prone to breaking something.

### DIFF
--- a/module/dice/baseRoll.mjs
+++ b/module/dice/baseRoll.mjs
@@ -14,22 +14,24 @@ export default class BaseRoll extends foundry.dice.Roll {
     async _evaluate(options) {
         await super._evaluate(options);
         
-        const hopeRoll = this.dice.find(x => x.modifiers.includes('h'));
-        const fearRoll = this.dice.find(x => x.modifiers.includes('f'));
-        if (hopeRoll && fearRoll) {
-            const diceSoNice = 
-                game.settings.get(CONFIG.DH.id, CONFIG.DH.SETTINGS.gameSettings.appearance).diceSoNiceData;
-            const isCritical = hopeRoll.total === fearRoll.total;
-            hopeRoll.options.sfx = getDiceSoNiceSFX({ 
-                critical: isCritical, 
-                higher: hopeRoll.total > fearRoll.total,
-                data: diceSoNice.hope.sfx
-            });
-            fearRoll.options.sfx = getDiceSoNiceSFX({ 
-                critical: isCritical, 
-                higher: fearRoll.total > hopeRoll.total,
-                data: diceSoNice.fear.sfx
-            });
+        if (game.dice3d) {
+            const hopeRoll = this.dice.find(x => x.modifiers.includes('h'));
+            const fearRoll = this.dice.find(x => x.modifiers.includes('f'));
+            if (hopeRoll && fearRoll) {
+                const diceSoNice = 
+                    game.settings.get(CONFIG.DH.id, CONFIG.DH.SETTINGS.gameSettings.appearance).diceSoNiceData;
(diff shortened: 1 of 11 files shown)


module/dice/die/baseDie.mjs
```diff
@@ -38,6 +43,30 @@ export default class BaseDie extends foundry.dice.terms.Die {
...
+
+    async fear() {
+        this.#setDualityDiePreset('fear');
+    }
+
+    async advantage() {
+        this.#setDualityDiePreset('advantage');
+    }
+
+    async disadvantage () {
```
CarlosFdez:
```suggestion
    async disadvantage() {
```

module/dice/baseRoll.mjs
```diff
@@ -8,4 +10,28 @@ export default class BaseRoll extends foundry.dice.Roll {
     get modifierTotal() {
         return this.total - this.dice.reduce((acc, dice) => acc + dice.total, 0);
     }
+
+    async _evaluate(options) {
+        await super._evaluate(options);
+        
+        const hopeRoll = this.dice.find(x => x.modifiers.includes('h'));
+        const fearRoll = this.dice.find(x => x.modifiers.includes('f'));
+        if (hopeRoll && fearRoll) {
```
CarlosFdez: Needs to check if DSN is enabled

module/dice/dualityRoll.mjs
```diff
@@ -135,11 +134,13 @@ export default class DualityRoll extends D20Roll {
         this.terms = [this.terms[0], this.terms[1], this.terms[2]];
 
         this.terms[0] = new game.system.api.dice.diceTypes.HopeDie({
-            faces: this.terms[0]?.faces ?? this.data.rules.dualityRoll?.defaultHopeDice ?? 12
+            faces: this.terms[0]?.faces ?? this.data.rules.dualityRoll?.defaultHopeDice ?? 12,
+            modifiers: ['h']
```
CarlosFdez: Would it be possible to bundle this in the constructor of the hope die class, if it doesn't already exist?

WBHarry: I'd like to postpone improvements here to after this PR.

I think a lot of this will just get ripped out. ⚔️

WBHarry: Or wait, no that's a good idea.

WBHarry: ..or maybe not, since the HopeDie class might get removed wholesale. 
I'll refrain from further comments in favor of sleep 😆

module/dice/baseRoll.mjs
WBHarry: Good catch. I don't think it actually hurts anything, but it's unncessary work.

I added it here, and also in the modifiers section on `BaseDie` :+1:

module/dice/dualityRoll.mjs
CarlosFdez: Even if it gets ripped out, less side effects to the rest of the code using the roll classes should hopefully make it easier to rip out later. You'd basically alter the modifiers in the constructor of the hope die before the super call (checking that h isn't already in there), so it wouldn't be all that difficult either.

WBHarry: Yah that's fair 👍 

I moved the addition of `h` and `f` modifiers to the constructors for each dice.

CarlosFdez (approved):
Looks good. I think you were planning on doing more so I won't merge it, but I like what I see so far.

PR #2229, opened 2026-08-21
Add links between features & granter, subclass & class

Dapoulp:
- FEATURES:
   1.  Add links between Features & granter, subclass & class
   2. Update default itemFeatures

- REFACTORS:
   1.  Move domainCards Vault/Loadout functions to DHDomainCard
   2.  Update toggleVault function to handle recall cost dialog
   3. Update downtimeTrigger hook to avoid PartySheet import into socket

- FIXES:
   1.  Fix circular dependencies


module/data/item/base.mjs
```diff
@@ -124,7 +124,7 @@ export default class BaseDataItem extends foundry.abstract.TypeDataModel {
     }
 
     get itemFeatures() {
-        return [];
+        return this.actor?.items.filter(i => i.system.granterItem === this.parent);
```
CarlosFdez:
```suggestion
        return this.actor?.items.filter(i => i.system.granterItem === this.parent) ?? [];
```

CarlosFdez:
I'm not 100% sure of subclass.class resolving to the character's class (or not being a async function that falls back to the compendium), but that's more of a concern with the name of the methods (and them being getters) than anything substantial. Maybe something like subclass.classItem or subclass.getConnectedClass() or something. If Harry doesn't see an issue or has any suggestions then I suppose its fine.

Are there any internal refactors that could be had from this as well? For example, would some of the preUpdate checks involving class/multiclass checks improve from this?

Dapoulp:
Yeah, I agree about naming, I did named it classItem, then removed it xD
I'll change that.

What don't you like with them being getters ?

About the fallback to compendium, it was a part of the goal, getting the real item, and not having to do async

CarlosFdez:
Getter implies something that already exists or is intrinsic to the item. In this case, its really just doing an actor lookup. It might as well have been an isSubclassOf() for the check method and then an `actor.items.find(c => this.isSubclassOf(c))`.

Its quite likely I might be overreacting, but getters are commonly assumed to be fast and efficient, so subclass.class might end up being called multiple times in a handlebars template. Its a common crime we do though. But assigning `this.classItem` during data preparation might avoid that trap and keep access simple?

For the async, it depends on how it can be used in the codebase, but most uses I imagine would want to get the class item at any cost for rendering. I could be wrong, perhaps we don't want subclass cards to ever grab the name of the class on the actor, and to always go into a compendium. In general I prefer api methods when possible to benefit the codebase, and then be made generic enough to support modules. This feels basic enough that I imagine there must be core uses.

CarlosFdez:
I'd wait for harry's input before doing code changes based on what I said, I'm genuinely not sure what the right path is here.

Dapoulp:
I renamed them to classItem & subclassItem and moved them into prepareDerivedData

WBHarry (approved):
Hm. This is fine for me as such if you think it's useful.

Typically I'd say that there isn't really a point to introduce getters and prepared properties stand-alone. They should be introduced for a reason - because they're helpfull/handy to solve a problem.

Dapoulp:
I didn't do it for fun, but because it's useful xD

module/systemRegistration/socket.mjs
```diff
@@ -16,7 +16,7 @@ export function handleSocketEvent({ action = null, data = {} } = {}) {
             Hooks.call(socketEvent.Refresh, data);
             break;
         case socketEvent.DowntimeTrigger:
-            PartySheet.downtimeMoveQuery(data);
+            Party.downtimeMoveQuery(data);
```
CarlosFdez: I think this method should either be moved to the party from party sheet, or perhaps the contents of downtimeMoveQuery should be executed here directly (edit: or a hook maybe?)

Sheets may be registered by module and attempting to grab one via import or our own exports ignores those.

Dapoulp: I didn't move the function itself (maybe it should be) but registered hook trigger like the others, so no need to import PartySheet into socket anymore

PR #2230, opened 2026-08-22
Update styling of tooltips of certain item types to more closely match cards

CarlosFdez:
Also adds `@Embed` support for those item types. When an item supports embedding, that embed is what will be displayed in a tooltip. There are also tooltip changes for the other types, with the hope of making them easier to maintain and more closely align visually.

While cards embedded into something will use the theme of the application, tooltips are currently stuck to dark mode. We'll have to consider an appearance setting.


Dapoulp:
I was thinking about few "improvments" in term of consistency:

1. Adding the same global design to all items tooltip even if there is no IRL card version (like weapons, etc)
2. Adding "granter item tooltip style" to features. For example, a feature from Ancestry get Ancestry tooltip design
3. Adding middle-click to lock tooltip on all of them. Right, you can see an Ancestry tooltip with the scrollbar on the description, but you can't scroll it (you have to open the Item sheet)

styles/less/global/cards.less
```diff
@@ -0,0 +1,499 @@
...
+ */
+.dh-card {
+    --width: calc(18.25rem);
+    --initial-portrait-ratio: 0.665;
+
+    position: relative;
+    border: 1px solid @card-trim-color;
+    border-radius: 12px;
+    width: var(--width);
+    height: calc(var(--width) * 1.4);
```
Dapoulp, line 115: Better to do:
`aspect-ratio: 5/7;`

styles/less/global/cards.less
```diff
@@ -0,0 +1,499 @@
...
+            pointer-events: none;
+            user-select: none;
+        }
+    }
+
+    > article {
+        flex: 1;
+        display: flex;
+        flex-direction: column;
+        max-height: calc(25rem - 7.25rem); // height of card - min portrait height
```
Dapoulp, line 168: `max-height: 17.75rem` ?

styles/less/global/cards.less
```diff
@@ -0,0 +1,499 @@
...
+    float: left; 
+
+    * {
+        scrollbar-color: light-dark(@dark-blue, @golden) transparent;
+    }
+
+    > header.dh-card-header {
+        position: relative;
+        max-height: calc(var(--width) * var(--initial-portrait-ratio));
+        flex: 0 1 calc(var(--width) * var(--initial-portrait-ratio));
```
Dapoulp, line 138: Could be:
```
> header.dh-card-header {
   ...
   --header-max-height: calc(var(--width) * var(--initial-portrait-ratio));
   flex: 0 1 var(--header-max-height);
   min-height: calc(var(--header-max-height) * 0.75);      // whatever min-height is, 0.75 for the example
   ...
}
...
> article {
   ...
   <remove flex: 1;>
   min-height: 0;
   ...
}
```

It reduces the number of calc and you control all those dimensions in a single element

WBHarry (commented):
Gone over it, and I think it's very well put together. The few things I was pondering on I become convinced over when grasping the totality of things, so for my part it's all good :+1:

CarlosFdez:
> I was thinking about few "improvments" in term of consistency:
> 
>     1. Adding the same global design to all items tooltip even if there is no IRL card version (like weapons, etc)
Further improvements to the basic item tooltip to get more in line might be nice, but I'm not sure where to further take them. I made some visual adjustments to look similar but they don't really have a trim I can use.

>     2. Adding "granter item tooltip style" to features. For example, a feature from Ancestry get Ancestry tooltip design

I don't think we should render the features tooltip with the visual flair of the granter. Those are genuinely features, not actual cards, and might throw someone off. We'd need to put "Feature" where the word "Ancestry" is for example, so its so many different flavors of feature. There is also the subtle difference of domain card and subclass card, including in how subclasses do the gold text, and how domain cards do the text inversion.

A case could be made for not showing the feature at all and instead showing the card it came from though. If more people agree with that, we could maybe do that. I'd need to handle subclass specialization/foundation/mastery but no biggie.

>     3. Adding middle-click to lock tooltip on all of them. Right, you can see an Ancestry tooltip with the scrollbar on the description, but you can't scroll it (you have to open the Item sheet)
I"d need more consensus on this. The general vibe I got for actual card tooltips is that this wasn't desired, and it does mean I'd have to make them taller and break aspect ratio to be able to fit the really tall cards without scrolling. When I last asked about this elsewhere, the reply I got was that it wasn't needed for these types of cards since scrolling will usually not happen and didn't pursue it.

Dapoulp:
I really don't like the fact sometime you have a nice tooltip, sometime a basic one, it give a feeling of non-finished design.

About the "granter tooltip style", I did it on my module and you get something like:
image 
image


About the middle-click, it's just how most of Foundry systems work. You don't have to add the text about it, just the possibility, and everybody is happy. Some of them even give you the possibility to "lock" the tooltip to move it and keep it if you want to.

WBHarry:
Hm. My opinion:

1) I think it makes sense for the Feature tooltips to be simpler. They will never have any bespoke, extensive artwork - only the foundry core images, so there isn't a need to put expansive focus on the artwork. 
I suppose we could -potentially- add in some manner of trim element - but we'd have to decide on a single look for it in that case.

2) I don't like this idea. If there is to be any merit to it, it'd have to include both the granter type + that's it's a feature, and writing out both `Subclass` and `Feature` is quite exessive in overlaying everything, and there won't be two "boxes" to put it in.
If it just says `Subclass` then it's confusing and strange since it's not a subclass. If it just says Feature, then why is it not uniform?
Normally you access your features through the sheet where they're put in categories for `Subclass`, `Ancestry` etc, so where they're from should normally come from context.

3) Hm..don't have a strong opinon on this. If all the core cards do not break, then I don't see it as something that needs immediate focus.
It would assumably be an improvement if middleclick could do the normal lock+scroll functionality.

CarlosFdez:
If middle click isn't working at all, its a bug, though now I remember that I may have not recreated the color highlight, so I should do that. I'll poke over that.

The "trim" element I added was a golden border bottom but I wouldn't be against changing it. We could also do something like the move tooltip as well if we want to instead put features and equipment towards the simpler side.

Dapoulp:
About "granter style", I totally disagree, but eh, do want you want

middle click is a "global" foundry behavior for tooltip, not a Daggerheart/Foundryborne specific.
People using other systems are used to it and will certainly ask for it

Also, we already have it on some of them

EDIT: It seems it still works for domain card for example

CarlosFdez:
For features and equipment, we could maybe go simpler instead and do something like these but with tags? If the issue is that it feels half finished. More of a @moliloo question 
image

CarlosFdez:
For those who haven't checked out the project yet, equipment currently look like this 
image

moliloo:
> For features and equipment, we could maybe go simpler instead and do something like these but with tags? If the issue is that it feels half finished. More of a @moliloo question 
> image
> 

I agree, at this point, weapons has more tags than i want making more rool, leaving it like a simples tooltip like rest move descriptions should be a neat approach

CarlosFdez:
Awesome! I'll merge once approved and start on that, in case anyone wants to back out when we do. Thank you.

WBHarry (approved):
I think that's a great idea. I'm a lover of big pictures, but this will make those tooltips really dense. Will be interesting to see.

styles/less/global/cards.less
CarlosFdez: Calc of this one is actually wrong. Should be using var(--height) or otherwise mathing it out.

styles/less/global/cards.less
CarlosFdez: I had originally tried to centralize it but failed and moved on. Looks like min-height: 0 did the trick, thanks!

CarlosFdez: Actually, looks like its causing some issues. On the left shouldn't have a scrollbar

image

Attribution came up, though it could be solved with margin-top: auto
image

I'll use variables more liberally here but I'll have to keep the current methodology, though looks like I can kill max height on the header at least.

PR #2231, opened 2026-08-22
[Feature] Otherwordly

WBHarry:
Added support for `Otherwordly` via a rules effect that allows you to select your damage type in the DamageDialog.
image

image

----
Implementation details to note:
- I added a user flag that stores which damage type the user chose last. This is used for which initial damage type should be selected when selection is available in the DamageDialog. Since I put it on the user, it does mean that if you have more than one items/things that let you select damage type, they don't individually track your last preference. 
I think this is fine though, or perhaps even preferable.
- The way I set it up only really works for a weapon. I added a check so that the choice is only available if the DamageDialog is popped by the item that has the ActiveEffect rule on itself. This is so that someone with a `Ghostblade` doesn't get to choose to make their fireballs physical.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1931,6 +1931,16 @@
                         }
                     }
                 },
+                "otherwordly": {
+                    "name": "Otherwordly",
+                    "description": "On a successful attack, you can deal physical or magic damage.",
+                    "effects": {
+                        "otherwordly": {
+                            "name": "Otherwordly",
+                            "description": "On a successful attack, you can deal physical or magic damage."
+                        }
+                    }
+                },
                 "painful": {
                     "name": "Painful",
                     "description": "Each time you make a successful attack, you must mark a Stress.",
--- a/module/applications/dialogs/damageDialog.mjs
+++ b/module/applications/dialogs/damageDialog.mjs
@@ -27,6 +27,7 @@ export default class DamageDialog extends HandlebarsApplicationMixin(Application
         },
         actions: {
             toggleSelectedEffect: this.toggleSelectedEffect,
+            updateDamageType: DamageDialog.#onUpdateDamageType,
             updateGroupAttack: this.updateGroupAttack,
             toggleCritical: this.toggleCritical,
             submitRoll: this.submitRoll
@@ -76,6 +77,25 @@ export default class DamageDialog extends HandlebarsApplicationMixin(Application
 
         context.damageOptions = this.config.damageOptions;
         context.rangeOptions = CONFIG.DH.GENERAL.groupAttackRange;
+        
+        if (this.config.canChooseDamageType) {
+            const previousDamageType = 
+                game.user.getFlag(CONFIG.DH.id, CONFIG.DH.FLAGS.userFlags.previousDamageTypeChoice);
+            context.selectedDamageTypes = 
+                previousDamageType ? new Set([previousDamageType]) : damageFormula.damageTypes;  
+
+            context.canChooseDamageType = true;
+            context.damageTypesChoices = Object.entries(CONFIG.DH.GENERAL.damageTypes).reduce((acc, [key, data]) => {
+                acc[key] = {
+                    ...data,
+                    selected: context.selectedDamageTypes.has(key)
+                };
+
+                return acc;
+            }, {});
+        } else {
+            context.selectedDamageTypes = damageFormula.damageTypes;
+        }
 
         return context;
     }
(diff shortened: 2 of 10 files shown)


CarlosFdez:
I'll check it out, though part of me is thinking it should be a weapon property and a feature that edits the weapon, so that damage formulas can show an icon for it. But perhaps there is merit in allowing an actor at any point, independent of weapon, to deal any type of damage.

That said I think most of these are predicated. There is also the crimson rite in the void, which is a buff that gives a specific weapon that property.

WBHarry:
Hm, in that case we've arrived at another thing that wants to have ActiveEffects that apply to Items

WBHarry:
An alternative logic we could use instead of having a rule would be to make it so that if any instance of damage has no DamageTypes - IE just an empty array, then damage has to be selected.

CarlosFdez:
> Hm, in that case we've arrived at another thing that wants to have ActiveEffects that apply to Items

Either that or its a case of a conditional (though AEs want to apply during data prep, not conditionally to circumstances. We kinda force it to work for the damage formula pipeline but its non-standard. That's why 5e started their applied rules thing I imagine). While I'm decent at following code, I don't have as much of a grasp of what the downsides would be with each approach as I imagine you do at the moment.

Found one more example btw. Brawler's have a stance called Otherworldly, which makes *any attack you do* otherworldly. So it seems to be leaning as either multiple ways to do it, conditionals, or conditional item applying effects (an item applying effect may be on the actor but specify either a specific item or types of items it'll apply to). This is quite the mess actually.

> An alternative logic we could use instead of having a rule would be to make it so that if any instance of damage has no DamageTypes - IE just an empty array, then damage has to be selected.

That could be what an item modifying AE ends up doing under the hood, funny enough. In hope/fear, on shadowblade, it looks like it does phy/mag as the damage type of a weapon, so another approach might be if a specific attack's base damage is configured as being both, it triggers a choice.

WBHarry:
We can't base it on having both, because a damage instance -can- actually have both. It has mechanical ramifications that I -think- we've implemented.
image

-----
Hm. The existance of the brawler's outerwordly makes me thing that we should probably mix:
- If a damage instance has no DamageTypes -> You have to pick one when rolling damage with it.
- We also add a rule/some rules to be able to specifiy a global state of getting to pick.

WBHarry:
The `no damage types` solution relies on AE or something else modifying the item with the `Outerwordly` feature.
We can of course cheat and not solve that right now - just manually set the weapons with it to have no DamageTypes and wait with the actual weapon feature.

CarlosFdez:
> We can't base it on having both, because a damage instance -can- actually have both. It has mechanical ramifications that I -think- we've implemented. image


My thought was to do that based on the *base weapon* being both, not the final result. But I suppose that prevents AE solutions, so maybe not.

> 
> Hm. The existance of the brawler's outerwordly makes me thing that we should probably mix:
> 
>     * If a damage instance has no DamageTypes -> You have to pick one when rolling damage with it.
> 
>     * We also add a rule/some rules to be able to specifiy a global state of getting to pick.

Could work. I think its either that or item altering AEs with dynamic application rules on what items to apply to (brawler would apply it to all items ever). That would skip needing global state.

WBHarry:
Ah, I see what you mean with dynamic application. It's an interesting thought.

Whichever way we go with the AE's though, that'll be a big system add. Maybe it makes sense to modify this PR to use the: No DamageTypes -> Select one, and nothing else. And we make the `Outerworldly` weapons have no DamageTypes, leaving the WeaponFeature itself until later.

WBHarry:
Or maybe we do include the Outerwordly feature, just so it's present in the descriptions of these weapons.
That does of course mean if someone homebrew and uses it, they'll be confused why it doesn't actually do anything.

Dapoulp:
Don't we use the empty damage type array to manage untyped damage ? (that bypass resistances)

WBHarry:
There is no such thing as untyped damage, no?

WBHarry:
Or I suppose maybe falling damage and such

Dapoulp:
I remember there was. Let me check

Dapoulp:
I think there was some oversight before that got fixed. But I'm pretty sure we had that discussion and asked 2 or 3 times to Cosmo to give me the same example again

EDIT: Direct Damage is an option into damage part (for weapon etc), it exists but we manage it this way, so empty array can be used for what you wanted to I guess

PR #2234, opened 2026-08-22
[Feature] Generalized Include Base Damage

WBHarry:
- Changed so that includeBaseDamage can be used by any attack/damage action. 
- Went over SRD Adversaries, updating them to use this functionality

### DIFF
--- a/module/applications/sheets/api/actor-setting.mjs
+++ b/module/applications/sheets/api/actor-setting.mjs
@@ -89,7 +89,7 @@ export default class DHBaseActorSettings extends DHApplicationMixin(ActorSheetV2
     async _onDropItem(event, item) {
         if (item?.type === 'feature') {
             if (item.parent?.uuid === this.actor.uuid) {
-                return super._onDropItem(event);
+                return super._onDropItem(event, item);
             }
 
             const itemData = item.toObject();
--- a/module/data/action/damageAction.mjs
+++ b/module/data/action/damageAction.mjs
@@ -3,6 +3,32 @@ import DHBaseAction from './baseAction.mjs';
 export default class DHDamageAction extends DHBaseAction {
     static extraSchemas = [...super.extraSchemas, 'damage', 'target', 'effects'];
 
+    prepareData() {
+        super.prepareData();
+        
+        const parentBaseDamage = this.getParentHitPointDamage();
+        if (this.damage?.main && parentBaseDamage) {
+            if (this.damage.main.includeBase) {
+                if (!this.damage.main) {
+                    this.damage.main = parentBaseDamage;
+                } else {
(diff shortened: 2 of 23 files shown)


module/applications/sheets/api/actor-setting.mjs
```diff
@@ -89,7 +89,7 @@ export default class DHBaseActorSettings extends DHApplicationMixin(ActorSheetV2
     async _onDropItem(event, item) {
         if (item?.type === 'feature') {
             if (item.parent?.uuid === this.actor.uuid) {
-                return super._onDropItem(event);
+                return super._onDropItem(event, item);
```
CarlosFdez, line 92: oops 🤡

module/data/fields/action/damageField.mjs
```diff
@@ -297,6 +297,11 @@ export class DHActionDiceData extends foundry.abstract.DataModel {
         };
     }
 
+    getHasEmptyFormula() {
```
CarlosFdez: Looking at the implementation of getFormula(), I think you can make this a hasFormula getter. While you could make it faster by checking data (0 multiplier and 0 bonus = no formula), getFormula() is a very cheap function regardless.

module/data/action/damageAction.mjs
```diff
@@ -3,6 +3,32 @@ import DHBaseAction from './baseAction.mjs';
...
+        const parentBaseDamage = this.getParentHitPointDamage();
+        if (this.damage?.main && parentBaseDamage) {
+            if (this.damage.main.includeBase) {
+                if (!this.damage.main) {
+                    this.damage.main = parentBaseDamage;
+                } else {
+                    for (const type of parentBaseDamage.type) this.damage.main.type.add(type);
+
+                    const actionDamage = this.damage.main.value.getHasEmptyFormula() ? 
+                        null : this.damage.main.value.getFormula();
```
CarlosFdez: I've been considering if getFormula() should return null instead of "0". What are the repercussions of doing that?

The `hasFormula` getter change would be nice in the short term though. I'd accept it with that.

module/data/fields/action/damageField.mjs
WBHarry: Works for me! Moved to getter :+1:

module/data/action/damageAction.mjs
WBHarry: Hm..I feel this could cause crashes in various places where it's expected to be a parsable as a roll or other things.
It sounds like a good thing if that could be avoided, but I'd rather not touch it right now.

CarlosFdez: It'd need a || "0" all over the place where it expects a formula yeah. It'd need to be its own thing.

PR #2235, opened 2026-08-22
Simplify non-card tooltips

CarlosFdez:
This also adds support for tooltips in the action selection dialog.


module/documents/tooltipManager.mjs
```diff
@@ -142,28 +143,70 @@ export default class DhTooltipManager extends foundry.helpers.interaction.Toolti
...
+                tags.push(`${_loc('DAGGERHEART.GENERAL.Target.single')} ${_loc('DAGGERHEART.GENERAL.none')}`);
+            }
+        }
 
         const html = await foundry.applications.handlebars.renderTemplate(
-            `systems/daggerheart/templates/ui/tooltip/${type}.hbs`,
+            `systems/daggerheart/templates/ui/tooltip/basic.hbs`,
             {
-                item: item,
+                ...pick(item, ['img', 'name']),
```
WBHarry, line 209: Hm. Why make a new object with limited properties instead of sending in the full `item`? Are there sideeffects?
To note that this is done in a few places, incase it should be changed

module/documents/tooltipManager.mjs
```diff
@@ -179,15 +222,29 @@ export default class DhTooltipManager extends foundry.helpers.interaction.Toolti
         const actorUuid = element.dataset.tooltip.slice(8);
         const actor = await foundry.utils.fromUuid(actorUuid);
         const attack = actor.system.attack;
+        if (!attack) return null;
 
         const description = await foundry.applications.ux.TextEditor.enrichHTML(attack.description);
+        const trait = CONFIG.DH.ACTOR.abilities[attack.roll?.trait];
+        const range = CONFIG.DH.GENERAL.range[attack.range];
+
+        const typeTags = Array.from(attack.damage?.main?.type)
```
WBHarry:
```suggestion
        const typeTags = Array.from(attack.damage?.main?.type??[])
```
`Array.from` will crash if passed undefined

module/documents/tooltipManager.mjs
```diff
@@ -142,28 +143,70 @@ export default class DhTooltipManager extends foundry.helpers.interaction.Toolti
...
         if (item instanceof Item && item.system.metadata.isQuantifiable) {
             tags.unshift(`${_loc('DAGGERHEART.GENERAL.quantity')} ${item.system.quantity}`)
         }
+        if (item instanceof game.system.api.models.actions.actionsTypes.base) {
+            // todo: should these be in action._getTags()?
+            function safeEval(formula) {
+                try {
+                    if (isNaN(formula)) {
+                        const data = item.getRollData.bind(item)(),
+                            roll = new Roll(Roll.replaceFormulaData(formula, data)).evaluateSync();
```
WBHarry, line 167: Neat, I was about to comment that `evaluateSync` throws an error if there are non-deterministic terms in the formula and that'd somehow make this not work,
But yeah, in that case we -want- them to show up as unrolled dice terms `1d6` and whatnot, and that should work out with returning formula as is.

The only thing I'm unsure about is what happens if you have: `1d6 + @system.traits.agility.value`. Does the error prevent the trait value from being resolved? Will have to test

module/documents/tooltipManager.mjs
CarlosFdez: Sometimes an item isn't passed. Sometimes it's a move. Can even just be somethijg just passed in. Bolting stuff onto  item is dangerous, and splatting everything is excessive.

Kinda a consequence of generalizing.

module/documents/tooltipManager.mjs
CarlosFdez: Unsure. This is mostly copy pasta with a try catch on top

CarlosFdez:
I think I forgot to rename the tooltip css class. I'll do that post walk.

module/documents/tooltipManager.mjs
WBHarry: Considering this is only used for `cost.max` I think it's fine. It'd be nonsensical to have a non-deterministic term in a feature's static maximum :+1:

module/documents/tooltipManager.mjs
WBHarry: Mm, then we'll go with that. It just stood out as strange to me. :+1:

WBHarry (approved):
Nothing else stands out as questionable to me. Can be pulled in after name changes you wanted to make :+1:

PR #2238, opened 2026-08-23
Card cleanup, some tooltip/accessibility fixes, and card scaling

CarlosFdez:
Adds inert to the i tags, answers some of the feedback, and makes more things scalable.

Card size variability will be required for uses cases such as the character creator and possibly things like the compendium browser depending on how we're feeling. So far it seems to work pretty fine, though there might be edgecases.


WBHarry (approved):
:+1:

PR #2240, opened 2026-08-23
beastform fixes

Dapoulp:
FIXES:

1. Fix Beastform drop not working properly
2. Fix Beastform ading its features twice
3. Fix a bug when closing Beastform Choices Dialog


WBHarry (approved):
Awesome, works great :+1:

Good thing this was never in an actual release, since `grantedBy` has only been on main so far.

PR #2241, opened 2026-08-23
updated welcome journal with team info and link to FoundryVTT store

cptn-cosmo:
very simple journal update with team update and link to foundryvtt store

### DIFF
--- a/src/packs/journals/journal_Welcome___Information_g7NhKvwltwafmMyR.json
+++ b/src/packs/journals/journal_Welcome___Information_g7NhKvwltwafmMyR.json
@@ -15,7 +15,7 @@
       "image": {},
       "text": {
         "format": 1,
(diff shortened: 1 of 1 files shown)


CarlosFdez:
I feel like instead of mentioning official modules without context, it should probably start by saying what its for. A new user won't really know what it is. Maybe start the sentence with something akin to "For the complete experience" or something, since the intent of those modules is for that maybe?

cptn-cosmo:
Good point! I updated the journal accordingly :)

CarlosFdez:
> Additional Classes &amp; Subclasses, new Domains, Ancestries, Communities, Transformations, Adversaries, Environments and Campaign Frames

While MM is using the marketing phrase of the book, I think we don't really have to. These additional elements will be in the core system when it releases.

I'm also now observing typos in there we should clean up. "You refer back" should be something like "You can refer back"

WBHarry:
Hm...I don't think we need to throw in the whole sale's pitch in the Journal. I think that might make it seem as if we are selling it, or directly sponsoring it or whatnot.

I do think it's probably a good thing to -inform- that official modules exist, since that's good for people that don't engage with the community much.

-----
So my thought would be to remove the pitch, 
image

And do the text like so:
```
There are official modules available for the system that are currently on preorder.
You can find all official content modules in the [FoundryVTT Store](https://www.foundryvtt.store/creators/darrington-press). They also come bundled in the [Daggerheart Bundle](https://www.foundryvtt.store/bundles/daggerheart-bundle). 
```

Ikraik:
image
Dunno why but the core version degraded... is that normal?

CarlosFdez:
I'll take a crack at this later I suppose.

PR #2244, opened 2026-08-23
Fix visual issues with compendium browsers and other card work

CarlosFdez:
* Makes it so that wide images don't cause issues in the compendium browser
* Added tooltip support to compendium browser
* Refactor an issue with typography mixins, and use the correct one for cards
* Fix certain class tooltips
* Add card tooltip theming, default to "light"
* Remove hide attribution setting


WBHarry (approved):
Awesome. Looks great in the Compendium Browser :+1:

PR #2246, opened 2026-08-25
Add support for classes in embed table

CarlosFdez:
example: `@EmbedTable[rollTable:Compendium.daggerheart.rolltables.RollTable.S61Shlt2I5CbLRjz|classes:blue]` adds the blue css class. The intent is to enable more styling reuse in journals.


WBHarry (approved):
:+1:

PR #2249, opened 2026-08-25
Update attribution and clean data

CarlosFdez:
clean was for non-weapon/armor, since attempting to do a no diff non recursive update on weapons/armor makes the actions explode. I think it might be a consequence of how weapon/armor features are coded.


WBHarry (approved):
My computer is dying from the file amount, nuuuuh

CarlosFdez:
dunno, the big 🥚 works for me!

PR #2256, opened 2026-08-26
[Feature] Martial Stances

WBHarry:
Added all martial stances in a subfolder in the Subclasses compendium
**Subclasses --> Subclass Features --> Martial Stances**

- The current thing a user would have to do is to pull in the stances they select onto the character sheet. I added a text stating this in the `Stance Fighter` feature.
- All stances are currently set up with a `Enter Stance` action that applies a marker effect on self, so you keep track of which stance you are in.


CarlosFdez:
Not a bad idea. This can work for now.

CarlosFdez (approved):

PR #2257, opened 2026-08-26
[Feature] Dread Icons And Images

WBHarry:
Added the correct dread icon and domain image
image


CarlosFdez:
The SVG will need some adjustments. Let me make some changes.

CarlosFdez (approved):
Updated. Dread icon is a bit bigger now, which is closer to their relative sizing in the actual game.

PR #2258, opened 2026-08-26
[Feature] Multi Action

WBHarry:
Added a new action that holds many other actions grouped. The grouped actions do not display an Action button in the sheet, instead you use the Grouping action to access them.
Currently the action type is named `grouping` with a label of `Grouping`, but this should be changed. I'm not entirely sure to what. `Multi` has been suggested.

The action has two different modes:
**Randomized**
image
This handles `Battle Box`, `Gobstalker` and `Grand Demiurge Adonix`

**Selected**
image
This handles `Dragon Mother Mitera`.
Do we have a tooltip for actions? That'd level-up this dialog.

----
It makes a stark difference in the usability and niceness of the sheets.
**Before**
image
**After**
image


module/data/fields/action/groupedField.mjs
```diff
@@ -0,0 +1,69 @@
...
+            })
+        };
+        super(groupedFields, options, context);
+    }
+
+    static async execute(config) {
+        const groupedActions = 
+            Array.from(this.grouped.groupedActions).map(x => this.item.system.actions.get(x)).filter(Boolean);
+        if (!groupedActions.length) {
+            return ui.notifications.error(_loc('Failed'));
```
CarlosFdez: This isn't a localization string. Nor should it be, they need to be namespaced to avoid potential issues with double localizes.

templates/sheets/global/partials/inventory-item-V2.hbs
```diff
@@ -147,24 +147,31 @@ Parameters:
...
-                {{action.name}}
-            </button>
-            {{/if}}
-            {{#if action.uses.max}}
-                <button type="button" class="action-uses-button" data-action="increaseActionUses" data-item-uuid="{{action.uuid}}" {{#if ../inactive}}disabled{{/if}}>
-                {{action.remainingUses}}/{{action.uses.max}}
-                </button>
-            {{/if}}
-        </div>
+            {{#unless action.isGrouped}}
```
CarlosFdez, line 150: There's a quirk with this. It is possible for item-buttons to render with nothing if there are only grouped actions. Depending on the intent, you might have to add something to action collection to handle that.

CarlosFdez: Well, actually, that's unlikely to happen. I think a visible/hidden field might be more generic here but this is fine for now.

CarlosFdez:
There is room for improvement here but this is a pretty good foundation. Not bad.

CarlosFdez:
Might be worth making an issue, but my main thought is that it might be nice if there is a way to reuse an action's description inside the main action's description, to cut down on text duplication. Maybe some sort of special `@Lookup` parameter off of some extra action roll data, if that's doable. However that may lead to double nesting `<p>` tags, which leads to an empty `<p></p>`, which will cause issues with current list styling. So it might not be doable cleanly.

PR #2259, opened 2026-08-26
[Feature] Two Column CreateAction Dialog

WBHarry:
Changed so the CreateAction Dialog has two columns for the actions, it was becoming too long.
image


WBHarry:
This makes me ponder if we should do alphabetic order maybe? Atm it's not really ordered by either alphabet or "most-commonly-picked"

WBHarry:
Yeah, I changed it to be alphabetical
image

PR #2260, opened 2026-08-26
[Fix] Remove dud migration

CarlosFdez:
As the dud migration edits the items, it likely hampers the function of the new one. I didn't test on an old world, but even if that's not true, might as well still clean it up.


WBHarry (approved):
:+1:

PR #2265, opened 2026-08-27
Start work on adversary/environment embeds

CarlosFdez:
The actual end result is not pretty. It is there so that something renders for the ongoing premium project in the adversaries and environments pages. If possible, please review the general structure. The actual appearance will be done later.


WBHarry (approved):
:+1:

PR #2271, opened 2026-08-28
Fix armor modifying features and update runeshields

CarlosFdez:
Because of default values, double duty doesn't have any forward facing fixes. Primarily this fixes protective and barrier.


WBHarry (approved):
:+1: I forgot about this for 2.8.0

PR #2272, opened 2026-08-28
Further improve styling for actor embeds

CarlosFdez:
Almost done, only thing really missing now are the potential adversaries and then bugfixes for when stuff comes up.

image


WBHarry (approved):
:+1:

PR #2276, opened 2026-08-28
[Feature] Feature Resources

WBHarry:
Added so Feature items can declare ActorResources.
The resource data is stored on the Feature itself and updates to the resource occur on the Feature. They are added in-memory to the character however so all resources are available from `actor.system.resources`.

Added so that the Warlock feature Favor and the Martial Artist feature Stance Fighter set up resources.


WBHarry:
One thing is bothering me though that I'm not sure how to solve well.

The followup goal I have after this is to make Custom Resources available to be spent as Costs by Actions.

That's hard to make happen like this, since the Feature items hold the costs, and they might be in compendiums, making it impossible to get a complete list of available Custom Resources without loading every compendium document.

WBHarry:
Moving this to draft as I think I'll remake this some to handle the above

WBHarry:
Changed to define optional resources in ResourceConfig. Features can pick from them.
The reason for this change is that if Actions are going to be able to select these resources as Costs - then we have to be able to consistently grab a list of the available Resources.

The previous solution where the definition was -in- the Features wouldn't work well, since the features can be in compendiums, forcing us to load all compendiums to get the whole list.

This solution will let us join up: 
`CONFIG.DH.RESOURCE.<actorType>.all + CONFIG.DH.RESOURCE.optionalResources + <Homebrew Resources With some later modifications>`
This final list is what actions will pick costs from without relying on reading items.

CarlosFdez:
An input with autocomplete would still worked for the old solution, though it wouldn't have felt as seamless I suppose. This is fine though.

I haven't checked the code yet, but what is the possibility of adding an optional boolean prop on the resource collection? That would allow homebrew resources to also become optional.

WBHarry:
> An input with autocomplete would still worked for the old solution, though it wouldn't have felt as seamless I suppose. This is fine though.

Would it? 
- Say that we we defined the `Favor` resource on the Favor subclass card from Warlock, 
- This card exists in the SRD compendium
- Actions on unrelated Features want to spend this resource
- If the card hasn't been loaded into the world, then there's no way for the action to list the resource as a cost option, since it doesn't exist.

It'd work fine once the card is on a character. The issue I perceived was when making new items and wanting to access costs.

CarlosFdez:
if the action cost was instead a text input with autocomplete, it not existing is not a problem (for it working).

WBHarry:
Well..if we consider typing unknown resources by hand to be user friendly I suppose :thinking:

CarlosFdez:
We do not. That's why I said not seamless :P 

Its also not a major problem to get all possible resource types by grabbing the whole compendium in the sheet.

CarlosFdez:
That said what you did is fine, especially if we also support module flags for it.

WBHarry:
Well, I'll leave you to consider the PR and the solution in full at your time. If you think it was better previously, or you see some  other way to go, then we can consider that 😄

CarlosFdez:
Its fine. I was refuting that it wouldn't work, not that it was good/bad.

If you can make it so optional-ness is a field on the resource, that might be better than using separate collections.

WBHarry:
Hm. Which collection would you place these in?

CarlosFdez:
None, it can be a field on the resource config entry itself. That would let homebrew resources be optional as well.

WBHarry:
Ah, that I didn't consider. I'll mess with that some

WBHarry:
Hm, the akward thing with bringing it into the existing `ResourceConfig` objects is that it forces it to be in an actor specific type.

When you're setting up an individual Feature, there is no information about what sort of actor it goes on.
We can of course grab a Set of all the resources from all the actor types, but dunno
image

templates/sheets/items/feature/settings.hbs
```diff
@@ -9,7 +9,32 @@
...
         </span>
+        <span>{{localize "DAGGERHEART.ITEMS.Feature.actorResources"}}</span>
+        {{formInput document.system.schema.fields.actorResources value=document.system.actorResources localize=true}}
     </fieldset>
 
     {{> "systems/daggerheart/templates/sheets/global/partials/resource-section/resource-section.hbs" }}
+
+    <fieldset>
+        <legend>
+            {{localize "Actor Resources"}}
```
CarlosFdez: Incorrect localization key

CarlosFdez:
> Hm, the akward thing with bringing it into the existing `ResourceConfig` objects is that it forces it to be in an actor specific type.

This is a valid concern. But its not really specific to optional resources. We're only getting around it now by hardcoding I think, so it might be the right play to grab all the types regardless.

If you want to keep it separate, one idea then is to keep it as a separate collection, but add `optional: true` when merging into the hardcoded list. Then it can be something for homebrew ones, which may want the same optional handling.

EDIT: If you've already solved it disregard, I haven't checked.

module/data/item/feature.mjs
```diff
@@ -39,7 +39,15 @@ export default class DHFeature extends BaseDataItem {
                 initial: 'passive',
                 choices: CONFIG.DH.ITEM.featureForm,
                 label: 'DAGGERHEART.CONFIG.FeatureForm.label'
-            })
+            }),
+            actorResources: new fields.TypedObjectField(new fields.SchemaField({
+                value: new fields.NumberField({
```
CarlosFdez: Is this value the current value?

module/applications/sheets/actors/character.mjs
```diff
@@ -1148,12 +1149,18 @@ export default class CharacterSheet extends DHBaseActorSheet {
...
-        const decreasing = inputValue <= this.document.system.resources[resource].value;
+        const decreasing = inputValue <= resource.value;
         const value = decreasing ? inputValue - 1 : inputValue;
-        await this.document.update({ [`system.resources.${resource}.value`]: value }, { render: false });
+        if (resource.featureParentId) {
+            const feature = this.document.items.get(resource.featureParentId);
+            await feature.update({ [`system.actorResources.${resourceKey}.value`]: value });
+        } else {
+            await this.document.update({ [`system.resources.${resourceKey}.value`]: value }, { render: false });
+        }
```
CarlosFdez: if resource values are gonna be saved on the source item rather than the actor resources, what you should probably do is move all of this to a function like `this.document.system.updateResource(key, value)`. This should keep everything more in sync and handle what happens if multiple features all set the same resource (you'd need to update all of them).

WBHarry: Good idea, definately makes it a lot more clear and future proof

module/data/item/feature.mjs
WBHarry: Yes. The icons/label/max etc are in the config and are fetched from there.
`Value` is seeded with `config.<resource>.initial` when it's added to a feature. 

I made it as a schema because I don't feel confident there won't be something more that needs to be added here.

module/applications/sheets/actors/character.mjs
CarlosFdez: hmmm, well, we'll see after. Its a step in the right direction but I just remembered we have a resource updater thing, and we may want to centralize all of that somewhere. Perhaps actor.system.updateResource() calls the resource updater, or perhaps the resource updater calls the former. Maybe it should support either key/value, or an object of everything to update.

CarlosFdez: I think it should still be done as its an iterative step that'll make it easier to suss out in its totality but if you have an idea of how to skip a few more steps forward, let me know. It looks like takeHealing() is fundamentally similar to this and should probably *also* redirect to the item.

module/data/item/feature.mjs
CarlosFdez: actorResources should probably be a comment, it wasn't intuitve at first glance that actorResources wasn't a definition of what resources to store, but rather a backing storage of what the values are to avoid putting it in actor.system.resources.

I think putting the values in actor.system.resources and having a check when items are deleted to clean it up might be much easier, but it is nice that the values clean up without special handling when the items are deleted at least.

WBHarry:
Rework done.

I didn't add so Custom Resources work for Damage/Healing because we need to get in:
https://github.com/Foundryborne/daggerheart/pull/2285
first as it migrates from `WeaponResource` to `Resource`

module/applications/sheets/actors/character.mjs
```diff
@@ -1148,12 +1145,18 @@ export default class CharacterSheet extends DHBaseActorSheet {
...
         const inputValue = Number.parseInt(textValue);
-        const decreasing = inputValue <= this.document.system.resources[resource].value;
+        const decreasing = inputValue <= resource.value;
         const value = decreasing ? inputValue - 1 : inputValue;
-        await this.document.update({ [`system.resources.${resource}.value`]: value }, { render: false });
+        if (resource.featureParentId) {
+            const feature = this.document.items.get(resource.featureParentId);
+            await feature.update({ [`system.actorResources.${resourceKey}.value`]: value });
+        } else {
+            await this.document.update({ [`system.resources.${resourceKey}.value`]: value }, { render: false });
```
CarlosFdez: Where is the followup thing that will actually trigger the render?

module/config/resourceConfig.mjs
```diff
@@ -63,6 +65,33 @@ const companionBaseResources = Object.freeze({
     }
 });
 
+export const optionalResources = Object.freeze({
```
CarlosFdez: We may not want to freeze this, or at least we may want to provide module flags (similar to pf2e's module flags) to extend this. Updating these is vastly superior for a module compared to updating homebrew settings, which don't clean up when the module is disabled.

module/data/actor/creature.mjs
```diff
@@ -27,6 +27,33 @@ export default class DhCreature extends BaseDataActor {
         return !vulnerableAppliedByOther;
     }
 
+    get availableExtraResources() {
+        const homebrewResources = 
+            game.settings.get(CONFIG.DH.id, CONFIG.DH.SETTINGS.gameSettings.Homebrew).toObject();
+        const applicableHomebrewResources = homebrewResources.resources[this.metadata.type]?.resources ?? {};
+        
+        return Object.entries(this.resources).reduce((acc, [key, data]) => {
```
CarlosFdez: We may want to replicate this function in our functional helpers, which would do this so much nicer:

Or its sister pickBy.

https://es-toolkit.dev/reference/object/omitBy.html#omitby

module/data/actor/creature.mjs
```diff
@@ -60,4 +87,60 @@ export default class DhCreature extends BaseDataActor {
             }
         }
     }
-}
+
+    // _preUpdateOperation() {
+
+    // }
```
CarlosFdez:
```suggestion
```

👀

WBHarry: You saw naathing.

module/applications/sheets/actors/character.mjs
WBHarry: Because the menu is a tooltip, we don't actually rerender. I just modify the DOM.
image

WBHarry: I did however notice that I still had the split update thing with `feature` still there. Removed it since it's no longer used 🪂

module/config/resourceConfig.mjs
WBHarry: Good point. Unfrozen 🔥

CarlosFdez:
Did the changes I was looking to do. Seems fine, but give it a second comb over.

WBHarry:
Huh, I had no idea `actor.itemTypes` was a thing. That's really neat.

module/documents/actor.mjs
```diff
@@ -218,78 +217,58 @@ export default class DhpActor extends Actor {
...
-                    if (item && item.type == 'feature') {
-                        acc.push(...item.system.actorResources);
-                    }
-
-                    return acc;
-                }, []);
-
-                this.cleanupOptionalResources(ids, possibleRemovedResources);
-            }
+            this._cleanupOptionalResources();
```
WBHarry: Has to be gated behind `if(game.user.id === userId)` so the cleanup isn't run for every connected user. 
Users without ownership of the actor would be throwing an error.

module/documents/actor.mjs
```diff
@@ -218,78 +217,58 @@ export default class DhpActor extends Actor {
...
-                            !(change.system?.actorResources ?? []).includes(x));
-                        acc.push(...possibleRemoved);
-                    }
-
-                    return acc;
-                }, []);
-
-                this.cleanupOptionalResources(features.map(x => x.id), possibleRemovedResources);
-            }
+            this._cleanupOptionalResources();
```
WBHarry: Has to be gated behind `if(game.user.id === userId)` so the cleanup isn't run for every connected user. 
Users without ownership of the actor would be throwing an error.

WBHarry (commented):
You did an amazing job with this 🥳 
- It's a lot cleaner now.
- The resource is immediately present on the actor instead of on first update.
- We clean up homebrew resources aswell.

The only thing I found after testing everything was that we need to guard by user since the cleanup is now run on `onDelete/onUpdate` which runs on each machine.

WBHarry:
I added in the user gating. Pulling this in.

CarlosFdez:
Good catch, I forgot that on callbacks have that issue.

PR #2280, opened 2026-08-28
Convert inline roll links to global listeners

CarlosFdez:
This improves support for module usage such as quest log modules.


WBHarry (approved):
Interesting! I wouldn't have thought `openDetachedWindow` would work for chat messages :+1:

CarlosFdez:
> Interesting! I wouldn't have thought `openDetachedWindow` would work for chat messages 👍

Its more that its applied to the window, and relies on events bubbling up to the window. Since its on the window, it also has to work for popout windows.

This enables it to work on things like quest log modules.

PR #2283, opened 2026-08-29
[Fix] Healing Improvements

WBHarry:
- Unlocked the hidden Full Restore option for healing actions so any resource can use that checkbox.
- Fixed the `Clear` logic in `Actor.modifyResource`, it was always ending up picking 0 to restore to. Now it correctly sets `isReverse` resources to 0 and "normal" resources to their max. So `Full Restore` on hope gets you to 6 hope for instance.
- Updated relevant SRD entries to make use of it.
- Fixed so chat messages don't just show 0 as healing value when FullRestore is true.

image

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -190,7 +190,8 @@
                 },
                 "fullRestore": {
                     "label": "Fully restore the resource.",
-                    "inChatRoll": "Full"
+                    "inChatRoll": "Full",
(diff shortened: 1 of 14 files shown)


CarlosFdez (approved):
Looks good to me, but another in the "value and valueAlt should be optional" box 🙃

PR #2285, opened 2026-08-29
[Feature] Heal Feature Resource

### LINKED ISSUE #982: [BUG] - Unleash Chaos resource not refillable
**Describe the bug**
'Unleash Chaos' resource is at value 0 by default, and stays at 0 after using the 'Replenish Token' action

**To Reproduce**
1. Have a character with the 'Unleash Chaos' domain card from Arcana domain, level 1
2. Click on 'Unleash Chaos'
3. Click on 'Cast': resource is not scalable for its value is 0
4. Click on 'Unleash Chaos'
5. Click on 'Replenish Token': resource is not scalable for it is 0

**Expected behavior**
- Default resource value is scaled at @cast
- 'Replenish Token' set resource value to @cast

**Setup Information:**
 - OS: Windows
- Browser: Chrome
- Foundry Version: v13 b347
- System Version: 1.0.5

WBHarry:
Fixes #982 
- Added so heal actions can restore the feature's resource.
- Swapped out the existing `weaponResource` config key for the generic `resource`.
- Updated SRD actions that should restore their feature's resource. There were more than I thought.


module/data/action/baseAction.mjs
```diff
@@ -501,6 +501,13 @@ export default class DHBaseAction extends ActionMixin(foundry.abstract.DataModel
 
             delete source.damage.parts;
         }
+
+        if (source.damage.resources.weaponResource) {
+            // source.damage.resources.resource = source.damage.resources.weaponResource;
+            // source.damage.resources.resource.applyTo = CONFIG.DH.GENERAL.healingTypes.resource.id;
+            // source.damage.resources.resource.itemId = ;
+            delete source.damage.resources.weaponResource;
```
WBHarry: WeaponResource needs to be migrated.

I was doing it here, but then I realised I can't access `itemId`, IE the ID of the feature that contains the action in `baseAction.migratedata`.

----
- We can either just do what I did here and delete any such instances with the justification that there are disappearingly few people that would have made use of it for some single weapon, and they can just correct such an action.
- Or we have to set up a migrateData on the document class `documents/Item` looping over all actions.

WBHarry:
I'm considering removing `itemId` from what I did, and instead just assuming that if you are healing a Feature's Resource (not related to actor resource - this is like the tokens on Unleash Chaos), it is -always- the action's Item.

I guess keeping it as `itemId` is more future proof as if we ever do make actions able to restore resource charges on other features, it'd assumably need that

CarlosFdez:
rebase/merge main when you can, and I"ll start looking into it and you said at my leisure. This will be the thing I dive into as I allocate review time.

WBHarry:
Merged in main :+1:

module/applications/sheets-configs/action-base-config.mjs
```diff
@@ -115,7 +115,7 @@ export default class DHActionBaseConfig extends DaggerheartSheet(ApplicationV2)
         }
     };
 
-    static CLEAN_ARRAYS = ['cost', 'effects', 'summon'];
+    static CLEAN_ARRAYS = ['cost', 'effects', 'summon', 'damage.resources'];
```
CarlosFdez: Thoughts on moving this to _prepareSubmitData()? This being here actually confused me a bit. If you'd rather keep it here, can you add a comment saying what this is for?

I'm surprised foundry isn't cleaning arrays with data models.

module/data/action/baseAction.mjs
CarlosFdez: Wouldn't doing the first blow up reloading weapons?

I think migrateData() on the item is fine.

WBHarry: I added the `documents/item.migrateData` :+1:

module/applications/sheets-configs/action-base-config.mjs
WBHarry: Moved the array to where it's used in `_prepareSubmitData` :+1:

PR #2287, opened 2026-08-29
[Feature] RollResolver Duality Labels

WBHarry:
Fixed so that the RollResolver dialog when using manual dice rolling will display Hope Roll and Fear Roll as labels when applicable.
image

I remade the previous PR that was delayed to instead use a hook. This is much better since Duality rolls can now be made anywhere using roll syntax (`/r 1d12h + 1d12f`)


CarlosFdez (approved):
Looks fine at a glance.

PR #2290, opened 2026-08-30
Add potential adversaries to actor embeds

CarlosFdez:
This makes the current adversary groups show in the embed, however it will not display hope/fear ones correctly. Without a change in our structure though, that is an impossibility.


WBHarry:
I've been holding off with this one because I've been messing with Environments back and forth 🙃

CarlosFdez:
I figured it was that. No rush.

WBHarry (approved):
:suspect:

PR #2291, opened 2026-08-30
Replace button enrichers with the content link styling

CarlosFdez:
Replaces the yellow buttons that look awkward mid paragraph and mess up line height with the content-link styling, which is used for the regular rolling and embeds.

image

The reason for create element refactor was because having spaces around the inner text causes some strange-ness in inline text elements where the button will be right up alongside the text. This seemed better than lines in our big string, and safer too.


WBHarry (approved):
Awesome :+1:

PR #2292, opened 2026-08-30
[Feature] Implement refresh from compendium

### LINKED ISSUE #913: [Feature] Refreshing updated compendium items
As compendium items are updated, the players/GM should know that a new version is available. I'm thinking something similar to how the CPR module has a way of telling you which items you have are not updated via button on top of the sheet, or I hear PF2e has something similar. This should probably include an option to ignore changes to a certain item (if they have manually modified them and set them to an ignore list)

CarlosFdez:
Closes https://github.com/Foundryborne/daggerheart/issues/913

This adds a button in the window header menu for adversaries/environments/most item types. On an adversary/environment actor, refresh an item instead checks the actor's copy of that item, rather than assuming that the adversary feature has some source feature to pull from.

I have some uncertainties about it.
* This is an inherently dangerous operation. We may want to further restrict it early on
* A lot of it is document level, not system level. We should maybe defer some of its operation to the system level to avoid some checks, but I'm not sure which parts. Maybe we put stuff in metadata.
* How do we handle mixed ancestries? We really shouldn't refresh those.
* Should we handle changes in features in ancestry/community/subclasses? In what ways? I haven't touched feature

molilo was uncertain about calling it refresh, if others share the same opinion, we can rename it.

### DIFF
--- a/module/applications/sheets/api/base-actor.mjs
+++ b/module/applications/sheets/api/base-actor.mjs
@@ -25,7 +25,8 @@ export default class DHBaseActorSheet extends DHApplicationMixin(ActorSheetV2) {
             openSettings: DHBaseActorSheet.#openSettings,
             sendExpToChat: DHBaseActorSheet.#sendExpToChat,
             increaseActionUses: event => DHBaseActorSheet.#modifyActionUses(event, true),
-            groupActionSelect: DHBaseActorSheet.#groupActionSelect
+            groupActionSelect: DHBaseActorSheet.#groupActionSelect,
+            refreshFromCompendium: DHBaseActorSheet.#onRefreshFromCompendium
         },
         contextMenus: [
             {
@@ -60,6 +61,20 @@ export default class DHBaseActorSheet extends DHApplicationMixin(ActorSheetV2) {
         const viewPermission = this.document.testUserPermission(game.user, this.options.viewPermission);
         const limitedOnly = this.document.testUserPermission(game.user, this.options.viewPermission, { exact: true });
         return limitedOnly ? this.document.system.metadata.hasLimitedView : viewPermission;
+    }
+
+    /** @inheritdoc */
+    _getHeaderControls() {
+        const controls = super._getHeaderControls();
+        if (this.actor.refreshSourceUuid) {
+            controls.push({
+                label: _loc('DAGGERHEART.ITEMS.Base.Refresh.Title'),
+                icon: 'fa-solid fa-arrow-rotate-left',
+                action: 'refreshFromCompendium'
+            });
+        }
+
+        return controls;
     }
 
     /* -------------------------------------------- */
@@ -304,6 +319,19 @@ export default class DHBaseActorSheet extends DHApplicationMixin(ActorSheetV2) {
     static async #groupActionSelect(event, button) {
         const action = await fromUuid(button.dataset.itemUuid);
         action.use(event, { groupAction: { forceSelect: true }});
+    }
+
+    /** @this DHBaseActorSheet */
+    static async #onRefreshFromCompendium() {
+        const refresh = await foundry.applications.api.DialogV2.confirm({
+            window: {
+                title: _loc('DAGGERHEART.ITEMS.Base.Refresh.Title')
+            },
+            content: _loc('DAGGERHEART.ITEMS.Base.Refresh.AreYouSure')
+        });
+        if (refresh) {
+            this.document.refreshFromCompendium();
+        }
     }
 
     /* -------------------------------------------- */
--- a/module/documents/actor.mjs
+++ b/module/documents/actor.mjs
@@ -1205,7 +1205,7 @@ export default class DhpActor extends Actor {
      */
     async refreshFromCompendium({ save = true } = {}) {
         const latest = await fromUuid(this.refreshSourceUuid);
-        if (!latest || !this._stats.compendiumSource?.startsWith('Compendium.')) {
+        if (!latest) {
             return ui.notifications.error(_loc('DAGGERHEART.ITEMS.Base.Refresh.Error.doesNotExist'));
         }
         if (latest.type !== this.type) {
--- a/module/documents/actor.mjs
+++ b/module/documents/actor.mjs
@@ -1216,9 +1216,6 @@ export default class DhpActor extends Actor {
             return ui.notifications.error(_loc('DAGGERHEART.ITEMS.Base.Refresh.Error.invalidTier'));
         }
 
-        // @todo defer preserved properties to system data somehow
-        // should gm notes be preserved or ko'd?
-        // what do we do about flags?
         const currentSource = this.toObject(true);
         const latestSource = latest.toObject(true);
         const system = foundry.utils.mergeObject(latestSource.system, {
--- a/module/documents/actor.mjs
+++ b/module/documents/actor.mjs
@@ -1,7 +1,7 @@
 import { emitGMUpdate, GMUpdateEvent } from '../systemRegistration/socket.mjs';
 import { LevelOptionType } from '../data/levelTier.mjs';
 import DHFeature from '../data/item/feature.mjs';
-import { createScrollText, damageKeyToNumber, getDamageKey, createShallowProxy } from '../helpers/utils.mjs';
+import { createScrollText, damageKeyToNumber, getDamageKey, createShallowProxy, pick } from '../helpers/utils.mjs';
 import DhCompanionLevelUp from '../applications/levelup/companionLevelup.mjs';
 import { ResourceUpdateMap } from '../data/action/baseAction.mjs';
 import { abilities } from '../config/actorConfig.mjs';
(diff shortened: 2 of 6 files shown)


module/documents/actor.mjs
```diff
@@ -1186,7 +1198,117 @@ export default class DhpActor extends Actor {
...
-    applyActiveEffects(phase) {
-        super.applyActiveEffects(phase);
+    /** 
+     * Refreshes this actor's data, effects, and items using information from the compendium.
+     * @param {options} [options]
+     * @param {boolean} [options.save] if set to false, returns the batch data to perform the operation instead of doing it
+     */
+    async refreshFromCompendium({ save = true } = {}) {
+        const latest = await fromUuid(this.refreshSourceUuid);
+        if (!latest || !this._stats.compendiumSource?.startsWith('Compendium.')) {
```
WBHarry: Shouldn't need the `!this._stats.compendiumSource?.startsWith('Compendium.')` check since `rerfreshSourceUuid` always returns null in that case.

module/documents/actor.mjs
```diff
@@ -1186,7 +1198,117 @@ export default class DhpActor extends Actor {
...
+        }
+
+        // @todo defer preserved properties to system data somehow
+        // should gm notes be preserved or ko'd?
+        // what do we do about flags?
+        const currentSource = this.toObject(true);
+        const latestSource = latest.toObject(true);
+        const system = foundry.utils.mergeObject(latestSource.system, {
+            notes: currentSource.system.notes || latestSource.system.notes
+        });
```
WBHarry, line 1223: I was about to write that it might be more correct to just use the `latestSource` straight up - but I suppose this way is needed to maintain various links.

module/applications/sheets/api/base-actor.mjs
```diff
@@ -62,6 +63,20 @@ export default class DHBaseActorSheet extends DHApplicationMixin(ActorSheetV2) {
...
+    /** @inheritdoc */
+    _getHeaderControls() {
+        const controls = super._getHeaderControls();
+        if (this.actor.refreshSourceUuid) {
+            controls.push({
+                label: _loc('DAGGERHEART.ITEMS.Base.Refresh.Title'),
+                icon: 'fa-solid fa-arrow-rotate-left',
+                action: 'refreshFromCompendium'
+            });
+        }
```
WBHarry, line 75: Should we excempt `Characters` from having this functionality?

I suppose it -could- be used for pregens in a compendium or some such, but overall it doesn't seem very useful in that situation.

module/documents/actor.mjs
```diff
@@ -1186,7 +1198,117 @@ export default class DhpActor extends Actor {
...
+            }
+        }
+
+        // Hnadle Items
+        const itemsToDelete = this.items.filter(e => !latest.items.has(e.id)).map(i => i.id);
+        const itemCreates = latestSource.items.filter(i => !this.items.has(i._id));
+        const batchFromItems = (await Promise.all(
+            this.items
+                .filter(i => latest.items.has(i.id))
+                .map(i => i.refreshFromCompendium({ save: false, latest: latest.items.get(i.id) }))
```
WBHarry, line 1244: Very nice. Clean how the `save: false` option comes in here. 🎊

module/documents/item.mjs
```diff
@@ -16,6 +16,52 @@ export default class DHItem extends foundry.documents.Item {
...
 
+    /**
+     * Returns the uuid of the item that is used for refreshing.
+     * This isn't necessarily the sourceUuid. Compendium items don't have a refresh source.
+     * @returns {string | null} the uuid to refresh from, or null if it can't be refreshed
+     */
+    get refreshSourceUuid() {
+        // no refreshing compendium items
+        // ancestry items also aren't refreshable
+        if (this.pack || this.type === 'ancestry') return null;
```
WBHarry, line 27: We should probably return a specific error message here so we don't end up with a general one that confuses the situation.

module/documents/item.mjs
```diff
@@ -317,4 +343,96 @@ export default class DHItem extends foundry.documents.Item {
...
     }
+
+    /** 
+     * Refreshes this item's data and effects using information from the compendium.
+     * @param {options} [options]
+     * @param {boolean} [options.save] if set to false, returns the batch data to perform the operation instead of doing it
+     */
+    async refreshFromCompendium({ save = true, latest } = {}) {
+        latest ??= await fromUuid(this.refreshSourceUuid);
+        if (!latest || !latest.pack) {
```
WBHarry: `!latest.pack` shouldn't be neccessary since `this.refreshSourceUuid` only returns a UUID if it's in a compendium.

module/documents/actor.mjs
```diff
@@ -1186,7 +1198,117 @@ export default class DhpActor extends Actor {
...
+            return ui.notifications.error(_loc('DAGGERHEART.ITEMS.Base.Refresh.Error.invalidType'));
+        }
+        if (latest.system.tier !== this.system.tier) {
+            // An adversary that has been re-tiered is not eligible for refresh
+            return ui.notifications.error(_loc('DAGGERHEART.ITEMS.Base.Refresh.Error.invalidTier'));
+        }
+
+        // @todo defer preserved properties to system data somehow
+        // should gm notes be preserved or ko'd?
+        // what do we do about flags?
```
WBHarry: If we want to defer to system data we can do it one of two ways.
1) Just move it to the base system datamodel for the type (IE, base-actor/base-item) to get the logic out of here.
2) Make a getter/function on each data model with that model's specific preserved properties.

I'm kind of leaning on 1 in this case actually. I think it'd just make it harder to grasp if it was spread out.

----
Do we view GM notes as the GM's notes to self? In that case I imagine they get preserved.

----
Probably safest to keep flags since they can relate to all manner of random functionality?

module/documents/actor.mjs
CarlosFdez: Thanks for catching, yeah I forgot to remove it from actor.

module/documents/actor.mjs
CarlosFdez: Yeah, while actor has less of it, if we simply just overwrite everything, things break. The major risk of the refresh feature, and the reason why it might be good to restrict what can be refreshed maybe, is because we might miss some spots.

I haven't overwrote flags either, I don't know if we generally use them.

module/applications/sheets/api/base-actor.mjs
CarlosFdez: Characters are already exempted by virtue of not returning a refreshSourceUuid.

module/documents/item.mjs
CarlosFdez: In this situation, the refresh button shouldn't even show up at all on the window, so the error of it being missing only happens if you invoke it programmatically.

WBHarry: Fair enough in that case :+1:

module/documents/actor.mjs
CarlosFdez: Even then the handling of the effects and the actual update aren't really a polymorphic operation that benefit from subclassing if we move *all* of it, though a specific item type could consider perhaps extending what gets updated (for example, if we refreshing a class should also refresh the features). But that involves spreading it out.

The parts that may differ by item type are:
1) What should we preserve? I kinda like it all being centralized for now, but I could see a case of adding specific metadata to the schema to mark as not refreshed
2) Should anything else go along for the ride?

module/documents/item.mjs
```diff
@@ -16,6 +16,52 @@ export default class DHItem extends foundry.documents.Item {
...
+     * This isn't necessarily the sourceUuid. Compendium items don't have a refresh source.
+     * @returns {string | null} the uuid to refresh from, or null if it can't be refreshed
+     */
+    get refreshSourceUuid() {
+        // no refreshing compendium items
+        // ancestry items also aren't refreshable
+        if (this.pack || this.type === 'ancestry') return null;
+
+        const actor = this.actor;
+        if (['adversary', 'environment'].includes(actor.type)) {
```
WBHarry:
```suggestion
        if (['adversary', 'environment'].includes(actor?.type)) {
```
Was testing, and noticed items not on an actor would crash because of this 🙃

WBHarry:
Otherwise it seems to work great. It's approved for me, baring any structural changes you want to/don't want to make with the `@todo`s you left

module/documents/actor.mjs
WBHarry: I'm not sure if it's right or not to update everything related to a class if you update it. I'd probably argue no, but by no means locked on that.

I'm happy enough with everything being in one place unless we start bringing in big logic like the above that differs.

CarlosFdez:
Changes done, I'll keep as it is for now. Approve and merge whenever (I can't approve it myself lmao).

We can do ancestries if we can add identifying information that it is a mixed ancestry.

WBHarry (approved):
:+1:

PR #2295, opened 2026-08-30
[Fix] Downtime Cleanup

WBHarry:
- Fixed downtime messages in chat having left padding. The container was randomly a `ul` element, which we've added a global rule with padding for.
- Fixed so that non-GMs do not see the Short Rest and Long Rest buttons in the Party Sheet as they're used for the GM to call downtime.

### DIFF
--- a/templates/sheets/actors/party/party-members.hbs
+++ b/templates/sheets/actors/party/party-members.hbs
@@ -14,7 +14,7 @@
             <span>{{localize "DAGGERHEART.APPLICATIONS.GroupRollSelect.title"}}</span>
         </button>
         {{#if isGM}}
-            <button data-action="triggerRest"  data-action="triggerRest" data-type="shortRest">
+            <button data-action="triggerRest" data-type="shortRest">
                 <i class="fa-solid fa-utensils"></i>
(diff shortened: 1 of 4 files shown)


templates/sheets/actors/party/party-members.hbs
```diff
@@ -13,14 +13,16 @@
...
-        <button data-action="triggerRest"  data-action="triggerRest" data-type="shortRest">
-            <i class="fa-solid fa-utensils"></i>
-            <span>{{localize "DAGGERHEART.APPLICATIONS.Downtime.shortRest.title"}}</span>
-        </button>
-        <button data-action="triggerRest"  data-type="longRest">
-            <i class="fa-solid fa-bed"></i>
-            <span>{{localize "DAGGERHEART.APPLICATIONS.Downtime.longRest.title"}}</span>
-        </button>
+        {{#if isGM}}
+            <button data-action="triggerRest"  data-action="triggerRest" data-type="shortRest">
```
CarlosFdez:
```suggestion
            <button data-action="triggerRest" data-type="shortRest">
```

CarlosFdez (approved):

CarlosFdez:
Thanks for fixing it!

PR #2296, opened 2026-08-31
[Feature] clickable character sheet feature group headers

CarlosFdez:
Sheet modules might have to update after. This makes it so that the transformation is actually accessible.

Later on we can make subclass clickable, but for that we may want tooltips on each available rank.

image


WBHarry (approved):
Nice! You finally got rid of your arch nemesis, the sheetlist getter 😁

PR #2300, opened 2026-08-31
[Feature] Actor Sheets Image Popout

WBHarry:
Added image popout for all actor sheets on rightclick of the portrait and in the context menu.
image

image

### DIFF
--- a/module/applications/sheets/api/base-actor.mjs
+++ b/module/applications/sheets/api/base-actor.mjs
@@ -66,6 +66,12 @@ export default class DHBaseActorSheet extends DHApplicationMixin(ActorSheetV2) {
     /** @inheritdoc */
     _getHeaderControls() {
         const controls = super._getHeaderControls();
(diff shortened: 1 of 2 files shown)


CarlosFdez (approved):
Untested but looks good to me. Simple enough that I doubt you missed anything.

WBHarry:
Yah, it's simple business here 👨‍🍳

CarlosFdez:
I realized controls could be in DEFAULT_OPTIONS given they're not conditional but that's a nothingburger.

WBHarry:
Hm, I hadn't actually considered that. But in this case in particular I think I actually prefer this. Since the function is needed anyway due to the conditional already there, I think it makes sense not to split it

CarlosFdez:
Other sheets do use the controls in DEFAULT_OPTIONS though, that's what character does.

I don't mind too much.

PR #2301, opened 2026-08-31
[Fix] Evolved Beastform Damage Change

WBHarry:
Fixed so evolving beastforms modify damage correctly.

Since beastforms are always using custom damage formulas now, the old evolution damage modification solution wasn't working. Changing things to terms and building up rolls made the `@prof` in the formula get evaluated to 1 all the time.


CarlosFdez (approved):
works for me!

PR #2302, opened 2026-08-31
[Fix] error when preparing already prepared data

CarlosFdez:
Putting a non configurable property again can cause errors. Maybe new 

It only happens in certain API usage.


WBHarry (approved):
More logical to set it up this way either way :+1:

PR #2304, opened 2026-09-01
[Feature] Support deleting the adversary's attack

CarlosFdez:
Precursor step for collosus adversaries regardless of how we end up doing it.


WBHarry (approved):
:+1:

PR #2306, opened 2026-09-01
[Feature] Custom Resource Damage And Healing

WBHarry:
Fixed so that custom/homebrew resources can be healed/damaged

### DIFF
--- a/module/dice/dhRoll.mjs
+++ b/module/dice/dhRoll.mjs
@@ -1,5 +1,5 @@
 import D20RollDialog from '../applications/dialogs/d20RollDialog.mjs';
-import { triggerChatRollFx } from '../helpers/utils.mjs';
+import { getAllResourceLabels, triggerChatRollFx } from '../helpers/utils.mjs';
 import BaseRoll from './baseRoll.mjs';
 
 export default class DHRoll extends BaseRoll {
@@ -210,6 +210,7 @@ export default class DHRoll extends BaseRoll {
             };
         } else {
             options.message.system.user = game.user.id;
+            options.message.system.allResourceLabels = getAllResourceLabels();
             return options.message.system;
         }
     }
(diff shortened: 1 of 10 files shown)


module/dice/dhRoll.mjs
```diff
@@ -210,6 +210,7 @@ export default class DHRoll extends BaseRoll {
             };
         } else {
             options.message.system.user = game.user.id;
+            options.message.system.allResourceLabels = getAllResourceLabels();
```
CarlosFdez, line 213: if message is a chatmessage document, perhaps this should be a better on the system data, or added during data preparation?

If its not a chatmessage document and its just regular data about the message, carry on!

module/helpers/handlebarsHelper.mjs
```diff
@@ -96,4 +97,8 @@ export default class RegisterHandlebarsHelpers {
     static isNullish(a) {
         return a === null || a === undefined;
     }
+
+    static debug(a) {
+        return a;
+    }
```
CarlosFdez: You accidentally slipped this in but there is already the `log` command in handlebars.

WBHarry: Ah, not intended yeah.
Good to know though, I often toss this in while working when handlebars doesn't do what I expect 🙃

module/dice/dhRoll.mjs
WBHarry: It is a chatmessage document, yeah.

For a chat message, this is pretty much the context prep for rendering though, no? I don't think the actual prepareData of the chatMessage should have this, as it's just a lookup for labels to be rendered.

CarlosFdez: I think its better to predictably bolt data onto the chat message than to unpredictably bolt data onto the chat message. Its not good, but you're technically still adding prepared data to it, you're just not doing it as part of the data prep cycle, but incidentally elsewhere.

Not adding it to the chat message system data at all would be best but I don't know if that's an option. So maybe a getter.

If this is rare then I can simply scrunch my nose a bit.

CarlosFdez: I suppose keep this, this might be something that can't be made better until we implement the renderHTML in the system model, which will let us add more to the context.

WBHarry: Yah, I don't really like it, I just don't see a strictly better way atm

PR #2307, opened 2026-09-01
[Fix] Default DowntimeMoves Migration

WBHarry:
Added a migration to force set all default downtime moves to their defaults. Appearantly some users -still- don't get the automation in chat messages, so this should make that go away while keeping any custom downtime moves.


CarlosFdez:
I'll trust it works, the code looks fine to me.

CarlosFdez (approved):

PR #2309, opened 2026-09-01
[Fix] Companion Levelup

WBHarry:
Fixed so that companion levelup options are correctly checked.
I honestly don't know how this was working before up until recently and how it broke, but this fixes it 🤷


CarlosFdez (approved):

CarlosFdez:
We can leave it as one of the system's great unsolved mysteries

PR #2310, opened 2026-09-01
[Fix] Translation Corrections

### LINKED ISSUE #2203: [Bug] Weapon subtype label not translated in the weapon-filtered browser view
**Describe the bug**
The subtype column in the Compendium Browser shows a properly translated value when browsing the general "Equipment" list, but as soon as I go into "Equipment > Weapons" specifically, that same column shows "Primary" / "Secondary" in English again, even though everything else on screen is in French.

**To Reproduce**
1. Open the Compendium Browser
2. Go to "Equipment" (top level), the Subtype column is translated
3. Click into "Equipment > Weapons"
4. Look at the Subtype  column, it now shows "Primary" / "Secondary" in English

**Expected behavior**
Same translated label in both views, since it's the same data (weapon primary/secondary).

**Screenshots**

Image

**Setup Information:**
 - OS: Windows
 - Browser: 
 - Foundry Version: v14
 - System Version: v2.7.3

**Additional context**
Running a personal French translation module, so I don't know if this only shows up with a non-English language. I peeked at the built JS a bit and it looks like there might be two separate column definitions for this, one that goes through a translation key, and one (used in the weapons-only view) that might just return a raw string instead. Not sure that's actually the cause, just what I noticed poking around.

### LINKED ISSUE #2227: [TYPO] - Duplicate word in Battle Box Relentless description
In the Battle Box's description of its Relentless item, there's a duplicated word "times": "up to two **times times** per GM turn."

### LINKED ISSUE #2308: [TYPO] - Apostrophe mismatch in Core Set Items roll table
Sorry in advance, this time it's just a small typo that's a bit annoying for us, the folks working on the system's localization. :D

In tables_Core_Set_Items_S61Shlt2I5CbLRjz.json, two result names use a curly apostrophe ( ’ ) while the matching items in the compendium use a straight one ( ' ):

Alistair’s Torch → item is Alistair's Torch
Paragon’s Chain → item is Paragon's Chain

Because the names don't match exactly, these rows can't be linked to their items for translation and stay in English.

WBHarry:
Fixes #2308 
Fixed three curly apostrophees in the rolltable result names

Fixes #2227 
Went ahead and corrected a description in the Battlebox adversary aswell

Fixes #2203 
Fixed missing translations on the ItemBrowser's subtype column for weapon

### DIFF
--- a/module/config/itemBrowserConfig.mjs
+++ b/module/config/itemBrowserConfig.mjs
@@ -208,7 +208,7 @@ export const typeConfig = {
             {
                 key: 'system.secondary',
                 label: 'DAGGERHEART.UI.ItemBrowser.subtype',
-                format: isSecondary => (isSecondary ? 'secondary' : isSecondary === false ? 'primary' : '-')
+                format: isSecondary => (isSecondary ? 'DAGGERHEART.ITEMS.Weapon.secondaryWeapon.short' : isSecondary === false ? 'DAGGERHEART.ITEMS.Weapon.primaryWeapon.short' : '-')
             },
             {
                 key: 'system.tier',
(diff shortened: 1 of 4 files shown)


CarlosFdez (approved):
Looks good to me, though I wonder how the roll table got the unconverted names.

PR #2312, opened 2026-09-01
Fix rolls in light mode interface

CarlosFdez:
The scope change is for molilo's char sheet stuff, it prevents any nested themed elements from getting style meant for the application (it was getting the sheet parchment).

But while testing it caught some regressions in chat in light mode.


WBHarry (approved):
:+1: Great catch. Light mode lives once more.

PR #2313, opened 2026-09-01
Rework environment header

### LINKED ISSUE #2174: [Feature] Reduce headings of environments
**Description**
Create a compact or way to shrink the top part of environments to just show the features (or compact version that shows difficulty+title). From the attached image, all the way to the tabs features, notes, etc. is a solid chunk that can't be changed.

This creates a LARGE portion of "Environment" cards that take up space on screen when you want to look at different features of an environment.

**Screenshots**

Image

CarlosFdez:
Closes https://github.com/Foundryborne/daggerheart/issues/2174

Also stealthily fixes Montserrat font registration to work with bold and higher weights. The registered one only went up as high as semibold.

image


WBHarry (approved):
I think this is great, so it's approved for my part :+1:

PR #2316, opened 2026-09-02
[Feature] Active Effect Path Viewer

WBHarry:
It's been requested a lot of times by now, so I went ahead and added in a dialog where a user can see the same ActiveEffect change paths that are displayed in the autocomplete of the ActiveEffect sheet.

There's a button in the Daggerheart Menu, but I set it up so it's available as a macro:
`game.system.api.macros.showActiveEffectPathViewer()`

There's a dialog frame button in the dialog that copies that macro to the clipboard so it's easy to set up such a macro if desired.
image

### DIFF
--- a/module/applications/sheets-configs/activeEffectConfig.mjs
+++ b/module/applications/sheets-configs/activeEffectConfig.mjs
@@ -46,14 +46,15 @@ export default class DhActiveEffectConfig extends foundry.applications.sheets.Ac
      */
     static getChangeChoices() { 
         const ignoredActorKeys = ['config', 'DhEnvironment', 'DhParty', 'DhNPC'];
+        const ignoredRuleKeys = ['standardAttack'];
 
-        const getAllLeaves = (root, group, parentPath = '') => {
+        const getAllLeaves = (root, group, ignoredKeys = [], parentPath = '') => {
             const leaves = [];
             const rootKey = `${parentPath ? `${parentPath}.` : ''}${root.name}`;
             for (const field of Object.values(root.fields)) {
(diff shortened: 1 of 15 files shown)


WBHarry:
Oh, I also added a config entry for `extraPaths` thing, to list the `@prof`/`@cast` etc we have set up.

module/applications/sheets-configs/activeEffectConfig.mjs
```diff
@@ -44,16 +44,17 @@ export default class DhActiveEffectConfig extends foundry.applications.sheets.Ac
      * Get ChangeChoices for the changes autocomplete. Static for use in this class aswell as in settings-active-effect-config.mjs
      * @returns {ChangeChoice { value: string, label: string, hint: string, group: string }[]}
      */
-    static getChangeChoices() {
+    static getChangeChoices() {
```
CarlosFdez:
```suggestion
    static getChangeChoices() {
```

CarlosFdez:
Did some minor adjustments. Also impacts how it handles scrolling, filtering below scrolling, and resize behaviors. I also did something to change what happens on hover (not depicted).

https://github.com/user-attachments/assets/09f03ce2-fc4b-4ae6-b3d4-a14fcbb8d884

CarlosFdez (approved):

CarlosFdez:
Merge if you like the changes, tweak it if you don't.

WBHarry:
A lot better. Awesome 👍 

Especially nice with the scrollbar placement and how the resize takes away from the right

PR #2317, opened 2026-09-02
[Fix] Character Creation Mixed Ancestries

WBHarry:
Fixed so that CharacterCreation mixed ancestry shows the features you won't be given as inactive again.
We removed `uuid` from the data a while back, and the related checks were failing because of that.


CarlosFdez:
Looks good to me

CarlosFdez (approved):

PR #2318, opened 2026-09-02
[Fix] Ancestry Feature Display Order

### LINKED ISSUE #2315: [Bug] Ancestry features are listed in the order they were added
Hello there,

**Issue**
When adding homebrewed ancestries to my game, I realized some of them listed the ancestry features in the wrong order on the card, which is an issue when a player wants a "mixed ancestry". To be clearer: I added a feature as primary, another as secondary, and the card displays the secondary first. See screenshots below.

Upon investigation, I found out the order depends on the order in which these features have been added to the card. Indeed, there is no sorting in `module/data/item/ancestry.mjs` around the line 61:

```js
const features = await getFeaturesHTMLData(await fromUuids(this._source.features.map(f => f.item)));
```

**Proposed fix**
Pseudo-code of a possible fix:

```js
const rankFeature = (a, b) => { ... };
const ordered = [...this._source.features].sort(rankFeature);
const features = await getFeaturesHTMLData(await fromUuids(ordered.map(f => f.item)));
```

Risk seem low since only one function is hit.

I'm a dev, I could test a fix and propose a pull request if you have your hands full and if you wish to integrate this fix to the codebase. I'm of course open to feedback or better ways to fix this matter.

---

**To Reproduce**
1. Create an ancestry
2. Add a feature as secondary feature
3. Add a feature as primary feature
4. See wrong order in the card description

**Expected behavior**
Primary feature must always be displayed first and the secondary second.

**Screenshots**
Image 

Image

**Setup Information:**
- Foundry Version 14.365
- System Version 2.9.0

WBHarry:
Fixes #2315
Fixed so the ancestry description always displays primary and secondary feature in the correct order

### DIFF
--- a/module/data/item/ancestry.mjs
+++ b/module/data/item/ancestry.mjs
@@ -1,6 +1,6 @@
 import BaseDataItem from './base.mjs';
 import ItemLinkFields from '../../data/fields/itemLinkFields.mjs';
-import { fromUuids, getFeaturesHTMLData } from '../../helpers/utils.mjs';
+import { fromUuids, getFeaturesHTMLData, sortBy } from '../../helpers/utils.mjs';
 
 const fields = foundry.data.fields;
 
@@ -59,8 +59,12 @@ export default class DHAncestry extends BaseDataItem {
 
         const baseDescription = `${this.description}${referenceLink}`;
         
-        const featureData = [this.primaryFeature, this.secondaryFeature].filter(Boolean).map(x => x.uuid ?? x);
-        const features = await getFeaturesHTMLData(await fromUuids(featureData));
+        const featureSubtypes = CONFIG.DH.ITEM.featureSubTypes;
+        const featureUuids = sortBy(
+            this._source.features, 
+            f => [featureSubtypes.primary, featureSubtypes.secondary].indexOf(f.type))
+            .map(f => f.item);
+        const features = await getFeaturesHTMLData(await fromUuids(featureUuids));
 
         if (!features.length) return { prefix: null, value: baseDescription, suffix: null };
         const suffix = await foundry.applications.handlebars.renderTemplate(
--- a/module/data/item/ancestry.mjs
+++ b/module/data/item/ancestry.mjs
@@ -59,7 +59,7 @@ export default class DHAncestry extends BaseDataItem {
 
         const baseDescription = `${this.description}${referenceLink}`;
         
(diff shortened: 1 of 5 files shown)


module/data/item/ancestry.mjs
```diff
@@ -58,7 +58,9 @@ export default class DHAncestry extends BaseDataItem {
             ? `<p>@UUID[${reference}]{${label}}</p>` : '';
 
         const baseDescription = `${this.description}${referenceLink}`;
-        const features = await getFeaturesHTMLData(await fromUuids(this._source.features.map(f => f.item)));
+        
+        const featureData = [this.primaryFeature, this.secondaryFeature].filter(Boolean).map(x => x.uuid);
```
CarlosFdez: This may have issues if it didn't load. We can't use primaryFeature/secondaryFeature here, as it may break if its in a separate compendium.

In general, `features[number].uuid`  is safe because the prep will use either item uuid or fallback to the item string source value. But `features[number].item.uuid` may fail.

We should probably refactor primaryFeature/secondaryFeature. While features is volatile in general, that's outright a footgun. Just not sure what to yet.

WBHarry: Hm, when would the issue arrise?
I tested putting the features in one compendium now and the ancestry in another, and it works without issue. (Without having loaded the feature compendium)

The code -is- running `fromUUID`, so it seems to me it should work
image

CarlosFdez: try via console on fresh load. It might be that the features tab is fetching it.

CarlosFdez: `await (await fromUuid("AncestryUUID")).system.getEnrichedDescription()` and then inspect the output.

CarlosFdez: This works I think if you do `x => x.item?.uuid ?? x.item` but I'd rather kill the footgun. ForeignDocumentUUIDField could also initialize with `{ uuid: value }` in the toObject() but then some code may have to change maybe.

WBHarry: Hm, I tested a bunch, and ran the uuid you posted above and it works without issue.

That being said, I'm fine with changing it around.

WBHarry: Moved `sortBy` to functional and updated it's references in files.
Added in your solution to shoot down the footguns with other guns

CarlosFdez (approved):
Tested, works great!

theholocoder:
Thanks for the quick fix guys ❤️ great work!

PR #2319, opened 2026-09-02
[Fix] Rename Loot

WBHarry:
We've hade a missunderstanding since the system's inception that general items are called `Loot`.
Such an item is actually just called an `Item`. 

I changed the translations ,but I don't see a reason to try to mess with the item type key. That'll just have to be a funny historical thing.
image

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -8,7 +8,7 @@
             "feature": "Feature",
             "domainCard": "Domain Card",
             "consumable": "Consumable",
-            "loot": "Loot",
+            "loot": "Item",
             "weapon": "Weapon",
             "armor": "Armor",
             "beastform": "Beastform",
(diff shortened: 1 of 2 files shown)


CarlosFdez:
We can try it. I'm a bit concerned that the foundry item and the loot item will get mixed up, but its just a translation string.

CarlosFdez (approved):

WBHarry:
We don't -have- to change it - but it is the correct verbage for Daggerheart.

I did always find it weird to speak about Loot as an item type. It sounds like something out of Munchkin when you say it like that.

CarlosFdez:
To be clear I don't have an alternative, and most of my fear is because of foundry's own naming conventions. This is fine.

PR #2320, opened 2026-09-02
[Feature] Customized Sidebar Settings

WBHarry:
Customized the sidebar settings tab in the style D&D uses. I think the link to discord is especially good to have around 🧑‍🍳 
I added the game-system section via `onRender` rather than overriding the whole template for greater stability.
image


CarlosFdez:
Would we lose the "has new update" notification, or does foundry inject that later?

WBHarry:
Ah, right. I thought about that at some point then forgot. We should have that yeah.

CarlosFdez (approved):
Neato

PR #2321, opened 2026-09-02
Restore title to environment header

CarlosFdez:
image


WBHarry (approved):
:+1:

CarlosFdez:
Asking for @moliloo approval. The title was their suggestion, and there is a hangup on the title repeating the name. We could replace the title with just the type, but then that'll affect the collapsed behavior, which currently shows the same title (but I supposed it wasn't visible anyways?)

image

moliloo (approved):
i don't see a problem the title repeat the name itself, its just how foundry works, to me its good to merge

PR #2322, opened 2026-09-02
Use card trim variable instead of golden

CarlosFdez:
Reduces the likelyhood of others massacring my boy.

Also sneaked in a thing to add an upper bound to recall border width (3px). I think it looks nicer.


WBHarry (approved):
https://tenor.com/view/look-how-they-massacred-my-boy-godfather-sad-cry-hurt-gif-16433641

PR #2325, opened 2026-09-04
Split DSN settings from appearance and convert to user setting

CarlosFdez:
This splits DSN settings into a new setting and a new menu. The new DSN settings are user settings, which means that they are specific to the world, not to the browser. This is also the case for the actual dice so nice module (which only has a client setting for the purposes of migrations, and otherwise uses user flags).

Also sneaks in the following features, that were kinda annoying to do separately:
- Updating appearance settings re-renders all open actor sheets like homebrew settings do
- Fixes regular migrations to only run for GMs (a player logging in first can trigger a migration and permission failure spam otherwise). I saw this happen when troubleshooting for someone.

There are however some things I withold on, because I didn't wanna increase the size of this with extra things.
- I would like to refactor the app files like `DHAppearanceSettings` to `AppearanceConfig`. This is the naming convention used by foundry. If we did `AppearanceSettingConfig`, this would be the 5e convention. I"m happy with either but am leaning the former.
- I would like the refactor the data models from `DhAppearance` to `AppearanceSetting`. This is the 5e convention for data models.
- I'm not sure about the migration. Basically, every world you log on, including new worlds, will be set to your user setting. We could update the appearance so it only ever happens on one world, but then that might cause issues with other worlds.

### DIFF
--- a/daggerheart.d.ts
+++ b/daggerheart.d.ts
@@ -16,6 +16,7 @@ import DhAutomation from './module/data/settings/Automation.mjs';
 import FearTracker from './module/applications/ui/fearTracker.mjs';
 import DhCountdowns from './module/data/countdowns.mjs';
 import DhEffectsDisplay from './module/applications/ui/effectsDisplay.mjs';
+import { DhDiceSoNice } from './module/data/settings/DiceSoNice.mjs';
 
 // Foundry's use of `Object.assign(globalThis) means many globally available objects are not read as such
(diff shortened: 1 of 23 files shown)


WBHarry:
With DSN v6.3.0 and us defering everything to the module now, this shouldn't be needed 🫡

CarlosFdez:
Nope. I'll draft it to keep around, as the here is a precursor to theming. But otherwise nop.

PR #2326, opened 2026-09-04
Fix styling in ability use messages

CarlosFdez:
Fixes issues with the tags in send to chat. Test it out with a domain card.

I was considering adding the domain card and recall cost here but that might need a greater discussion, so only fixes for now.

image


WBHarry (approved):
:+1:

PR #2328, opened 2026-09-04
[Feature] Conditional Active Effects

WBHarry:
I've added 3 ConditionalTypes:
- **DataCompare**
_Phase:_ `Preparation`
_failureMode:_ `suppress`
This compares a value in rollData to a formula via a chosen comparator method
- **ActionType**
_Phase:_ `Roll`
_failureMode:_ `remove`
This ensures that the action being used is of an allowed type. Allowed types is a SetField, so many could be allowed.
- **Weapon Requirement**
_Phase:_ `Roll`
_failureMode:_ `remove`
This ensures that the action being used is of a specific sort of weapon.

----
I updated the SRD for Horde adversaries using `DataCompare` - removed all the current logic for Horde damage using altValue and deleted the special `horde` ActiveEffect type.

----
I have not made use of `Action Type` and `Weapon Requirement` in this PR. That is handled in the stacked PR.

### DIFF
--- a/module/data/activeEffect/baseEffect.mjs
+++ b/module/data/activeEffect/baseEffect.mjs
@@ -125,8 +125,8 @@ export default class BaseEffect extends foundry.data.ActiveEffectTypeDataModel {
         }
 
         return rollData && this.conditionals.some(x => 
-            x.phase === CONFIG.DH.EFFECTS.conditionalPhases.preparation.id && 
-            x.failureMode === CONFIG.DH.EFFECTS.conditionalFailureModes.suppress.id &&
+            x.constructor.metadata.phase === CONFIG.DH.EFFECTS.conditionalPhases.preparation.id && 
+            x.constructor.metadata.failureMode === CONFIG.DH.EFFECTS.conditionalFailureModes.suppress.id &&
             !x.doesConditionalPass(rollData)
         );
     }
--- a/module/data/activeEffect/baseEffect.mjs
+++ b/module/data/activeEffect/baseEffect.mjs
@@ -127,7 +127,7 @@ export default class BaseEffect extends foundry.data.ActiveEffectTypeDataModel {
         return rollData && this.conditionals.some(x => 
             x.constructor.metadata.phase === CONFIG.DH.EFFECTS.conditionalPhases.preparation.id && 
             x.constructor.metadata.failureMode === CONFIG.DH.EFFECTS.conditionalFailureModes.suppress.id &&
-            !x.doesConditionalPass(rollData)
+            !x.test(rollData)
         );
     }
 
--- a/module/data/activeEffect/baseEffect.mjs
+++ b/module/data/activeEffect/baseEffect.mjs
@@ -124,7 +124,11 @@ export default class BaseEffect extends foundry.data.ActiveEffectTypeDataModel {
             if (change.isSuppressed) return true;
         }
 
-        return rollData && this.conditionals.some(x => !x.doesConditionalPass(rollData));
+        return rollData && this.conditionals.some(x => 
+            x.phase === CONFIG.DH.EFFECTS.conditionalPhases.preparation.id && 
+            x.failureMode === CONFIG.DH.EFFECTS.conditionalFailureModes.suppress.id &&
+            !x.doesConditionalPass(rollData)
+        );
     }
 
     get armorChange() {
--- a/module/data/activeEffect/baseEffect.mjs
+++ b/module/data/activeEffect/baseEffect.mjs
@@ -14,6 +14,7 @@
 
 import { getScrollTextData } from '../../helpers/utils.mjs';
 import { changeTypes } from './changeTypes/_module.mjs'
+import { conditionalTypes } from './conditionalTypes/_module.mjs';
 
 export default class BaseEffect extends foundry.data.ActiveEffectTypeDataModel {
     static defineSchema() {
@@ -40,6 +41,8 @@ export default class BaseEffect extends foundry.data.ActiveEffectTypeDataModel {
             return r;
         }, {});
 
+        
+
         return {
             ...super.defineSchema(),
             changes: new fields.ArrayField(
@@ -56,6 +59,7 @@ export default class BaseEffect extends foundry.data.ActiveEffectTypeDataModel {
                 }),
                 description: new fields.HTMLField({ label: 'DAGGERHEART.GENERAL.description' })
             }),
+            conditionals: new fields.ArrayField(new fields.TypedSchemaField(conditionalTypes)),
             rangeDependence: new fields.SchemaField({
                 type: new fields.StringField({
                     required: true,
@@ -115,11 +119,12 @@ export default class BaseEffect extends foundry.data.ActiveEffectTypeDataModel {
         return true;
     }
 
-    get isSuppressed() {
+    getIsSuppressed(rollData) {
         for (const change of this.changes) {
             if (change.isSuppressed) return true;
         }
-        return false;
+
+        return rollData && this.conditionals.some(x => !x.doesConditionalPass(rollData));
     }
 
     get armorChange() {
--- a/module/data/activeEffect/baseEffect.mjs
+++ b/module/data/activeEffect/baseEffect.mjs
@@ -41,8 +41,6 @@ export default class BaseEffect extends foundry.data.ActiveEffectTypeDataModel {
             return r;
         }, {});
 
-        
-
         return {
             ...super.defineSchema(),
             changes: new fields.ArrayField(
--- a/module/data/activeEffect/baseEffect.mjs
+++ b/module/data/activeEffect/baseEffect.mjs
@@ -117,7 +117,7 @@ export default class BaseEffect extends foundry.data.ActiveEffectTypeDataModel {
         return true;
     }
 
-    getIsSuppressed(rollData) {
+    testIsSuppressed(rollData) {
         for (const change of this.changes) {
             if (change.isSuppressed) return true;
         }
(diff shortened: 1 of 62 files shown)


WBHarry:
Two things to resolve for Horde Damage:

**1)** Do we remove the Automation setting for `Automate Horde Damage`? It used to determine if horde damage got lowered automatically when marking half hp or more. Otherwise we'll have to work it into the conditional flow in some fashion, or have a special case.

**2)** How, if at all, can we migrate existing Horde adversaries (the SRD is of course already updated)? What's needed is to delete the old `Horde` effect, and to create a new one on the `Horde` Feature the adversary has. We can't do either of those in `migrateData` - and a one-shot will of course not hit compendiums.
For stuff coming from the SRD, users can of course just use `refresh` now. For others, I dunno. We -could- add in some system that pops up a warning in sheets that something is out of date and needs to be manually fixed - some manual list of rare cases like this we could fill in.

CarlosFdez:
I don't know what to do with the setting yet (I'll get ideas later), but for migration, I think its a combination of a migration on the actor to delete the horde effect, and a migration on the item to add a horde effect. I don't know if its possible to add/remove items in parent migrations, but that might work? The data for the horde stuff is not part of the effect, its part of the actor. However this may be dangerous to do as a "every data prep" migration, which has trouble with 3pp content in the current version (one day we can version stamp compendiums or compendium data...unless it already is?)

Perhaps an alternative is to forget a perfect migration. There's nothing wrong mechanically with the horde effect being a non transfer effect. Doesn't really matter where it lives.

CarlosFdez:
I attempted to resolve the branch rebase issues, since stacks insist on that. Let me know if I messed up. Roll & Damage dhRoll.mjs's test invocation is the part where im not sure.

WBHarry:
> I attempted to resolve the branch rebase issues, since stacks insist on that. Let me know if I messed up. Roll & Damage dhRoll.mjs's test invocation is the part where im not sure.

The logic looks intact there to me :+1:

module/data/activeEffect/conditionalTypes/actionTypeConditional.mjs
```diff
@@ -0,0 +1,42 @@
...
+                initial: conditionalTypes.actionType.id 
+            }),
+            actionTypes: new fields.SetField(new fields.StringField({
+                label: 'DAGGERHEART.EFFECTS.Conditionals.actionType.actionType',
+                nullable: true,
+                choices: CONFIG.DH.EFFECTS.actionType,
+                initial: null
+            }))
+        }
+    }
```
CarlosFdez, line 29: Still going through code, I'll have to see it in practice, but I feel the the way this is set up might lead to confusion or some incorrect applied things.

So reading the rule, reaction rolls *are* action rolls. Anything that applies to action rolls applies to reaction rolls. The rule is that reaction rolls work like action rules except that they don't generate hope/fear/additional GM moves.

A bonus to action rolls will need someone to select both "action" and "reaction" here. If they select just "action", it'll be wrong. That's a footgun waiting to happen.

EDIT: I'll test it for real later though and see if I can come up with an alternative. "duality" Is distinct from reaction roll at least.

WBHarry:
Migration achieved 😎 

I could make use of the always present `horde` effect for the new one, making it possible.

styles/less/sheets/activeEffects/activeEffects.less
```diff
@@ -23,6 +27,31 @@
         }
     }
 
+    .tab.conditionals {
+        .conditional-outer-container {
```
CarlosFdez: doesn't seem like this has any styling of its own. You could do `.conditional-outer-container .conditional-container` or you could omit entirely to reduce specificity.

I'll have to see why there is even this outer container though.

module/documents/actor.mjs
```diff
@@ -1238,12 +1269,22 @@ export default class DhpActor extends Actor {
 
     /**@inheritdoc */
     *allApplicableEffects({ noSelfArmor, noTransferArmor } = {}) {
+        const isRemovedByConditional = effect => {
+            const { preparation } = CONFIG.DH.EFFECTS.conditionalPhases;
+            const { remove } = CONFIG.DH.EFFECTS.conditionalFailureModes;
+            return effect.system.conditionals.some(x => 
+                x.constructor.metadata.phase === preparation.id && 
+                x.constructor.metadata.failureMode === remove.id &&
+                !x.test(this.getRollData())
```
CarlosFdez: Move this getRollData() outside of the some loop for perf reasons. It will also eventually allow us to splice extra data in there later.

module/documents/activeEffect.mjs
```diff
@@ -7,7 +7,7 @@ export default class DhActiveEffect extends foundry.documents.ActiveEffect {
 
     /**@override */
     get isSuppressed() {
-        if (this.system.isSuppressed === true) return true;
+        if (this.system.getIsSuppressed(this.actor?.getRollData()) === true) return true;
```
CarlosFdez: could this also be called test?

module/data/activeEffect/baseEffect.mjs
```diff
@@ -115,11 +119,16 @@ export default class BaseEffect extends foundry.data.ActiveEffectTypeDataModel {
         return true;
     }
 
-    get isSuppressed() {
+    getIsSuppressed(rollData) {
```
CarlosFdez: I think you can maybe keep isSuppressed(), but have test() "invalidate" it by setting to false?

module/data/activeEffect/baseEffect.mjs
```diff
@@ -40,6 +41,8 @@ export default class BaseEffect extends foundry.data.ActiveEffectTypeDataModel {
             return r;
         }, {});
 
+        
+
```
CarlosFdez:
```suggestion
```

module/data/activeEffect/baseEffect.mjs
WBHarry: Hm, but we can't pass the rollData into a getter. Or did you just mean the name?

module/documents/activeEffect.mjs
WBHarry: `testIsSuppressed` maybe? `system` here is `BaseEffect`. Just `test` doesn't really tell me what that is at least.

module/documents/actor.mjs
```diff
@@ -1238,12 +1269,22 @@ export default class DhpActor extends Actor {
...
+        }
+
         for (const effect of this.effects) {
-            if (!noSelfArmor || effect.type !== 'armor') yield effect;
+            if (!isRemovedByConditional(effect) && (!noSelfArmor || effect.type !== 'armor')) yield effect;
         }
         for (const item of this.items) {
             for (const effect of item.effects) {
-                if (effect.transfer && (!noTransferArmor || effect.type !== 'armor')) yield effect;
+                if (!isRemovedByConditional(effect) && effect.transfer && (!noTransferArmor || effect.type !== 'armor')) yield effect;
```
CarlosFdez: move the isRemoved to the end of both ifs to take advantage of short circuiting.

WBHarry: Smart - most expensive check goes last. :+1:

module/documents/activeEffect.mjs
WBHarry: I changed it to `testIsSuppressed`.

module/documents/actor.mjs
WBHarry: Good call. Moved it :+1:

styles/less/sheets/activeEffects/activeEffects.less
WBHarry: It had no function actually, looking at it. I probably thought I would do something with it, but never did. 
Removed it :+1:

module/documents/activeEffect.mjs
CarlosFdez: There isn't really anything wrong with them both implementing the same "testable" interface. That's a pretty common programming practice.

If both testable things sharing the method signature for a test() function doesn't appeal then sure, but the name still is a bit confusing. You're not testing if it *is* suppressed, rather you're testing if you should early return that its suppressed if you're in the suppression getter.

module/data/activeEffect/baseEffect.mjs
CarlosFdez: I wrote this in another comment chain, but this is a test function, there isn't actually a suppression state. I also now realize why you had issues with multi-phase conditionals.

You can change the implementation of the test function here (call it whatever you want) to set an internal "is suppressed" parameter to true if the test fails. When calling the function again, it can return early if the internal suppressed variable is already set to true "is suppressed". This flag can reset at the very start of data prep. This would allow conditionals run at different stages to add on.

Instead of an internal is suppressed flag, it could instead be called invalid or valid or whatever else you think fits. The name isn't that important.

module/documents/activeEffect.mjs
CarlosFdez: I guess its also weird for an isSuppressed getter to get if its suppressed state from a xIsSuppressed function. Makes it hard to figure out what the relationship is in a few months. From the outside it looks like you figure out if something is suppressed by doing effect.testIsSuppressed(), but that's not quite right.

module/data/activeEffect/baseEffect.mjs
CarlosFdez: That said, there is an appeal with re-evaluating, if you like the idea of a prep phase conditional possibly returning a different result at a later phase. So its not like there's necessarily a right answer here.

WBHarry:
It seems all good to me now.

Actually adding in uses of it in the SRD aside from the Hordes is something we'll do once this is in 🥳

CarlosFdez (approved):

PR #2333, opened 2026-09-05
[Fix] Autovulnerable 0 Fix

### LINKED ISSUE #2331: [Bug] Automatic Application of Vulnerable via Stress Causes Vulnerable to become stuck
**Describe the bug**
When a player character gains the vulnerable condition because they've marked all stress through the vulnerable automation:
- the Vulnerable condition stays on their character sheet even when the stress is cleared
- It's impossible to remove it even after clearing stress through right clicking the condition on the top right, when it should already be unlocked

Image

- If you deactivate it, it is impossible to click the vulnerable condition on the status effects context menu, as it does nothing.
Image

- To fix it, you need to delete the effect directly from the effects tab

**Expected behavior**
- If vulnerable automation is on, that instance of vulnerable should be removed when the character is not maxed on stress
- The condition should not become stuck after deactivating it and become impossible to remove via right click on the upper right corner
- Clicking on the status effects menu should interact with this vulnerable condition to activate/deactivate it

**Setup Information:**
 - OS: Windows
 - Browser Chrome

Image

WBHarry:
Fixes #2331 
Fixed so Vulnerable Automation doesn't fail when stress is set to 0. Just a silly falsy fail.

### DIFF
--- a/module/data/actor/creature.mjs
+++ b/module/data/actor/creature.mjs
@@ -53,7 +53,7 @@ export default class DhCreature extends BaseDataActor {
         if (
             automationSettings.vulnerableAutomation &&
             this.parent.type !== 'companion' &&
-            changes.system?.resources?.stress?.value
+            changes.system?.resources?.stress?.value !== undefined
         ) {
             const { name, description, img, autoApplyFlagId } = CONFIG.DH.GENERAL.conditions().vulnerable;
             const autoEffects = this.parent.effects.filter(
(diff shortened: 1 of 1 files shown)


module/data/actor/creature.mjs
```diff
@@ -53,7 +53,7 @@ export default class DhCreature extends BaseDataActor {
         if (
             automationSettings.vulnerableAutomation &&
             this.parent.type !== 'companion' &&
-            changes.system?.resources?.stress?.value
+            changes.system?.resources?.stress?.value !== undefined
```
CarlosFdez:
```suggestion
            typeof changes.system?.resources?.stress?.value === 'number'
```

Does that work as well? I would prefer a explicit check.

WBHarry: Tested, and it works. Changed to it :+1:

PR #2334, opened 2026-09-05
[Fix] FearTrigger On Reactions

### LINKED ISSUE #2330: [Bug] Reaction rolls trigger "Fear Roll" trigger conditions
**Describe the bug**
Reaction rolls with a fear dice higher than a hope die trigger "Fear Roll" trigger conditions even though reaction rolls are not supposed to be rolls with fear or roll

**To Reproduce**
1. Create a feature with a generic action
2. Give the action a trigger of "Fear Roll" & "Self"
3. Roll a roll with fear with that character that character
4. The trigger activates

**Expected behavior**
It should not activate on reaction rolls. This is applicable to anything that is triggered on a Fear Roll, but I found it because of Know the Tide. which currently gains tokens even on reaction rolls.

**Screenshots**

Image

Image

Image

**Setup Information:**
 - OS: Windows
 - Browser Chrome

Image

WBHarry:
Fixes #2330 
Changed so that `Triggers` triggering onFear only occur on 'action' rolls

### DIFF
--- a/module/dice/dualityRoll.mjs
+++ b/module/dice/dualityRoll.mjs
@@ -285,7 +285,7 @@ export default class DualityRoll extends D20Roll {
         );
         if (dualityUpdates?.length) updates.push(...dualityUpdates);
 
-        if (config.roll.result.duality === -1) {
+        if (config.roll.result.duality === -1 && config.actionType === 'action') {
             const fearUpdates = await game.system.registeredTriggers.runTrigger(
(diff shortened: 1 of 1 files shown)


WBHarry:
I guess this is semi-contentious. It's not hard-defined in the rules. The major consensus seems to be that reaction rolls do not produce `With Hope` or `WithFear` results tho 🤷

CarlosFdez (approved):
Makes sense to me, though I haven't tested it. Go for it if it works.

PR #2335, opened 2026-09-05
[Feature] Conditionals - Roll & Damage

WBHarry:
Added a new Conditional Type
**DamageType**
_Phase:_ `Roll`
_failureMode:_ `remove`
This ensures that the the damage of the action has the correct damage types.

- I removed the separate Roll and Damage entries in `creature.system.bonuses`. Now there's only a singular entry for each and it'll be handled via conditionals.
- Cleaned up the keys and bonus handling for actions so it all works.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -2862,7 +2862,7 @@
                     "actionTypes": "Action Types"
                 },
                 "damageType": {
-                    "damageTypes": "Damage Types"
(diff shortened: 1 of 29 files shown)


CarlosFdez (approved):
It works fine for me, though give the changes another review and feel free to merge.

I didn't look or test things like the migration too closely.

PR #2341, opened 2026-09-06
[Fix] ActiveEffect Displaying

### LINKED ISSUE #2340: [Bug] Visual wording for effects not appearing
**Describe the bug**
When applying effects to tokens, if the effect configuration does not have anything in the "change" tab, the +effectName or -effectName wording does not appear when gaining/losing the effect.
The effect works correctly and appears correctly in the right upper corner.

**To Reproduce**
1. Create an action/reaction/anything that would apply an effect to a target but leave the "change" tab empty
2. Click on apply effect to target
3. See the effect is applied correctly but no visual wording appears.

**Expected behavior**
Gaining and losing an effect without any changes to any attribute keys should still show the visual prompt of the token gaining or losing that effect.

**Screenshots**
This visual prompt is what I'm referring to.
Image

**Setup Information:**
 - OS: Windows
 - Browser Firefox, Foundry app
 - Foundry Version v14 b364
- System Version 2.9.2

WBHarry:
**1)** Fixed so that statuses applied from another effect will be displayed on the canvas token. This had become since we added stacking for effects, since the simple objects generated for statuses would error out on a conditional.
image

----
Fixes #2340 
**2)** This one I'm not 100% on, but I -think- it's good?
When an ActiveEffect is created/deleted enabled/disabled Foundry typically generates a scroll text on the canvas token showing that:
image

Foundry's standard logic does this check though:
`if ( !(this.statuses.size || this.system.changes.length) ) return;`
I'm not sure I agree with that. There are a lot of ActiveEffects being applied that don't have any mechanical things attached to them, but that serve as markers for something that's handled narratively. I think these should still generate the text.
I copied the Foundry function and removed that conditional.

### DIFF
--- a/module/canvas/placeables/token.mjs
+++ b/module/canvas/placeables/token.mjs
@@ -51,7 +51,7 @@ export default class DhTokenPlaceable extends foundry.canvas.placeables.Token {
         this.effects.overlay = null;
 
         // Categorize effects
-        const activeEffects = getIconVisibleActiveEffects(Array.from(this.actor?.allApplicableEffects() ?? []));
+        const activeEffects = getIconVisibleActiveEffects(this.actor?.getActiveEffects() ?? [])
         const overlayEffect = activeEffects.findLast(e => e.img && e.getFlag?.('core', 'overlay'));
 
         // Draw effects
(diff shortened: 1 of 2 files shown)


CarlosFdez:
I agree with you here.

CarlosFdez (approved):

PR #2343, opened 2026-09-06
[Feature] Scene Environments Always Visible

WBHarry:
Changed so that the SceneEnvironment icon is always displayed even if the scene is inactive. 
It's been a point of confusion for users when nothing seems to happen while they're doing session prep with a scene inactive and nothing seems to change.


CarlosFdez:
This is a bit weird looking. There a way to avoid this?

image

CarlosFdez (approved):

WBHarry:
Epic 😎

PR #2346, opened 2026-09-06
[Fix] Move ActionPoint Automation

### LINKED ISSUE #2193: [Feature] Move action points automation setting to homebrew settings
Something that is a bit uinintuitive is how the action points configuration is split up, and I've seen it cause confusion.

The first is that Action Points automation is in the Automation Settings, and it defaults to off (I have it on here)

Image

The second is the Action Points being Enabled in the Variant Rules

Image

Given that the automation setting doesn't make sense at all without the variant rule, I'm thinking perhaps we include whether its automatic or not as part of the variant rule settings. Another alternative is to place variant rules before automation settings, so that the action tokens rule gets seen before the action tokens automation.

WBHarry:
Fixes #2193 
Moved ActionPoint Automation to variant rules and migrated previous value.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -3747,10 +3747,6 @@
                         "label": "Levelup Automation",
                         "hint": "When you've made your choices and finish levelup, the numerical changes are automatically applied to your character."
                     },
-                    "actionPoints": {
-                        "label": "Action Points",
-                        "hint": "Automatically give and take Action Points as combatants take their turns."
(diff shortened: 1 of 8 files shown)


CarlosFdez (approved):
Looks good to me. I went ahead and fixed a problem that can occur on some newer worlds (or those that the GM hasn't touched the settings yet)

PR #2349, opened 2026-09-07
[Feature] DiceChange ActiveEffect change type

WBHarry:
The DiceChange change type handles interacting with data that has to be a dice size (d4, d6, d8, d12, d20). 
It has 3 `diceMode`s that always result in a dice size value or skipping being applied at all:
- **Override**: Set an absolute dice value
- **Upgrade**: Upgrades a dice by an integer number of steps. If there is no dice value already set, skip. 
- **Downgrade**: Dwongrades a dice by an integer number of steps. If there is no dice value already set, skip.

I added this to the Assassin's `Marked For Death` feature and tested leveling a character with it.
- They now start with a `d4` on `system.tokens.assassin.markedForDeath`.
- Executioner starts with a `d6` as it's upgraded once.
- A master Executioner gets a `d8` as it's upgraded a second time.

The damage bonus on the Marked For Death feature is set to: 
`(@system.tier)@system.tokens.assassin.markedForDeath`

----
I added the DiceChanges programatically.
One soft blocker for this issue is that we really have to change the ActiveEffect sheet at this point. It's untenable to handle adding in different types now I feel.

### DIFF
--- a/daggerheart.mjs
+++ b/daggerheart.mjs
@@ -254,7 +254,7 @@ Hooks.once('init', () => {
         SYSTEM.id,
         applications.sheetConfigs.ActiveEffectConfig,
         {
-            types: ['base', 'beastform', 'horde'],
+            types: ['base', 'beastform'],
             makeDefault: true,
             label: sheetLabel('DOCUMENT.ActiveEffect')
         }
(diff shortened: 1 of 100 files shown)


WBHarry:
This `DiceChange` isn't strictly needed for Marked For Death, since just overriding the value would work.
But we've talked about this thing enough times that I think it's worthwhile to have.

If needed it's easy to make a `DiceField` aswell later, since fields can define their `_changeOnAdd`/`_changeOnUpgrade` etc.

PR #2352, opened 2026-09-09
Resolve double formulas, add safety harnesses, and add shim

CarlosFdez:
Some functions were also moved around for organizational reasons


WBHarry (approved):
👍

PR #2353, opened 2026-09-09
[Fix] Weapon/Armor Feature Interaction With Forced Refresh

WBHarry:
Fixed so Weapon and Armor Features don't get removed when doing a forced data refresh


CarlosFdez (approved):
Nice catch

PR #2354, opened 2026-09-09
[Feature] ActiveEffect Ephemeral Type

WBHarry:
Added a new type of ActiveEffect, Ephemeral.

I set it so that by default they are not included by `*allApplicableEffects`. I did this so that they won't get picked up by the standard active effect flow. We do however pass in an option everywhere where we do want them to be considered - such as showing it in the effects tab.

### DIFF
--- a/daggerheart.mjs
+++ b/daggerheart.mjs
@@ -254,7 +254,7 @@ Hooks.once('init', () => {
         SYSTEM.id,
         applications.sheetConfigs.ActiveEffectConfig,
         {
-            types: ['base', 'beastform', 'horde'],
+            types: ['base', 'beastform', 'ephemeral', 'horde'],
             makeDefault: true,
             label: sheetLabel('DOCUMENT.ActiveEffect')
         }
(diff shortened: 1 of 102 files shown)


CarlosFdez:
image

Might need work here, maybe should use the styling that some forms use where the label is above the input.

WBHarry:
Oh yeah, I forgot about getting back to that.
This should do I think 🙃 
image

WBHarry:
Something I've thought of is that there are a bunch of things that want to let you add to a roll -after- it's been made.
image

This is probably a follow up, but for various reasons, we probably need to add some Modify Roll access thing in the chat messages - and aside from freeform editing, it could add support for the sort of thing as above.

WBHarry:
There are also a lot of features that want to do things -before- the roll.
These should assumably not be handled through ephemerals, but we should perhaps add a `Next Roll` kind of duration to make the flow better 🤷 
image

image

CarlosFdez:
hmm, actually, that looks like a buff that activates during the attack, that lasts until the damage of that attack. Bit trickier, but still within the context of what an ephemeral should be doing (even if it wasn't designed for it).

While the attack roll and the damage roll are separate rolls, it is a single exchange and all part of the same action.

WBHarry:
Yeah, there are some expansion points to cover these things. Could be that an Ephemeral should have something like `ApplicationPoint` to designate for things like this.

PR #2355, opened 2026-09-10
[Fix] valueAlt formulas when rolling with fear

CarlosFdez:
the duality getter no longer exists it seems.

Will have to rebase the other stuff on top, but I can do that. Just in case we want to get the fixes any more fixes out there.


WBHarry (approved):
Ah, awesome. I kept meaning to look into this but kept forgetting it.
Tested that it makes it work again :+1:

PR #2356, opened 2026-09-10
[Fix] environment embed issues

CarlosFdez:
The data doesn't have everything, so it can't all be represented, but reduces the divide a bit. Fixes some label issues, what happens on a long potential adversary list, and adds support for dark mode embeds in dark mode journals.


WBHarry (approved):
:+1:
Easy enough to add in the actual stuff if we get the potential adversaries settled.

PR #2357, opened 2026-09-10
[Fix] AutoRoll Criticals

WBHarry:
- Fixed so that critical damage ends up correctly in the chat message when using rollMode `Show Dialog` 
- Fixed so that rollMode `Always` doesn't add double critical damage

### DIFF
--- a/module/data/chat-message/chatDamageData.mjs
+++ b/module/data/chat-message/chatDamageData.mjs
@@ -18,7 +18,7 @@ export class ChatDamageData extends foundry.abstract.DataModel {
             return Roll.fromJSON(this.parent.parent._source.rolls[0]).isCritical;
         }
 
-        return false;
+        return undefined;
     }
 
     static defineSchema() {
--- a/module/dice/damageRoll.mjs
+++ b/module/dice/damageRoll.mjs
@@ -217,7 +217,7 @@ export default class DamageRoll extends DHRoll {
                 }
             }
 
-            if (config.isCritical) {
+            if (config.isCritical && config.dialog.configure !== false) {
                 const total = formulaData.roll.dice.reduce((acc, term) => acc + term._faces * term._number, 0);
                 if (total > 0) {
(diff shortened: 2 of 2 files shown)


CarlosFdez (approved):
Works for me, though so far it looks like critical handling is quite fragile even in the new version. The damage rolls themselves don't actually flag as criticals.

WBHarry:
So the damage rolls only flag it if there's an override, since you can set that the damage should be critical via a radiobutton.
Otherwise it goes by the attack roll's `isCritical` state

CarlosFdez:
Perhaps, though the standalone roll not being a critical is odd. Though it looks like it gets it set anyways when you pull from the chat message, not sure where. Honestly this structure is still quite a bit confusing.

Might be that the roll itself shouldn't have context about the other roll and not be critical, and overrides should only set the damage data itself as critical. Short of ensuring that damage does gain context, that's the other direction my head would go. I was pretty convinced in order to fix this that I needed to ensure the crit flag was set.

Beyond scope but that was my impression from diving into this.

WBHarry:
It's pretty complex, and could probably be done better yeah. We are just contending with a lot of things:
- Rerolls will change the critical status. Continually rewriting the damage data due to that might be bothersome
- The override is a DamageRoll only affair that is separate from the main roll state

Atm these things are reconsiled by the DamageRoll having the base damage stored.
- If the override is true, we upgrade the damage to a critical
- If the override is explici false we ignore any critical state from the main roll
- If the override is undefined we go by the main roll critical state

CarlosFdez:
I see, so a contextless roll solution can work, but it'd rely on initializing the damage roll based on the action roll crit state and skipping the undefined case (which isn't really needed after constructed initialized). But then re-rolls might still be tricky.

Well you satisfied my curiousity. For the scope of the issue, I already approved this one, so merge away! I've decided not ot merge stuff myself unless I need it asap in main as it might interfere with patch note writing going by your workflow.

PR #2358, opened 2026-09-10
[Feature] Countdown Edit Menu Categories

### LINKED ISSUE #2351: [Bug] Countdown creation order mismatch
**Describe the bug**
If I already have countdowns and I create a new one using an environment/adversary action, the countdown does not appear at the bottom. The order seems wrong

**To Reproduce**
1. Have 2 or more countdowns created
2. Create a new countdown using an environment or adversary action
3. Check the order is wrong

**Expected behavior**
There should be an order (or different sort systems) where a new countdown does not appear in the middle of old countdowns

**Screenshots**

Image

**Setup Information:**
 - OS: Windows
 - Browser Firefox, Foundry app
 - Foundry Version v14 b367
- System Version v.2.9.3

WBHarry:
Fixes #2351 
Changed so that EditCountdowns has categories for each countdown type.
I didn't think this would be good at first, but I think it's actually neater than the current display.

image

The benefits I see aside from the look is:
- The order will be the same as the UI element since the categories are in the right order, and the countdowns within also.
- When you add a new countdown, it'll be extremely clear where it ends up since you explicitly click on a + button for that section, instead of always getting a Shortterm countdown that ends up sorted somewhere.

### DIFF
--- a/module/applications/ui/countdownEdit.mjs
+++ b/module/applications/ui/countdownEdit.mjs
@@ -58,12 +58,13 @@ export default class CountdownEdit extends HandlebarsApplicationMixin(Applicatio
                         : 'DAGGERHEART.UI.Countdowns.loop'
                 : null;
             const randomizeValid = !new Roll(countdown.progress.startFormula ?? '').isDeterministic;
-            acc[key] = {
+
+            acc[countdown.type].countdowns[key] = {
                 ...countdown,
-                typeName: game.i18n.localize(CONFIG.DH.GENERAL.countdownTypes[countdown.type].label),
+                typeName: acc[countdown.type].name,
                 progress: {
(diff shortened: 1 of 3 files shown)


CarlosFdez:
I think the add buttons need to be a bit bigger (perhaps some padding to improve click area) and it looks like there's a styling issue with multiple of them

image

WBHarry:
Ah, I thought I was slick that it looked good without any separators, but I guess that was only when none were expanded

WBHarry:
This should be better.
image

CarlosFdez (approved):

CarlosFdez (approved):
Can see how this treats us, though the extra spacing means that even without BG it'd be fine.

Image

PR #2360, opened 2026-09-11
[Feature] Embed tooltips for adversaries/environments in the compendium browser

CarlosFdez:
Makes it easier to preview adversaries and environments in the compendium browser. Merge after conditionals are in.

While doing this I observed there's some consistency differences in the click behavior between adversaries and non-adversaries, but I don't think there's a good solution for it.


WBHarry (approved):
🥳

PR #2361, opened 2026-09-11
Remove world setting to skip roll dialogs

CarlosFdez:
Let me know if you think I should make this into a client setting (rather than a world setting) but I recall nobody really wanted to keep this.

Merge after conditional AEs are in.


WBHarry (approved):
:+1:

PR #2362, opened 2026-09-11
Allow valueAlt to be nullable

CarlosFdez:
If it looks good, we can do an update on all data later, which will definitely be a doozy.


WBHarry (approved):
All good :+1:

PR #2363, opened 2026-09-11
[Fix] errors in valueAlt and conditional migration

CarlosFdez:
change key is nullable for armor effects and such. And damage can be undefined.


WBHarry (approved):
Approving so it can be merged as desired - what I've glossed over looks good

CarlosFdez:
I looked a bit more and it looks fine to me as well. We don't use resultBased anywhere (even though we could) so it seems like its fine. I also don't see explosions around weapon/armor.

This actually removed less than I was expecting it too, looks like there's a lot of new fields.

PR #2365, opened 2026-09-12
[Feature] suppressed indicator and hide unequipped gear effect

CarlosFdez:
The hidden effects are only hidden in the actor sheet.

Also snuck in some refactors:
- Actor effects also use the same partial now
- The overrides aren't utils, so I moved to the main file
- arraysEqual was unused so removed
- shuffleArray is available in lodash and derivatives, so it was moved to functional.mjs


daggerheart.mjs
```diff
@@ -116,6 +116,39 @@ CONFIG.ux.TooltipManager = documents.DhTooltipManager;
...
+    if (!match) return false;
+    let [target] = match.slice(1);
+    target = parseInt(target);
+    for (const r of this.results) {
+        if (r.result === 1) {
+            r.result = target;
+        }
+    }
+};
+
```
WBHarry: This should be set up in our `BaseDie` class
Image

CarlosFdez: I can give it a shot, though I don't know what this feature is, so I might not do it justice. I kinda just moved it elsewhere.

CarlosFdez: I think I moved it properly, though sc seems to give poor feedback in dice so nice and the roll formula.

WBHarry (approved):
Awesum :+1:

PR #2372, opened 2026-09-13
[Feature] Status Conditional

WBHarry:
- Added StatusConditional
- Used for the Rogue subclass card Adrenaline


CarlosFdez:
This can only work with core conditions right? Not arbitrary ones?

For arbitrary ones to work well, source uuid *can* work but it'd be nice if we had optional identifiers for effects for the purposes of targeting them.

WBHarry:
Yeah, this is just for the core statuses.

Originally I started making a `EffectConditional` aswell with the intent of handling stuff depending on specific arbitrary effects, but that's a whole can of worms.

CarlosFdez:
Might be worth going into the can of worms and making it inclusive of this, I imagine there's more examples of that.

Maybe we can add a presence in rules for these sorts of effects that basic ones automatically write to? Others can maybe write themselves to it too in the AE. Then a Data Compare can probably implement them.

PR #2373, opened 2026-09-14
Convert custom change type creation to a select menu with add button

CarlosFdez:
Looks better once there are things in there but otherwise looks a bit weird. The list is currently hardcoded, but later I guess it could be made more generic.

image


WBHarry (approved):
:+1:

PR #2374, opened 2026-09-14
[Fix] Missing Damage Thresholds

WBHarry:
Fixed how damage thresholds work when missing or 0. The logic wasn't correct in these cases. Since severe threshold was checked first, a falsy value there would lead to always taking 3 damage. Now:
- If severe threshold is missing, max damage that can be dealt is 2.
- If major & severe thresholds are missing, max damage that can be dealt is 1.
- If only major is missing then damage will become 1 or 3. (🤷 Why not, it's an undefined case)

----
I also changed so that falsy damage thresholds get rendered as `-` which I think is more correct to the book. The books writes out `None`, but we just don't have space for that.
image


templates/sheets/actors/adversary/sidebar.hbs
```diff
@@ -49,9 +49,9 @@
         <div class="status-section">
             <div class="threshold-section">
                 <h4 class="threshold-label">{{localize "DAGGERHEART.GENERAL.DamageThresholds.minor"}}</h4>
-                <h4 class="threshold-value">{{document.system.damageThresholds.major}}</h4>
+                <h4 class="threshold-value">{{ifThen document.system.damageThresholds.major document.system.damageThresholds.major '-'}}</h4>
```
CarlosFdez, line 52: coalesce helper might actually work better here. Try `{{coalesce document.system.damageThresholds.major "-"}}`

WBHarry: Ah, that's a shame. It actually doesn't work because `coalesce` works strictly of `undefined` and `null`, but we need to also exclude 0 here.

CarlosFdez: thanks for testing it

CarlosFdez (approved):

PR #2375, opened 2026-09-14
[Fix] Weapon Fixes

WBHarry:
- Corrected a long standing display bug where weapon's no longer displayed the weapon damage in the header text
- Fixed so that item/action names are localized in the chat. Unnarmed strike for instance was being rendered unlocalized.
- Corrected Advanced Casting Dagger to have magical damage
- Corrected Bladed Star to be Tier 3 weapon


CarlosFdez:
#2 is the odd one. Using localization keys for item/action names is odd, and translation modules would usually tweak those. If they allow localization keys, then the best practice would be to replace them in our SRD with localization keys, so that stuff like "Mark Stress" become a "DAGGERHEART.Actions.MarkStress.name" or something like that, which might be weird when editing.

Any way to pre-localize unarmed strike when generating the thing, rather than when it gets to the template?

WBHarry:
Hm. It's a bit akward. Normally we definately don't localize any item-names/descriptions and whatnot. 
Unnarmed strike is a little special thought since it's not an item, just a fact of the rules so to speak. 

If we don't want the `localize` in the template, I think we'd have to either just not have it localized, and just write `Unnarmed Strike` in the data model here:
image

Or alternatively localize it during data prep I guess :thinking:

CarlosFdez:
I think initial can be a function. Does it work if we do _loc() there? Do foundry translations require reload and if not, would it interfere with that (maybe not?).

WBHarry:
Curiously enough, doing it as a function makes foundry explode.
But it does work to just straight up do it 🤷 
image

Pushed that up

CarlosFdez (approved):

PR #2376, opened 2026-09-15
[Refactor] Add support for caching settings

CarlosFdez:
Avoids the clean and validation and construction every access. I only applied it to character for now to simplify testing, and because that's where its a bit annoying to diagnose persist: false on the character attack.


WBHarry (approved):
Very elegant with the `settings.onChange` :+1:

PR #2377, opened 2026-09-16
[Fix] Ensure square character/adversary portraits and improve resizing on chrome

CarlosFdez:
Because of the border, the portrait was 1px off before. Chrome in general has pretty crunchy image resizing, so making it easier to target for less rendered art will help. will-change also makes chrome's resizing behavior a bit smarter.

To test will-change, use a very large simple image that is very not square. Chrome will switch to use nearest neighbor in situations where the both the size and the aspect ratio differs.


WBHarry (approved):
:+1: 🦊

PR #2380, opened 2026-09-17
Allow specifying type for ActionField

CarlosFdez:
This makes the field a bit more reliable, fixing an issue I actually had on a weapon for reasons. But it otherwise doesn't really fix any bugs that can occur normally in the system (as far as I can tell), just from console stuff and certain commands.


WBHarry (approved):
:+1:

PR #2381, opened 2026-09-17
[Fix] removing feature when effect has been manually removed

CarlosFdez:
The relevant fix is the `filter(id => document.effects.has(id));` but I also refactored the item feature updates to be shared between weapon and armor.

To test, add barrier, remove the effect, then remove barrier. It'll error without this change, and a new update to the weapon will bring the feature back. After this change, the feature should successfully be removed.


WBHarry (approved):
Nice catch, and nice breaking the function out into one place :+1:

CarlosFdez:
> Nice catch

Surprisingly not a new bug. I caught it because while working on the separate tagify tooltip thing, I had a failed update that made it impossible to update anymore. I realized it would be an issue for regular users if they manually delete the effect, so I split it off into a new fix.

PR #2382, opened 2026-09-17
[Feature] Also update prototype token name when updating actor names that match

### LINKED ISSUE #2348: [Feature] Keep token name synced with actor name when updating actor name
When an actor is first created, the  token name and the actor name match. However, if you then change the actor name, the prototype token name will not update. This can lead to dragging in tokens for "Bob's Character" when Bob has already decided a name for their actor but you forgot to change the token name. This can also lead to mistakes when renaming a creature to flavor it.

In the case that the actor name and the token name (or prototype token) match, and the actor name is updated, it would be nice if the token name was updated to reflect the new actor name. 

This relationship should not go the other way around though. If someone changes the token name explicitly, its likely because they want a different name to be displayed to players.

CarlosFdez:
Closes https://github.com/Foundryborne/daggerheart/issues/2348


WBHarry (approved):
QOL ⭐

PR #2385, opened 2026-09-17
Conditionals On ItemFeatures

WBHarry:
- Updated ActionTypeConditional with a Traits field
- Updated `ItemConfig` entries with conditionals
- Updated the SRD items with zeh script of doom.


WBHarry:
I've updated the ItemConfig to use conditionals. There is an issue I haven't been able to figure out though.
For some reason, if there are any `changes` on an effect, then all `conditionals` disappear and cannot be added back. 🤯 

Assumably due to the same thing, if you add on a WeaponFeature like `paired` on a weapon, the effect doesn't get any conditionals.

module/applications/sheets-configs/activeEffectConfig.mjs
```diff
@@ -230,6 +230,7 @@ export default class DhActiveEffectConfig extends foundry.applications.sheets.Ac
                 break;
             case 'conditionals': 
                 partContext.conditionalOptions = CONFIG.DH.EFFECTS.conditionalTypes;
+
```
CarlosFdez:
```suggestion
```

module/config/itemConfig.mjs
```diff
@@ -1746,6 +1777,16 @@ export const weaponFeatures = {
...
                         img: 'icons/skills/melee/strike-sword-steel-yellow.webp',
                         system: {
+                            conditionals: [
+                                {
+                                    type: 'weaponRestriction',
+                                    weaponType: 'primary'                            
+                                },
+                                {
+                                    type: 'actionType',
+                                    actionTypes: ['attack']
```
CarlosFdez, line 1787: Wouldn't this be superfluous with the weapon restriction?

WBHarry: Not neccessarily. Weapons can have non-attack actions, I figured 🤔

CarlosFdez: hmm, if we go with that line of reasoning, then every weapon feature needs an extra action type restriction. I think the base thing should handle it, just not certain in what way specifically.

WBHarry: Hm..yeah I'm not really sure. It felt strange not to add when the ability is very explicit about it being for attacks

Image

CarlosFdez:
While followup is a bit cumbersome, it will likely be much nicer once ephemerals are in.

CarlosFdez (approved):
While items need a double check (I saw that hatchet was actually incorrect for example and needed a feature cycle) this looks and works well. Nice going.

PR #2386, opened 2026-09-18
[Feature] Make suppression due to unequip more intuitive

CarlosFdez:
This does two things:
- Adds suppression message for unequip specifically
- Does not suppress transfer effects on items that are not on an actor

Because it now only returns if it is suppressed, the logic now also allows suppression due to other reasons like duration expiration, effect type specific restrictions, or what have you, but I suspect that'll never come up.


WBHarry (approved):
:+1:

PR #2387, opened 2026-09-18
Add migration for effects of weapon features

CarlosFdez:
Adds a migration so that things like hatchet update appropriately.

The source only migrations are missing a lot of context, which make it impossible to determine what feature its coming from. While I considered a full feature refresh, this is likely to destroy data, so I did a very simple conditional match.

This will not update items with new effects, primarily because updating the actions as well is a bit cumbersome. If you want to give that a shot though, go for it.


module/systemRegistration/migrations.mjs
```diff
@@ -344,8 +345,9 @@ export async function runMigrations() {
         new Migration_2_8_0_hotfix(),
         new Migration_2_9_1(),
         new Migration_2_9_3(),
-        new Migration_2_10_0
-    ].filter(m => m.version && foundry.utils.isNewerVersion(m.version, lastMigrationVersion));
+        new Migration_2_10_0_DSN(),
+        new Migration_2_10_0_Refresh()
+    ].filter(m => m.version && foundry.utils.isNewerVersion(m.version, '2.9.5'));
```
CarlosFdez:
```suggestion
    ].filter(m => m.version && foundry.utils.isNewerVersion(m.version, lastMigrationVersion));
```

module/systemRegistration/migration-handlers/2_10_0-refresh.mjs
```diff
@@ -0,0 +1,32 @@
+import { MigrationHandlerBase } from './base.mjs';
+
+export class Migration_2_10_0_Refresh extends MigrationHandlerBase {
+    /** @inheritdoc */
+    version = '2.10.0';
+
+    async updateItemSource(item) {
```
WBHarry, line 7: Wouldn't this work as `async updateActiveEffectSource(effectSource, item) {` since the item is available there, and thus removing the need to extract effect updates from the item updates?

CarlosFdez: in the original I did add new effects, but in this one I suppose since I downscaled to just a conditional update, maybe.

I'd like to keep the effect update extraction, as it does add support for adding/removing effects, but there isn't a need to actually use it if we keep it to conditionals.

WBHarry (approved):
:+1:

PR #2388, opened 2026-09-18
[Fix] issue with removing features once multiple have been added

CarlosFdez:
Add accelerator, add followup, then remove accelerator. Before: accelerator effect stuck around. Now: accelerator effect is removed.

Not a new issue, it exists in release. I just discovered it while working on something else.


WBHarry (approved):
:+1:

PR #2389, opened 2026-09-18
[Refactor] Batch item feature updates and preserve ids

CarlosFdez:
Minor optimization that will allow the future potential of stable feature effect ids. Can be merged post release if you prefer.


WBHarry (approved):
:+1:

PR #2391, opened 2026-09-18
[Fix] weapon attack settings not updating when weapon is updated

CarlosFdez:
I tested weapon but please test other sheets, hand hurty too much for full test.


WBHarry (approved):
Huzzah :+1:

PR #2392, opened 2026-09-18
[Feature] Support changing scene darkness via canvas control slider

CarlosFdez:
image


CarlosFdez:
Pf2e uses 5000 as the max delay. I'm using the canvas configs for transitions, which default to 10000. I don't know if I should keep using that or use the 5000 (or perhaps change the defaults for day/night transition to 5000).

WBHarry:
Great feature :+1:
I do think the delay should be lowered. 5000ms works, but I could actually see it being lower after having tested both.

----
Currently the DarknessSlider app sticks around permanently until it's clicked away.
I think we need to do one of two things:
- Either we do the PF2e thing where we make it close if you swap from the Lighting control category
- Or we should add in the `fa-circle-half-stroke` on the left of the DarknessSlider app so there can't be confusion on what that thing on the screen is at some point when you're on the tokens controls. In that case, the icon could also serve as a close button.

CarlosFdez:
Good catch. I handled by adding a swap. Its what I also did in world explorer.

WBHarry (approved):
Awesome :+1:

PR #2393, opened 2026-09-18
[Fix] regression with token tag z index

CarlosFdez:
Fixes recent issue caused by the will-change thing.

Basically the avatar was covering the [Token] thing. Most noticeable that it was happening on adversaries using the default image (anything with a full image would just hide it).


WBHarry (approved):
:+1:

PR #2395, opened 2026-09-18
[Feature] SRD Weapon Pass

WBHarry:
- Fixed `ConditionalMigration` so it correctly sets the distinction between `system.bonuses.roll.bonus` and `system.bonuses.roll.dice` for both roll and damage.
- Went over the SRD to correct cases where the above had created truncated non-func keys like `system.bonuses.roll`
- Checked all SRD weapons so that they're actually updated


CarlosFdez (approved):
Going through all the changes is a bitch much but it looks like you were pretty thorough, and that's good enough for me!

PR #2396, opened 2026-09-18
[Feature] RollType Reaction

WBHarry:
- Added so that an action can set rolltype = reaction. Reaction rolls can be made with a trait or the specific option spellcast trait. (adversaries can leave it blank)
- Updated the SRD with the few cases where the main roll is a Reaction roll.

----
Since Reactions are prominently used via this switch behavior, it really won't work well without us adding the functionality to recalculate Conditionals when stuff changes in the d20Dialog and damageDialog.


module/applications/dialogs/d20RollDialog.mjs
```diff
@@ -210,12 +209,10 @@ export default class D20RollDialog extends HandlebarsApplicationMixin(Applicatio
 
     static toggleReaction() {
         if (this.config.roll) {
-            this.reactionOverride = !this.reactionOverride;
-            this.config.actionType = this.reactionOverride
-                ? 'reaction'
-                : this.config.actionType === 'reaction'
-                    ? 'action'
-                    : this.config.actionType;
+            const nonReactionType = this.originalActionType === 'reaction' ? 'trait' : 'reaction';
```
CarlosFdez, line 212: non reaction type can be reaction?

module/applications/sheets-configs/action-base-config.mjs
```diff
@@ -205,6 +205,15 @@ export default class DHActionBaseConfig extends DaggerheartSheet(ApplicationV2)
             };
         }
 
+        if (context.source.roll) {
+            context.rollTraits = Object.entries(CONFIG.DH.ACTIONS.rollTypeTraits).reduce((acc, [key, value]) => {
```
CarlosFdez: This looks like a job for pickBy!

module/applications/dialogs/d20RollDialog.mjs
WBHarry: Uh, poor variable name I guess 😆 

Going to clean that up

module/applications/sheets-configs/action-base-config.mjs
CarlosFdez: pickBy has saved the day 🦸

templates/dialogs/dice-roll/header.hbs
```diff
@@ -1,11 +1,9 @@
 <header class="dialog-header">
     <div class="dialog-header-inner">
         <h1>{{ifThen rollConfig.headerTitle rollConfig.headerTitle rollConfig.title}}</h1>
-        {{#if showReaction}}
```
CarlosFdez, line 4: What did this do before? I tested with [[/dr]] and it looks like the reaction toggle was still shown.

module/data/fields/action/rollField.mjs
```diff
@@ -98,11 +98,16 @@ export class DHActionRollData extends foundry.abstract.DataModel {
 
     get rollTrait() {
```
CarlosFdez, line 99: In general I think we should make the contents of this getter more generic to facilitate standardAttack reuse. I see you also only check this.trait === 'spellcast' in reaction, when it should be available all the time.

Maybe some helper function for `resolveTrait(actor, trait)` that converts "spellcast" into the spellcastTrait and otherwise just returns trait? Could be `actor.system.resolveTrait(trait)` as well, but it'd need to also exist on base actor and return null in that case.

CarlosFdez: Then in a new PR for I Am the Weapon, we can cram `highest` into resolve trait.

CarlosFdez: Made a PR for it, let me know what you think.

templates/dialogs/dice-roll/header.hbs
WBHarry: ShowReaction was previously gated on the action being used having a DualityRoll.

My thinking was that's incorrect, as an Adversary certainly has `Reaction` features. It seems to me like they would have the same usecases as a Character does for some random action.

module/applications/dialogs/d20RollDialog.mjs
WBHarry: I renamed it to `overrideType` 🪄

templates/dialogs/dice-roll/header.hbs
CarlosFdez: Makes sense!

CarlosFdez (approved):
I'm not sure what test cases to run this through, but my preliminary test shows its fine other than conditional recalculation needs implementing.

PR #2400, opened 2026-09-19
added blue logo and media section in system.json

cptn-cosmo:
small PR to add the blue logo and a media section to system.json


WBHarry (approved):
Can't find any info about the media property, but I -think- it's correct. Guess we'll find out

PR #2401, opened 2026-09-19
[Fix] TakeDamage Button Break

WBHarry:
Fixed takeDamage button not working. Thought I fixed this like 2 times during the main dev cycle, but was appearantly never commited.


CarlosFdez (approved):
Not tested but looks fine to me given what the little I know of this bit of code.

PR #2403, opened 2026-09-19
[Feature] Damage Reduction Conditional Useage

WBHarry:
Added so that DamageReduction considers conditionals. Lets various bonuses like `unstoppable` work specifically against physical damage as intended.

I ended up giving the actor cloning a try to make this happen. Seems to work well. I did end up needing to put in a bit of an oddity perhaps to avoid clones applying active effects in the normal flow - I don't -think- that's an issue though :thinking:.
We want to run -all- the effects of the actor clone through conditionals before applying them, which means the normal prepareEmbeddedDocuments flow should be avoided.
```js
    /**
     * All actual documents should have a UUID. We prevent activeEffect change application if it's a temporary document from a clone.
     */
    prepareEmbeddedDocuments() {
        if (this.uuid) {
            super.prepareEmbeddedDocuments();
        }
    }
```


CarlosFdez:
You can't avoid running the AEs beforehand, as stuff in prepareDerivedData might rely on it.

Instead, when creating the clone, filter the effects list using the conditionals.

WBHarry:
The issue is that if things in `prepareDerivedData` -does- rely on it, then it could end up wrong, since effects we don't want to be applied might have been :thinking:

WBHarry:
Oh wait, I think I get what you mean. The clone should only be created with effects that have already been veted by Conditionals.
Might be a bit rough for those inherited by items, but yeah

CarlosFdez:
I don't know if we currently do, but we eventually do. PF2e system for example initializes skills based on data in there.

```js
const clone = actor.clone({
    effects: actor.effects.filter(testConditionsHere).map(e => e.toObject(true),
}, { keepId: true });
```

CarlosFdez:
Hmm, might also have to filter that its enabled and not suppressed in that filter, but same same.

module/applications/dialogs/damageReductionDialog.mjs
```diff
@@ -36,16 +36,74 @@ export default class DamageReductionDialog extends HandlebarsApplicationMixin(Ap
...
         }, []);
-        const stress = [...Array(actor.system.rules.damageReduction.maxArmorMarked.stressExtra ?? 0).keys()].reduce(
+
+        const { 
+            stress, 
+            maxArmorUsed, 
+            increasePerArmorMark,
+            availableStressReductions, 
+            reduceSeverity, 
+            thresholdImmunities } = this.#getRulesData(actor, action);
```
CarlosFdez:
```suggestion
            thresholdImmunities
        } = this.#getRulesData(actor, action);
```

This will be easier to read and scale better to new props.

Though given that you're reassigning after, I think doing `const rules = this.#getRulesData(actor, action)` and doing stuff like this.maxArmorUsed = rules.maxArmorUsed; will probably work fine and use less lines of code.

module/documents/actor.mjs
```diff
@@ -858,7 +887,7 @@ export default class DhpActor extends Actor {
         return canUseArmor || canUseStress || hasReduceSeverity || hasThresholdImmunity;
     }
 
-    async takeDamage(args, isDirect = false) {
+    async takeDamage(args, actionUuid = null, isDirect = false) {
```
CarlosFdez: If we're breaking API anyways, can we swap everything after the first parameter to object params? That should prevent breaks in the future.

We'll need to determine the type of the first args for the jsdoc as well but that's a separate thing and I can do that later.

CarlosFdez: Could be that one of the args is "action" and have it do something like `const action = typeof options.action === 'string' ? await fromUuid(actionUuid) : options.action`? Not sure. Perhaps the vague resolution is inherently unsafe.

module/documents/actor.mjs
```diff
@@ -909,11 +939,10 @@ export default class DhpActor extends Actor {
                     }
                 }
             } else if (this.type === 'adversary') {
-                const reducedSeverity = hpDamage.damageTypes.reduce((value, curr) => {
-                    return Math.max(this.system.rules.damageReduction.reduceSeverity[curr], value);
-                }, 0);
+                const cloneData = this.getActionClone(await fromUuid(actionUuid));
```
CarlosFdez: Maybe rename to actorClone? cloneData is ambiguous with data to make a clone or the toObject() output of a clone.

CarlosFdez:
Did the nitty gritty review but I like the core idea of what I see here. Good place to do it.

module/applications/dialogs/damageReductionDialog.mjs
WBHarry: I went with `const rules = this.#getRulesData()` :+1:

module/documents/actor.mjs
WBHarry: Hm, I moved it to the second argument being a single param object.
I don't think I prefer letting action be either string or BaseAction though, feels like it makes it more complicated since we need to pass the uuid into the armorQuery either way.

Currently set it as:
```js
/**
     * Handling of the actor taking damage
     * @param {Object} args 
     * @param {{ actionUuid: string, isDirect: bool  }} params 
     * @returns { Object }
     */
    async takeDamage(args, { actionUuid = null, isDirect = false } = {}) {
```
(the jsdoc isn't good but eh 🙃)

module/documents/actor.mjs
WBHarry: Renamed to `actorClone` :+1:

PR #2407, opened 2026-09-20
Allow players to see the compendium browser

CarlosFdez:
Actors are not shown to players though


WBHarry (approved):
I added so that the Adversary & Environment compendiums set `None` as the ownership for non-GM users

PR #2408, opened 2026-09-21
Add more linter rules

CarlosFdez:
Adds brace style and anti multiple newline rules. No more oops spaces.

Enforcing a brace style does prevent the structure you had for if/else, which was fine, but I'd prefer the enforced consistency, and I think this is a decent enough alternative to that.

The value 2 meant error and I swapped it to actually say error which I hope is more explicit.


WBHarry (approved):
I'm fine with it yah. But no more oops space reactions 🫨

PR #2411, opened 2026-09-21
[Fix] Countdown Automation

WBHarry:
Fixed so countdown automation works again. Somehow a negation sign had happened infront which made the check always be undefined 🙃


CarlosFdez (commented):
Oops

CarlosFdez (approved):

PR #2413, opened 2026-09-21
[Feature] Fallback ActiveParty For BP

### LINKED ISSUE #2394: [Feature] Use active party for BP calculation when no PCs are in the encounter
**Description**
In the combat tracker, use the number of PCs in the active party when creating an encounter while there are no active players. This should make it easier to plan out an encounter without having to first add 4 players to the scene.

WBHarry:
Fixes #2394 
Changed so that the active party is used for BP calculations if no characters are in the encounter

**4 person active party**
image

**2 added to encounter**
image


module/applications/ui/combatTracker.mjs
```diff
@@ -44,7 +44,12 @@ export default class DhCombatTracker extends foundry.applications.sidebar.tabs.C
             this.combats
                 .find(x => x.active)
                 ?.system?.extendedBattleToggles?.reduce((acc, toggle) => (acc ?? 0) + toggle.category, null) ?? null;
-        const maxBP = CONFIG.DH.ENCOUNTER.BaseBPPerEncounter(context.allCharacters.length) + modifierBP;
+
+        const activePartyActors = game.actors.find(x => x.type === 'party' && x.system.active)?.system.partyMembers ?? [];
+        const activePartyCharacters = activePartyActors.filter(x => Boolean(x) && x.type === 'character');
```
CarlosFdez:
```suggestion
        const activePartyActors = game.actors.party?.system.partyMembers ?? [];
        const activePartyCharacters = activePartyActors.filter(x => Boolean(x) && x.type === 'character');
```

Untested but try this

I think the boolean check can be removed (and then it can fit on one line) but I'm not sure if that'll stay safe in the future. Right now its a foreigndocumentuuidarrayfield which prunes.

module/documents/tooltipManager.mjs
```diff
@@ -507,10 +507,13 @@ export default class DhTooltipManager extends foundry.helpers.interaction.Toolti
             combat.turns
                 ?.filter(x => x.actor?.isNPC && x.token.disposition === CONST.TOKEN_DISPOSITIONS.HOSTILE)
                 ?.map(x => ({ ...x.actor, type: x.actor.system.type })) ?? [];
-        const characters = combat.turns?.filter(x => !x.isNPC && x.actor) ?? [];
+        
+        const activePartyActors = game.actors.find(x => x.type === 'party' && x.system.active)?.system.partyMembers ?? [];
```
CarlosFdez:
```suggestion
        const activePartyActors = game.actors.party?.system.partyMembers ?? [];
```

same here

WBHarry: Had forgotten `game.actors.party` was a thing. Made the change :+1:

module/applications/ui/combatTracker.mjs
WBHarry: I swapped over to `game.actors.party`. I kept the boolean because I have no faith when uuid stuff is involved 🙃

CarlosFdez: haha, fair. There is a prune during initialization but that's not a guarantee in the future.

CarlosFdez (approved):

PR #2414, opened 2026-09-21
[Fix] No Auto Lose Spotlight

WBHarry:
Removed the previous functionality to remove the spotlight on fail/roll with fear. No ez beuno


module/dice/dualityRoll.mjs
```diff
@@ -333,15 +333,6 @@ export default class DualityRoll extends D20Roll {
         await DualityRoll.addDualityResourceUpdates(config);
 
         if (!config.roll.hasOwnProperty('success') && !config.targets?.length) return;
```
CarlosFdez:
```suggestion
```

CarlosFdez (approved):

